Issue #25, opened 2012-05-29
Add a way to group the doc into sections


stof:
When using the FOSRestBundle custom actions, they all end up in the `others` part of the doc as they don't have the same url than the main resource. Thus, the list and the individual item are different resources and are separated. It would be good to have sections to group actions together. This way, the `other` part would be splitted into each section, staying with the other actions applied on the same objects.
And the bundle would define a default section containing all methods which don't have a section defined (which could be done at the class level once #3 is implemented)

PR #138, opened 2013-02-13
Section grouping

drgomesp:
This implements a new feature called Section Grouping, discussed on issue #25.

The solution is quite simple, and could be reviewed to later refactoring.

All you have to do is declare the same `section` property to two or more actions, and they will be displayed grouped, just the way resources are.

This has precedence over the `resource`, but that can be easily changed.

The default `section` is `'none'`.

**PS** Tests are failing because of the datetime handler for the JMSSerializerBundle, as discussed in #106.


semiaLi:
Hello, I haven't understood how can I get the summary of the sections groups. In fact, I had declared the section and I can see groups but I have not a summary that can help me to go to different sections.

PR #353, opened 2014-03-24
Disable usage of endpoint if host is defined

ruben-podadera:
If host is defined for api doc routes, the generated url is wrong. I used this this to fix the issue.

```
app/console router:debug
[...]
backend_dashboard                   ANY      ANY    back.local.com /
nelmio_api_doc_index                GET      ANY    api.local.com  /doc/ 
```

### DIFF
--- a/Resources/views/layout.html.twig
+++ b/Resources/views/layout.html.twig
@@ -273,9 +273,12 @@
                     }
                     var data = content.length ? content : body;
 
+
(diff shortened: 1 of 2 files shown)


wodka:
@willdurand for me it is not working - see my previous comment. I'm also having issues with the app_dev link for those paths. (meaning it is calling the routes without app_dev.php/)

PR #365, opened 2014-04-17
Fix #357 - doc broken with Validator Constraints in FOSRestBundle requirements

### LINKED ISSUE #357: Doc broken with Validator Constraints in FOSRestBundle requirements
I just used a new feature of the FOSRestBundle (https://github.com/FriendsOfSymfony/FOSRestBundle/pull/700). This PR allow to use Constraints from the sf validator as (Query|Request)Param's requirement, not just a string containing the regex. We can now do:

```
use FOS\RestBundle\Controller\Annotations as Rest;
...
@Rest\QueryParam(name="foo", requirements=@Assert\Regex("#bar#u"), description="toto")
```

Howewer, there is an issue in the rendering of the doc. A Twig_Error_Runtime exception is thrown in "NelmioApiDocBundle::method.html.twig" at line 71:

```
{{ infos.requirement is defined ? infos.requirement : ''}}
```

I suppose this is due to the fact that requirement is not printable as this is a Constraint. Is this confirmed?

pyrech:
Fix #357

PR complete. Do you have any feedback?

### DIFF
--- a/Extractor/Handler/FosRestHandler.php
+++ b/Extractor/Handler/FosRestHandler.php
@@ -14,41 +14,64 @@ namespace Nelmio\ApiDocBundle\Extractor\Handler;
 use Nelmio\ApiDocBundle\Extractor\HandlerInterface;
 use Nelmio\ApiDocBundle\Annotation\ApiDoc;
 use Symfony\Component\Routing\Route;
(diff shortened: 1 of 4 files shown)


chalasr:
Thank's :+1:

PR #457, opened 2014-08-01
No header deletion for forms with files

### LINKED ISSUE #453: Custom headers are not sent from sandbox for multipart/form-data forms
I'm trying to post a form with file field and set additional custom header (internally used for authentication). The problem is that the header is not sent. All works fine for standard forms (application/x-www-form-urlencoded).

deegital:
Fixes #453


dragosprotung:
This fix breaks file upload in sandbox.

PR #497, opened 2014-08-27
Added support for Security annotation

### LINKED ISSUE #494: Add support for @Security annotation
The [`@Security` annotation from SensioFrameworkExtraBundle](http://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/annotations/security.html) was introduced recently. It would be great to see support for this introduced as a viable alternative to [JMS `@Secure` annotation](http://jmsyst.com/bundles/JMSSecurityExtraBundle/master/annotations#secure).

pyrech:
Fixes #494 
- Bumped the required dev version of SensioFrameworkExtraBundle from ~2.1 to ~3.0 because the annotation is not available before
- Added test for `SensioFrameworkExtraHandler` (for `Security` annotation but also for `Cache`)


devantoine:
So, how is it supposed to work?
I've got the `@Security("has_role('ROLE_QSE')")` annotation on my class but nothing shows up in the doc.

I've tried the following annotation on a method too: `authenticationRoles={"ROLE_QSE"}` but nothing happens.

Is it possible to indicate the needed role for a given method or controller based on symfony Security annotation?

Issue #474, opened 2014-08-14
@Input accepts array


mangelsnc:
Hi to everybody!

This it's a really nice bundle to get ready the documentation of an API at the time you write it, but I found something I think forces you to do the things in an only way: the annotation _input_ accepts only  a FormType or a Entity... why not also plain text? or an array, like _statusCodes_?

I found this way will do more flexible this annotation.

dbu:
version 3 is a complete rewrite and uses swagger-php annotations. please check if this is still a problem with version 3 and a recent symfony version.

PR #499, opened 2014-08-28
Annotation: Specify parameters array directly in `input` option

bezhermoso:
Here is a suggestion for implementing this functionality (recently brought up again here: https://github.com/nelmio/NelmioApiDocBundle/issues/474)

We introduce an `@ApiModel` annotation:

```php
@ApiDoc(
     input=@ApiModel({
         "name" = {"type" = "string", "required" = true},
         "age" = {"type" = "integer", "required" = false},
         "group" = { "type" = "integer", "default" = "Users"},
         ...
     })
)
```
Internally we would just normalize the parameters and skip the parsers entirely.

`@ApiModel` can be named, too: 

```php
@ApiDoc(
    output=@ApiModel("User", {"first_name"={ ... }})
)
```
...mainly for use in `output` if it is needed to support that.

The `type` property should be a valid `DataTypes` constant but in lowercase i.e., `DataTypes::INTEGER => integer`

__UPDATE__:

I updated the logic for resolving the parameters by using a the `OptionResolver` component with `ApiDocExtractor`. 

It used to be that the `AbstractFormatter` would ensure that specific parameters are present (`dataType`, `actualType`, `description`, `sinceVersion`, etc.), and all sub-classing formatters would just assume that they are set. But it seems like its the wrong place for this as theoretically not all formatters would need to extend `AbstractFormatter`. 

This is also required for `@ApiModel` parameters to be resolved consistently.

## TODO
* [x] `@ApiModel`
* [x] `@ApiModelCollection` (for collection outputs: https://github.com/nelmio/NelmioApiDocBundle/pull/469)
* [x] Unit tests

### DIFF
--- a/Annotation/ApiDoc.php
+++ b/Annotation/ApiDoc.php
@@ -444,7 +444,8 @@ class ApiDoc
      * @param string $name
      * @param array  $parameter
      */
-    public function addParameter($name, array $parameter)
+    public function
+    addParameter($name, array $parameter)
     {
         $this->parameters[$name] = $parameter;
     }
@@ -709,7 +710,7 @@ class ApiDoc
      */
     public function getResponseMap()
     {
-        if (!isset($this->responseMap[200]) && null !== $this->output) {
+        if (!isset($this->responseMap[200]) && null !== $this->output && !$this->output instanceof ApiModelCollection) {
             $this->responseMap[200] = $this->output;
         }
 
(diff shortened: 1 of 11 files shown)


willdurand:
@bezhermoso willing to rebase your PR?

bezhermoso:
@willdurand OK. Still working through the unit tests.

jaimz22:
What was the reason for closing this? it was never merged and it'd be a really nice feature. Unless something like this is now possible and it's not documented.

PR #543, opened 2014-11-12
Fix/issue 398

### LINKED ISSUE #398: Choices values autoloading can cause excessive memory use
Since PR #329 (v2.5.2), the annotation parser [attempts to load every possible values of FormType ​​"choice" type fields for display in the sandbox](https://github.com/nelmio/NelmioApiDocBundle/blob/master/Parser/FormTypeParser.php#L183) 

But in the case of a form where a field is bound to a Doctrine entity (cf. exemple bellow), the parser loads the entire table, which can cause huge memory usage, even "memory limit exceed" errors. 

This is a big regression for me. 
I have no clear idea on how to solve it. Can be check if the list of choices is static or not before loading, or allow to disable this feature.

Any idea ?

```php
<?php

use Symfony\Component\Form\AbstractType;
use Symfony\Component\Form\FormBuilderInterface;

class ProductType extends AbstractType
{
    {
        $builder
            ->add('title', 'text', [
                'max_length' => 255,
                'required' => true
            ])
            ->add('program_id', 'entity', [
                'required' => true,
                'class' => 'Some\Namespace\Entity\Program',
                'query_builder' => function ($repo) {
                    return $repo->createQueryBuilder('program');
                },
                'property_path' => 'program'
            ]); 
    }
}
```

### LINKED ISSUE #500: Form "entity" type interpretation
For example if I have a form with an entity type field, which holds >3000 entries parser will select all of them and print in format section. I think this is not very efficient and "entity" type should not be interpreted as choice.

### LINKED ISSUE #571: FormTypeParser multiple entity choice
Using this bundle within symfony2 project. 

Form type has an entity field with a class. 

After doing some performance testing and adding 10'000 records, API doc generates infinitely until it times out.

The issue is that it's generating choice format array, which just holds duplicates of the same entity type:
"["Proxies\\__CG__\\\\ClientBundle\\Entity\\Client","Proxies\\__CG__\\ClientBundle\\Entity\\Client",...]"

Any solution to this? Thanks

mojoLyon:
Fixes #398 .
Fixes #500 
Fixes #571

The solution for testing is not very good , regarding on the trick for EntityType. Doing this in this way to avoid the use of doctrine bridge bundle configuration changement, mocking a lot of object,..).

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -132,6 +132,7 @@ class Configuration implements ConfigurationInterface
                                 })
                             ->end()
                         ->end()
+                        ->booleanNode('entity_to_choice')->defaultValue(true)->end()
                     ->end()
                 ->end()
                 ->arrayNode('swagger')
--- a/DependencyInjection/NelmioApiDocExtension.php
+++ b/DependencyInjection/NelmioApiDocExtension.php
@@ -41,6 +41,7 @@ class NelmioApiDocExtension extends Extension
         $container->setParameter('nelmio_api_doc.sandbox.request_format.method', $config['sandbox']['request_format']['method']);
         $container->setParameter('nelmio_api_doc.sandbox.request_format.default_format', $config['sandbox']['request_format']['default_format']);
         $container->setParameter('nelmio_api_doc.sandbox.request_format.formats', $config['sandbox']['request_format']['formats']);
+        $container->setParameter('nelmio_api_doc.sandbox.entity_to_choice', $config['sandbox']['entity_to_choice']);
 
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
         $loader->load('formatters.xml');
--- a/Parser/FormTypeParser.php
+++ b/Parser/FormTypeParser.php
@@ -36,6 +36,11 @@ class FormTypeParser implements ParserInterface
     protected $formRegistry;
 
     /**
+     * @var boolean
+     */
+    protected $entityToChoice;
+
+    /**
      * @var array
      */
     protected $mapTypes = array(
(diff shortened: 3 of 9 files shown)


jrjohnson:
Any status on this?  It would be awesome if generating docs didn't also cause 700+ queries to be run.  Can I help?

mojoLyon:
I think @willdurand is a bit overbooked at this time. Just be patient :)

lopsided:
PR looks good to me, would be great to get this merged in. At the moment we've had to set our entity fields to type=text/integer so we can still use the sandbox without hitting memory limits!

TomAdam:
This was causing me some serious issues too, but it made me realise that the entity type is not really suitable in an API context. I'm not certain this is really a problem that this bundle needs to solve. While the solution presented here is cool, and probably should be merged, there is a better way to solve this.

If the docs are running the queries because they are loading the forms, so are your form posts. The entity type is based on the choice type which loads all the possible options into a ChoiceList during the form build so that it can be used to display them in a select widget, or so that an identifier can be transformed during form submission. Given the context of an API, we didn't need this list and were just looking to ensure the entity referenced existed. I suspect most users will be in the same position.

To solve this we created a new form type that took a repository as a parameter and worked with a transformer that managed the hydration of an entity. It is pretty basic and tailored to our use case, but it could serve as a good starting point for others with the same problem. See https://gist.github.com/TomAdam/ec6e39998653d043adb6

nspyke:
+1 for having this merged. Need it ASAP.

willdurand:
@mojoLyon could you please rebase this PR? It will be merged right after :)

willdurand:
thanks for your work btw

mojoLyon:
@willdurand working on rebase. 
I resolved conflict but tests failed, 9 errors and 6 failures. It seem error are the same on master branch.

mojoLyon:
@willdurand rebasing and squashing done.

willdurand:
thank you!

PR #557, opened 2014-12-15
Fix PHP 5.4 dependency in FormTypeParser and also fix 'something else'

TomiS:
Hi

Here's a pr that should remove the hard dependency for PHP 5.4 by falling back to older solution for older PHP versions.

I also added isset() check for $embeddedType because adding it was the only way I managed to make my project's apidoc work at all (after upgrading from version 2.3.1 in which the doc worked nicely). I have no idea of the meaning or side effects of this addition but without it the whole thing tried to create form with null object and crashed.


willdurand:
I'd like to avoid such hacks, the bundle already supports a large set of PHP/Symfony versions

PR #558, opened 2014-12-22
Update FormTypeParser.php

DocLommi:
The json_encode parameter 'JSON_UNESCAPED_UNICODE' enables german Umlaute to be displayed correctly.
Having methods protected instead of private facilitates extension of the class (for whatever reasons, here: fixing Umlaute issue).


willdurand:
I think it is fixed now

DocLommi:
Cool, thnx :)

PR #572, opened 2015-02-05
Update layout.html.twig

ReinDeVos:
When submitting a sandbox-form containing 'file'-fields the formData-object is posted instead of the params-object.

The '_format'-field should only be added to the formData-object when {_format} is not present in the URL.
'_format' gets stripped from the params-object when {_format} is present (~ line 380), but it doesn't get stripped from the formData-object as it's not available in the FormData-api. 

When not necessary, it should not be added. It could result in a 'This form should not contain extra fields' error.

Keep up the good work! The api-doc-bundle saved me hours of work already! :-)


willdurand:
Closing it for now as the `master` branch has changed a lot.. Feel free to re-submit this PR, sorry.

PR #577, opened 2015-02-16
Allow specifying the array name for jms parsed inputs

jhogendorn:
When you utilise a formtype as the input, you get the various elements as a subarray. Theres no way to get this behaviour currently when you utilise the jms serializer. This patch checks for a name specification in the annotation and subarrays the parameters if it exists.

This PR is discussion at the moment. If the maintainer is roughly happy with the concept I will fix up unit tests and documentation appropriately and update the PR.


willdurand:
Can we move this discussion to an issue first? I am not sure to understand..

PR #582, opened 2015-03-02
Update ApiDocController.php for fix error: Maximum execution time of 30 seconds exceeded

Nolwennig:
Fix error: Maximum execution time of 30 seconds exceeded


willdurand:
nope

PR #586, opened 2015-03-16
Pass children options to subform

PedroTroller:
Hi.
Just a little fix.
When you use a form and you pass options to a subform, into documentation, options are not transmitted.

```php
class MyFrom extends AbstractType
{
    {
        $builder
            ->add('name', 'text')
            ->add('title', 'text')
            ->add('media', 'media', ['with_url' => false])
        ;
    }
}
```

```php
class MediaType extends AbstractType
{
    {
        if (true === $options['with_url']) {
            $builder
                ->add('url', 'url')
            ;
        }

        if (true === $options['with_file']) {
            $builder
                ->add('file', 'file')
            ;
        }
    }
}
```

### DIFF
--- a/Parser/FormTypeParser.php
+++ b/Parser/FormTypeParser.php
@@ -126,6 +126,7 @@ class FormTypeParser implements ParserInterface
         $parameters = array();
         foreach ($form as $name => $child) {
             $config     = $child->getConfig();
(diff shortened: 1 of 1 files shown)


willdurand:
thanks!

PR #590, opened 2015-03-18
Add html template

valiknet18:


spolischook:
:+1:

willdurand:
I'd rather use template overriding: http://symfony.com/doc/current/book/templating.html#overriding-bundle-templates.

PR #591, opened 2015-03-19
Sandbox - don't set Content-Type header when form contains a file type

pasinter:
Setting Content-Type manually when a file is present breaks the multipart request as request boundary is never appended. FormData will automatically append Content-Type=multipart/form-data with the correct multipart boundary if it contains at least one file, so simply omitting the Content-Type fixes this issue


willdurand:
I guess it is fixed already.

PR #595, opened 2015-03-27
No parameter representing sub-form on compound form types

JMLamodiere:
On compound forms using sub form, avoid displaying an useless parameter with a type like "object (FooFormType)". Directly displays parameters of the subform instead.

Otherwise, the sandbox will display a useless field requirering an "[object (FooFormType)] Value" above the sub-form parameters themselves.


sroze:
I think that you're fixing the problem at the wrong place. I agree that fields "requiring" an object in the sandbox is a non-sense.
By the way, you should simply hide (or do not render) that kind of inputs in the sandbox.

PR #599, opened 2015-04-07
Fix DunglasJsonLdApiBundle support

dunglas:
* Match the new bundle API
* Type now default to `string` if not automatically guessed (instead of an error before)


dunglas:
Thank you @willdurand!

PR #605, opened 2015-04-16
allow multi-line return description based on multi-line comments

angrygreenfrogs:
I had a small problem where my output class had larger descriptions of data members like this for example:

    /**
     * Returned on error (status code 400, 403)
     * 400 - Pre-processing error occurred (see message)
     * 403 - Authorization failure occurred (see message)
     * 1050 - API Key is invalid
     * @MongoDB\String @MongoDB\Index(unique=true, order="asc")
     * @Expose
     * @Groups({"User"})
     * @Type("string")
     */
    protected $code;

I wanted that to display nicely in the "Description" of the Return section on separate lines, so just added the twig filter, otherwise the \n's were just being directly put in the description in the output.

### DIFF
--- a/Resources/views/method.html.twig
+++ b/Resources/views/method.html.twig
@@ -156,7 +156,7 @@
                                 <td>{{ name }}</td>
                                 <td>{{ infos.dataType }}</td>
                                 <td>{% include 'NelmioApiDocBundle:Components:version.html.twig' with {'sinceVersion': infos.sinceVersion, 'untilVersion': infos.untilVersion} only %}</td>
-                                <td>{{ infos.description }}</td>
+                                <td>{{ infos.description|nl2br }}</td>
                             </tr>
                         {% endfor %}
                     </tbody>


willdurand:
For status code, there is a parameter for that in the `@ApiDoc` annotation.

angrygreenfrogs:
Hi there,

True, but I have custom error codes in there as well, not just HTTP status codes.

But also that wasn't the point about status codes, just an example, in general it's nice to be able to have multiline comments be formatted in the apidoc output with line breaks.

It could be for anything, for example we have a wine that's output like this:
    /**
     * Returned on success (status code 200)
     * Wine object
     * name - Wine name
     * color - Wine color (e.g. "Red")
     * sweetness - Wine sweetness (e.g. "Dry")
     * winery_name - Winery name
     * winery_description - Paragraph description of the winery
     * region_name - Wine region name
     * blend_name - Blend name (e.g. "Shiraz Viognier")
     * @Expose
     * @Groups({"Success"})
     * @Type("ApiBundle\Document\Wine")
     */
    protected $wine;

In the current output those multiline comments are all bunched together on one line, so that's why I suggested to just add in the twig filter to break them onto separate lines.

Cheers,
Andrew

willdurand:
IIRC such a part of the doc is converted into markdown, so you may be able to add `\n` by adding blank lines.

angrygreenfrogs:
Hi there,

I was hoping to re-open this request and ask if you'd still consider it?

I've checked again and adding blank lines doesn't work.

The only way is to add the |nl2br twig filter.

If you do, this very nicely outputs the multi-line descriptions for the return values with line-breaks in the HTML output.

I've been maintaining a fork of the api doc bundle myself just for adding that one line of code.

PR #606, opened 2015-04-22
You should be able to mix parameter settings from other annotations ...

gigo1980:
You should be able to mix parameter settings from other annotations also from the 'parameters' section in den the apidoc. currently you can only use "parameters" or "input". 
if you try to fill from a own apidoc annotation "Nelmio\ApiDocBundle\Extractor\HandlerInterface" the parameters, they will be removed if you define an "input" section.


willdurand:
could you add some tests?

PR #609, opened 2015-04-28
Upgrade jQuery 1.7.2 to 1.11.2

SoboLAN:
The jQuery library which is bundled even with the latest version of ApiDocBundle is about 3 years old (which is half of a lifetime in the software world).

This PR upgrades it to a more recent version in order to benefit from the features and bug fixes that were added in those 3 years.

Compressed version of the library is kept (obviously).


SoboLAN:
@willdurand any comments ? Is there something I should have changed here which I forgot to do ?

willdurand:
@SoboLAN sorry, I am travelling, and the comment did not get saved... Here is my comment:

It was a bad idea to bundle jQuery with this bundle, unfortunately it is a bit too late, and now I am not sure about how to deal with that. This part is not tested, so I am not confident enough to bump the jQuery version..

PR #610, opened 2015-04-28
Enhanced DunglasApiBundle support

dunglas:
- Works with the last version of the bundle
- Supports nested objects (read and write)
- Lot of enhancement and bug fixes
- Some doc

nelmioapidoc-and-dunglasapi


dunglas:
The error is not related with the PR: https://github.com/sebastianbergmann/phpunit/issues/1684

willdurand:
thank you, really appreciated!

Issue #350, opened 2014-03-14
API Doc Dump returns twig template


cryptiklemur:
When I'm running the following command,

``` sh
php app/console api:doc:dump --format=html
```

It returns a twig template, instead of html

``` html
{% extends "NelmioApiDocBundle::layout.html.twig" %}

{% block content %}
    {% for section, sections in resources  %}
        {% if section != '_others' %}
            <div id="section">
            <h1>{{ section }}</h1>
        {% endif %}
        {% for resource, methods in sections %}
            <li class="resource">
                <div class="heading">
                    {% if section == '_others' and resource != 'others' %}
                        <h2>{{ resource }}</h2>
                    {% elseif resource != 'others' %}
                        <h2>{{ resource }}</h2>
                    {% endif %}
                </div>
                <ul class="endpoints">
                    <li class="endpoint">
                        <ul class="operations">
                            {% for data in methods %}
                                {% include 'NelmioApiDocBundle::method.html.twig' %}
                            {% endfor %}
                        </ul>
                    </li>
                </ul>
            </li>
        {% endfor %}
        {% if section != '_others' %}
            </div>
        {% endif %}
    {% endfor %}
{% endblock content %}
```

PR #615, opened 2015-04-30
Add documentation statement about twig dependency

matks:
I built a Symfony2 API without twig, so when I first added NelmioApiDocBundle it returned a not-rendered twig template, juste like https://github.com/nelmio/NelmioApiDocBundle/issues/350 . It took me some time to understand the issue so it might be worth a statement in the doc, to save other people's time).


willdurand:
thanks!

PR #617, opened 2015-05-01
Fix deprecated forms setDefaultOptions method

Soullivaneuh:
This PR fixes deprecated notices since SF 2.7.

Symfony 2.3+ BC kept.


willdurand:
thanks!

PR #627, opened 2015-05-22
Styles for lists inside .content

csuarez:
Hello,

I have added styles for the ul and ol tags inside the content of an operation. You can see here the before:

api documentation

An the after:
api documentation2

The change is straightforward, I think that it doesn't need more explanations :)

Regards!


willdurand:
tip top, thanks!

PR #630, opened 2015-05-26
Add view option to api:doc:dump command

munkie:
Add _--view_ option to _api:doc:dump_ command to be able to dump docs only for provided view.


willdurand:
Awesome! Thank you!

Issue #636, opened 2015-06-02
All tests fails because of Symfony 2.7 deprecation notice


magnetik:
Hi,

Since symfony 2.7 have been released, all tests are failing because of deprecation notices.

willdurand:
Right. Would you mind fix the test suite?

magnetik:
It seems that #614 is on its way to fix this. I'll try to help on that.

magnetik:
Looks like it's okay now since the tests of #637 are passing.

skonsoft:
Hello,

there is some deprecations are not fixed yet:

DEPRECATED - The "pattern" option in file "/var/www/dev/my_project/vendor/nelmio/api-doc-bundle/Nelmio/ApiDocBundle/Resources/config/routing.yml" is deprecated since version 2.2 and will be removed in 3.0. Use the "path" option in the route definition instead. 

DEPRECATED - The "_method" requirement of route "nelmio_api_doc_index" in file "/var/www/dev/my_project/vendor/nelmio/api-doc-bundle/Nelmio/ApiDocBundle/Resources/config/routing.yml" is deprecated since version 2.2 and will be removed in 3.0. Use the "methods" option instead. 

Can you fix this please ?

Thank you !

PR #637, opened 2015-06-02
Handle "array" parameter in FOSRestBundle QueryParam or RequestParam

magnetik:
It allows to display `[]` after a requirement when `array` is defined. 

Not a big fan of the duplication of `$annot->array` test, but was not fan of modifying `handleRequirements` for accepting the annotation directly (and not only the parameter)


magnetik:
The tests will fail because of #636

magnetik:
All green :+1:

PR #614, opened 2015-04-30
Fix deprecated route options

Soullivaneuh:
This PR fix deprecated routing config throwing related errors:

> The "pattern" is deprecated since version 2.2 and will be removed in 3.0. Use the "path" option in the route definition instead.

> The "_method" requirement is deprecated since version 2.2 and will be removed in 3.0. Use the setMethods() method instead or the "methods" option in the route definition.

> The "_scheme" requirement is deprecated since version 2.2 and will be removed in 3.0. Use the setSchemes() method instead or the "schemes" option in the route definition.

We have to bump `symfony/framework-bundle` to `~2.2` version. Not a big deal because 2.1 is not maintained anymore.

PR #640, opened 2015-06-04
Include Symfony 2.7 stable on Travis and improve configuration

Soullivaneuh:
Also added `symfony/phpunit-bridge` to handle deprecated issues with ease.

Tests will fail because of deprecation (see #636).

Route config issues are solved on #614. I'll take look for form issues on another PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,26 +13,35 @@ php:
   - nightly
   - hhvm
 
+env:
+    global:
+        - SYMFONY_DEPRECATIONS_HELPER=weak
+
 matrix:
     include:
         - php: 5.6
-          env: SYMFONY_VERSION='2.3.*'
+          env: SYMFONY_VERSION=2.3.*
         - php: 5.6
-          env: SYMFONY_VERSION='2.4.*'
+          env: SYMFONY_VERSION=2.4.*
         - php: 5.6
-          env: SYMFONY_VERSION='2.6.*'
+          env: SYMFONY_VERSION=2.6.*
         - php: 5.6
-          env: SYMFONY_VERSION='dev-master'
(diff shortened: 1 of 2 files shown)


stof:
You can avoid making test fail because of deprecations by setting the appropriate env variable though

.travis.yml
```diff
@@ -12,27 +12,34 @@ php:
   - 5.6
   - nightly
   - hhvm
+  - hhvm-nightly
```
stof: this should be reverted. hhvm-nightly does not exist anymore on Travis

.travis.yml
```diff
@@ -12,27 +12,34 @@ php:
...
-    allow_failures:
+          env: SYMFONY_VERSION=2.7.*
+        - php: 5.6
+          env: SYMFONY_VERSION=2.8.*@dev
         - php: 5.6
-          env: SYMFONY_VERSION='dev-master'
+          env: SYMFONY_VERSION="3.0.x-dev as 2.8"
+    allow_failures:
         - php: nightly
+        - php: hhvm
```
stof: why allowing failures on hhvm ? it was not allowed previously

Soullivaneuh:
@stof And what is this magic env variable? :-)

stof:
It is ``SYMFONY_DEPRECATIONS_HELPER=weak``, as explained in the readme of the bridge: https://github.com/symfony/symfony/blob/2.8/src/Symfony/Bridge/PhpUnit/README.md

Soullivaneuh:
@stof Thanks for your advice.

@willdurand Tests should be green now. Ready to merge. :+1:

willdurand:
thanks!

PR #641, opened 2015-06-04
Fix deprecated FormType::setDefaultOptions usage

Soullivaneuh:
Related to #636.

Symfony 2.3+ BC kept.

Failing test about routing configuration are fixed in #614.

### DIFF
--- a/Tests/Fixtures/Form/DependencyType.php
+++ b/Tests/Fixtures/Form/DependencyType.php
@@ -13,6 +13,7 @@ namespace Nelmio\ApiDocBundle\Tests\Fixtures\Form;
 
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\OptionsResolver\OptionsResolverInterface;
 
 class DependencyType extends AbstractType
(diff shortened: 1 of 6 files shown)


Tests/Fixtures/Form/DependencyType.php
```diff
@@ -32,10 +33,20 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
 
     /**
-     * {@inheritdoc}
+     * {@inheritdoc}
+     *
+     * @deprecated Remove it when bumping requirements to Symfony 2.7+
      */
     public function setDefaultOptions(OptionsResolverInterface $resolver)
     {
+        $this->configureOptions($resolver);
```
phansys: I have made the same update in sonata-project/SonataMediaBundle#760, but I've added an extra check for the implementation given in ```$resolver```. I don't know any case where ```OptionsResolverInterface``` have other implementation than ```OptionsResolver```, but since technically it would be allowed, I think it should be checked.

Soullivaneuh:
@willdurand Branch rebased since #640. Should be OK now. :+1:

lsmith77:
minimum currently allowed version of Symfony is 2.1, but it should probably get bumped to 2.3 anyway.

relates to https://github.com/nelmio/NelmioApiDocBundle/pull/647

Soullivaneuh:
> minimum currently allowed version of Symfony is 2.1, but it should probably get bumped to 2.3 anyway.

Didn't notice. Will make a PR for that. :+1:

Tests/Fixtures/Form/DependencyType.php
stof: Symfony itself does not have such check in the forward compatibility layer in 2.7+, so there is no need to add it in bundles wanting to implement the forward compatibility for 2.3+ IMO

PR #643, opened 2015-06-05
Fix DunglasApiBundle filters documentation support.

ogizanagi:


Extractor/AnnotationsProvider/DunglasApiProvider.php
```diff
@@ -127,8 +127,8 @@ private function getApiDoc(
 
         $data['filters'] = [];
         foreach ($resource->getFilters() as $filter) {
-            foreach ($filter->getDescription($resource) as $name => $data) {
-                $data['filters'][] = ['name' => $name] + $data;
+            foreach ($filter->getDescription($resource) as $name => $definition) {
+                $data['filters'][] = ['name' => $name] + $definition;
```
willdurand: `+`?

ogizanagi: $definition is an array, and we need to add in the filters collection the union of $definition and `['name' => $name]`.
I didn't changed this piece of code. The only issue was the `$data` variable name in the foreach loop, which was already used above (`$data['filters']`). Which lead to the fact only the last filter was available in the sandbox & documentation.

willdurand: ok

PR #645, opened 2015-06-08
Only display filters on collections GET services

ogizanagi:
Filters were documented on every operations on the resource, but it does not make sense for operations on single entities.

@dunglas : Do you think it might be an issue retrieving the resource filters only for GET operations on collections ?


dunglas:
Looks good to me :+1:

willdurand:
thanks!

PR #646, opened 2015-06-09
DunglasApiBundle name converter support

ogizanagi:
This follows up https://github.com/dunglas/DunglasApiBundle/pull/94 were serializer name converters are properly handled when building the Json-Ld context.
Properties names should also be converted using the specified converter in the documentation & sandbox, hence this PR.

Ping @dunglas

### DIFF
--- a/Parser/DunglasApiParser.php
+++ b/Parser/DunglasApiParser.php
@@ -17,6 +17,7 @@ use Dunglas\ApiBundle\Mapping\AttributeMetadataInterface;
 use Dunglas\ApiBundle\Mapping\ClassMetadataFactoryInterface;
 use Nelmio\ApiDocBundle\DataTypes;
 use PropertyInfo\Type;
+use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
 
 /**
  * Use DunglasApiBundle to extract input and output information.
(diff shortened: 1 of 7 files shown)


Tests/Fixtures/app/config/dunglas_api.yml
```diff
@@ -16,6 +16,9 @@ dunglas_api:
     description: Test API
 
 services:
+    api.name_converter:
```
dunglas: Class parameters are deprecated and should no longer be used.

Parser/DunglasApiParser.php
```diff
@@ -101,7 +104,11 @@ private function parseClass(ResourceInterface $resource, $entityClass, $io)
                 ($attributeMetadata->isReadable() && self::OUT_PREFIX === $io) ||
                 ($attributeMetadata->isWritable() && self::IN_PREFIX === $io)
             ) {
-                $data[$attributeMetadata->getName()] = $this->parseAttribute($resource, $attributeMetadata, $io);
+                $attributeName =  $attributeMetadata->getName();
```
dunglas: one extra space after the equal sign

dunglas:
Looks good to me.

@willdurand are you ok to tag a new release when all @ogizanagi changes will be merged?

Tests/Fixtures/app/config/dunglas_api.yml
ogizanagi: What do you mean ? I'm not using a %class% parameter, but declaring the `api.name_converter` service, used by your bundle if it exists.

Parser/DunglasApiParser.php
ogizanagi: good catch

Tests/Fixtures/app/config/dunglas_api.yml
dunglas: I've read to fast sorry.

PR #656, opened 2015-06-22
Prevent error "Undefined index" when no subtype provided

soyuka:
I don't know if this is intended or not, but `$info['subType']` was not defined after a clean installation. I'm using NelmioApiDocBundle with DunglasApiBundle.


willdurand:
thanks

PR #665, opened 2015-07-01
Add support for JsonSerializable classes

mcfedr:
Adds a parser that adds information from the JSON generated from classes that are implement `JsonSerializable`

Also allows that same `input={"class"="Some\Class", "name"="class"}` syntax that can be used for forms to with data from JsonSerializable and validation information


willdurand:
thank you :+1:

PR #675, opened 2015-07-16
Move the new parameters button so it's available when no parameters are specified

jhogendorn:
It's just handy/useful to have this button available at all times in case you need to throw extra params through.

### DIFF
--- a/Resources/views/method.html.twig
+++ b/Resources/views/method.html.twig
@@ -241,8 +241,8 @@
                                         </p>
                                     {% endif %}
                                     {% endfor %}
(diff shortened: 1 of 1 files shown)


jhogendorn:
Hey @willdurand just curious why this isn't worth merging? :)

willdurand:
It is worth merging ;-)  but please read this: https://github.com/nelmio/NelmioApiDocBundle/releases/tag/2.10.0

jhogendorn:
Oh, completely understandable!

We heavily utilise the api doc here, so I will see about trying to contribute back and actively develop on the bundle more.

PR #681, opened 2015-07-21
Fix CachingApiDocExtractor caches only first accessed view

munkie:
This PR fixes issue when caching is enabled and _CachingApiDocExtractor_ saves cache only for the first view accessed. When other api doc views are requested same cache file with api doc data is used instead of creating and using different cache files for different views.

### DIFF
--- a/Extractor/CachingApiDocExtractor.php
+++ b/Extractor/CachingApiDocExtractor.php
@@ -28,19 +28,14 @@ use Symfony\Component\Routing\RouterInterface;
 class CachingApiDocExtractor extends ApiDocExtractor
 {
     /**
-     * @var \Symfony\Component\Config\ConfigCache
-     */
-    protected $cache;
-
-    /**
      * @var string
      */
-    protected $cacheFile;
+    private $cacheFile;
 
     /**
      * @var bool
      */
-    protected $debug;
+    private $debug;
 
     /**
      * @param ContainerInterface $container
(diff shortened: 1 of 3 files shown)


Extractor/CachingApiDocExtractor.php
```diff
@@ -67,12 +92,23 @@ public function all($view = ApiDoc::DEFAULT_VIEW)
...
-        return unserialize(file_get_contents($this->cacheFile));
+        return unserialize(file_get_contents($cache));
+
+    }
 
+    /**
+     * @param string $view
+     * @return ConfigCache
+     */
+    protected function getViewCache($view)
```
stof: should be private IMO

munkie: And what about cacheFile and debug properties should they became private ?

munkie: @stof I've changed visibility to private, is it ok to be merged ?

munkie:
@willdurand Can you check this PR ?

willdurand:
thanks @munkie

oktapodia:
@munkie This pull request broke the tests :(

munkie:
@oktapodia Actually tests were broken before i submitted this PR. You can check Travis-CI build history - https://travis-ci.org/nelmio/NelmioApiDocBundle/builds and see that tests were broken since this PR merge - https://travis-ci.org/nelmio/NelmioApiDocBundle/builds/70353048

oktapodia:
Oh yes sorry !

PR #685, opened 2015-07-28
Prevent BC break in DunglasApiBundle

dunglas:
* Use the last beta version of ApiBundle to avoid BC break that will be merged soon in master
* Use the stable version of the Symfony Serializer Component

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,10 +31,10 @@
         "symfony/yaml": "~2.3",
         "symfony/form": "~2.3",
         "symfony/finder": "~2.3",
-        "symfony/serializer": "~2.7@dev",
+        "symfony/serializer": "~2.7",
         "friendsofsymfony/rest-bundle": "~1.0",
         "jms/serializer-bundle": ">=0.11",
-        "dunglas/api-bundle": "dev-master",
+        "dunglas/api-bundle": "~1.0@beta",
         "sensio/framework-extra-bundle": "~3.0",
         "symfony/phpunit-bridge": "~2.7"
     },


willdurand:
thanks!

teohhanhui:
Where do we go from here? What should a user like myself do in the meantime (before you have a 2.0 version of DunglasApiBundle)?

teohhanhui:
For now it seems I only have to change

```php
use Dunglas\ApiBundle\Mapping\ClassMetadataFactoryInterface;
```

to

```php
use Dunglas\ApiBundle\Mapping\Factory\ClassMetadataFactoryInterface;
```

in

```
Nelmio\ApiDocBundle\Extractor\AnnotationsProvider\DunglasApiProvider
Nelmio\ApiDocBundle\Parser\DunglasApiParser
```

But what's stopping these from living in DunglasApiBundle itself?

dunglas:
@teohhanhui mainly tests.
From now BC break are limited but they will become larger soon: https://github.com/dunglas/api-platform/issues/9

teohhanhui:
@dunglas Perhaps we should maintain a fork of NelmioApiDocBundle that works with the dev version of DunglasApiBundle?

dunglas:
Why not. A PR on NelmioApiDoc updated in sync with DunglasApiBundle that will be merged when the refactoring is done should do the trick.

teohhanhui:
I guess it's probably not the aim then, for the future version of NelmioApiDocBundle to remain compatible with version 1.0 series of DunglasApiBundle?

dunglas:
All breaks will be done before the release of DunglasApiBundle 1.0 (we will be renamed `api-platform/core` btw).

teohhanhui:
Uhh, I didn't think of it that way. I suppose it's not a violation of semantic versioning to break BC at the beta stage...

dunglas:
Yes it's OK before the first stable version and it's why we want to do these changes now.
And anyone using for instance DunglasApiBundle 1.0 beta 3 will still be able to use NelmioApiDoc 2.9.

PR #686, opened 2015-07-30
[WIP] Compatibility with DunglasApiBundle dev version

teohhanhui:
Keep up with the BC breaks

### DIFF
--- a/Extractor/ApiDocExtractor.php
+++ b/Extractor/ApiDocExtractor.php
@@ -220,6 +220,12 @@ class ApiDocExtractor
                 $class = ClassUtils::getRealClass(get_class($this->container->get($controller)));
                 $this->container->leaveScope('request');
             }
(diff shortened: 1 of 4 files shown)


teohhanhui:
@dunglas As discussed in #685

Extractor/ApiDocExtractor.php
```diff
@@ -220,6 +220,12 @@ public function getReflectionMethod($controller)
                 $class = ClassUtils::getRealClass(get_class($this->container->get($controller)));
                 $this->container->leaveScope('request');
             }
+        } elseif ($this->container->has($controller)) {
+            $method = '__invoke';
```
teohhanhui: We should perhaps fix the other methods to not always expect `\ReflectionMethod` instead, since this is useless?

teohhanhui:
Closing in favour of dunglas/DunglasApiBundle#225

PR #694, opened 2015-08-11
revert fix "LINK workaround for firefox"

### LINKED ISSUE #621: Sandbox - LINK is sent via POST with _method parameter (UNLINK is not)
Sending a LINK request is done via [method faking](http://symfony.com/doc/current/cookbook/routing/method_parameters.html#faking-the-method-with-method), so it sends a POST.
I believe it is a bug, since UNLINK and other methods are used directly.
Also, since LINKing is used on a non-collection resource like /something/1, the POST method would be disabled on it and a '405 Method Not Allowed' error is thrown.

If it is because of the [firefox bug](http://stackoverflow.com/questions/9649972/firefox-overwrites-custom-http-method-link-to-mixed-case-link-on-xhr), that one seems to be [fixed](https://bugzilla.mozilla.org/show_bug.cgi?id=477578) already.

tonivdv:
Reverts a previous fix for Firefox which is causing issues with Chrome and which is not necessary anymore for Firefox because they fixed the issue (https://bugzilla.mozilla.org/show_bug.cgi?id=477578)

Also Fixes #621

### DIFF
--- a/Resources/views/layout.html.twig
+++ b/Resources/views/layout.html.twig
@@ -471,7 +471,6 @@
 
 
 
-
                     // retrieve the additional headers to send
                     $('.headers .tuple', $(this)).each(function() {
                         var key, value;
--- a/Resources/views/layout.html.twig
+++ b/Resources/views/layout.html.twig
@@ -520,12 +520,6 @@
                     }
                     {% endif %}
 
-                    // Workaround for Firefox bug and a thereby resulting nginx incompatibility
-                    if (method == "LINK") {
-                        method = "POST";
-                        params._method = "LINK";
-                    }
-
                     // prepare final parameters
                     var body = {};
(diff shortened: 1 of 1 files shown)


tonivdv:
Hm, not sure though why tests are failing ... any way to relaunch tests without committing something? /cc @willdurand

tonivdv:
@willdurand I don't get why those tests fail for a change in the twig file? If you can help me out and can fix it!?

willdurand:
@tonivdv I tracked the error. It appears since v2.7 as things related to `ChoiceList` have changed (cf. https://github.com/symfony/symfony/commit/03efce1b568379eac21d880e427090e43035f505)

willdurand:
Well, actually there are different issues... Fixing (more or less) them one by one.

tonivdv:
:+1:

PR #695, opened 2015-08-11
Fix ApiDocExtractor to accept callable classes as controllers

ogizanagi:
This patch allows the parser to properly obtain the `\ReflectionMethod` from a controller when declared as a service not specifying a method name but using a callable class instead.

http://symfony.com/doc/current/cookbook/controller/service.html#defining-the-controller-as-a-service
> 2.6: If your controller service implements the __invoke method, you can simply refer to the service id (app.hello_controller). 

Also, related to https://github.com/dunglas/DunglasApiBundle/issues/219

Failing tests aren't related AFAIK.


dunglas:
:+1:

willdurand:
thanks

PR #705, opened 2015-09-08
[WIP] Swagger 2.0 support

bezhermoso:
#### Progress

* [x] Configuration
* [ ] Formatter implementation (WIP)
* [ ] Tests

Hi all,

Here is my rewrite of crude Swagger 2.0 formatter I wrote months ago. Previous implementation was simply a decorator around the existing Swagger 1.x formatter, but I wasn't quite satisfied with it for a number of reasons (basically it wasn't a good idea in the first place). This implementation no longer depends on the existing formatter, although it re-uses some of the core logic. This is far from complete, but wanted to create this PR so that we have a place to discuss certain aspects. There is a good amount of interest shown, it would be great if we can discuss it here.

For example...

I opted to have a separate configuration block for Swagger 2 (`nelmio_api_doc > swagger2`). Reason being it would be easier to configure it this way without having to rely on legacy formats dictated by Swagger 1.x. Would love your guys thoughts on this.


andreaslarssen:
Thanks for getting back on this, and the work this far. I think this is the right way to go as it maintains backwards compatibility

piotrantosik:
Hi!

Any progress? For me separate configuration is :+1:

DependencyInjection/Swagger2ConfigCompilerPass.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
```
piotrantosik: Invalid comment, should something like:
`This file is part of the NelmioApiDocBundle....`
https://github.com/nelmio/NelmioApiDocBundle/blob/master/EventListener/RequestListener.php#L4

nassuf:
Hi @bezhermoso 
Thanks for all the job you have done !
Is there any progress ?

bezhermoso:
@nassuf Just got back from vacation. Will be able to work more on this soon once I catch up to work. Apologies for the constant delays.

andreaslarssen:
?

willdurand:
@andreaslarssen please read this: https://github.com/nelmio/NelmioApiDocBundle/releases/tag/2.10.0.

andreaslarssen:
I see. @bezhermoso: Think this will change your plans for the Swagger 2.0 support?

willdurand:
To me it means more time to come up with a PR

andreaslarssen:
I'm sorry, but I'm not entirely sure what you mean by that.

willdurand:
This PR will get merged at some point, it is important. But I'll merge it when the bundle will be more stable.

andreaslarssen:
I see. Do you have a time estimate for when you'll consider the bundle more stable? 

@bezhermoso: I have to have a swagger 2.0 bundle ready soon, so I'm considering starting from scratch, and build a new bundle. What's your thoughts?

willdurand:
In a few weeks I'd say, given how many installs per day this bundle gets. It should be fairly quick :)

I'd be happy to merge Swagger 2.0 support here if you don't mind. We can probably create a dedicated branch for that, if you can take care of it :)

andreaslarssen:
A few weeks could work. I can help, no problem, but I think it's better if @bezhermoso leads, as he's already started the work.

bezhermoso:
@willdurand  @andreaslarssen A separate branch sounds good, perhaps branching off from `sf-2.8` branch. Being honest one of the reason that keeps me off from working more is the failing tests (and from a glance, it starts getting harder to get a handle of that because of legacy support). I just can't find the extra time to do work on that front with work and all. You are probably in the same boat. I'd say we fix that first before we proceed with anything else. Let me know if you need help and I can jump in on a few stuff if needed.

willdurand:
@bezhermoso I've fixed many stuff lately. Deprecation notices put apart, everything should work as expected :)

andreaslarssen:
OK, so @bezhermoso, everything seems to work as exptected, and you'll branch out? Fill us in on the progress, and we can see if there's anything we could do to help.

bezhermoso:
Just rebased work on top of current master. How do I go about making a branch here?

hutchic:
@bezhermoso you'll probably want to branch off a fork and PR to here

alexspark21:
when you push on the master your changes about swagger 2  ?

PR #729, opened 2015-10-13
Twig_Filter_Method is deprecated since 1.12, is replaced by Twig_Simp…

damien-thiesson:
…leFilter


damien-thiesson:
I remove this PR, already fixed in a queued PR.

Issue #739, opened 2015-10-27
All parameters are merged from bottom to top, but parameters option in ApiDoc annotation is ignored


fstr:
Currently it's possible to add the "parameters" option to the @ApiDoc annotation.
```
     * @ApiDoc(
     *      views = { "default"},
     *      authentication=true,
     *      input={
     *          "class" = "\MyEntity",
     *          "groups" = {"put"}
     *      },
     *      parameters={
     *          {"name"="name", "dataType"="integer", "requirement"="foobar", "format"="integer", "description"="this will be the new description"}
     *      },
     * )
```

Since this setting is at the top most level, it should overwrite previous settings for the given parameter.

Currently this parameters setting has no effect if there are more specific settings in MyEntity.

I would propose to merge the ApiDoc parameters option in the end on top of all other settings.

A use case would be, if you have an internal type (custom JMS Type for example) that you want to shield from the API docs and just show it as a simple array or something in the docs.

willdurand:
Wanna work on this?

PR #740, opened 2015-10-27
[ISSUE-739] ApiDoc parameters setting will override lower parameter definitions in the hierarchy

fstr:
Patch for #739

### DIFF
--- a/Extractor/ApiDocExtractor.php
+++ b/Extractor/ApiDocExtractor.php
@@ -322,6 +322,8 @@ class ApiDocExtractor
                 });
             }
 
+            // merge parameters with parameters block from ApiDoc annotation in controller method
+            $parameters = $this->mergeParameters($parameters, $annotation->getParameters());
             $annotation->setParameters($parameters);
         }
 
--- a/Tests/Extractor/ApiDocExtractorTest.php
+++ b/Tests/Extractor/ApiDocExtractorTest.php
@@ -371,4 +371,69 @@ class ApiDocExtractorTest extends WebTestCase
         $this->assertTrue(is_array($data));
         $this->assertCount($count, $data);
     }
+
+    public function testOverrideJmsAnnotationWithApiDocParameters()
+    {
+        $container  = $this->getContainer();
+        $extractor  = $container->get('nelmio_api_doc.extractor.api_doc_extractor');
+        $annotation = $extractor->get(
+            'Nelmio\ApiDocBundle\Tests\Fixtures\Controller\TestController::overrideJmsAnnotationWithApiDocParametersAction',
+            'test_route_27'
+        );
+
+        $this->assertInstanceOf('Nelmio\ApiDocBundle\Annotation\ApiDoc', $annotation);
+
+        $array = $annotation->toArray();
+        $this->assertTrue(is_array($array['parameters']));
+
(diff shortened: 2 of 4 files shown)


fstr:
I'm using the parameters setting which is read from the ApiDoc annotation and merge it in the end on top of all other parameters definitions that have been read from lower annotations in the hierarchy.

Sadly I can't prove in my test that I override an already read parameter definition, because they are empty. In my production code the JMSMetaDataParser will already fill the annotation::parameters property.

willdurand:
that was fast :p

fstr:
I already wrote it but wanted to create an issue about it first.

willdurand:
> Sadly I can't prove in my test that I override an already read parameter definition, because they are empty. In my production code the JMSMetaDataParser will already fill the annotation::parameters property.
> 
> I'm not sure why they are empty in the tests. Since the handler is registered.

arf, that's a bit sad indeed. could you look at why there is no information available from the JMSMetaDataParser?

fstr:
I think this needs some more improvement to work with nested objects.

fstr:
Added a new test which tests and shows how to use the properties ApiDoc setting on nested objects.
It's a bit cumbersome, but that way it is conform with the way the ApiDoc annotation is read and how it can be merged using the existing mergeParameters method.

```
    /**
     * @ApiDoc(
     *     output="Nelmio\ApiDocBundle\Tests\Fixtures\Model\JmsTest",
     *     input={
     *         "class" = "Nelmio\ApiDocBundle\Tests\Fixtures\Model\JmsTest"
     *     },
     *     parameters={
     *          {
     *              "name"="number",
     *              "dataType"="integer",
     *              "actualType"="string",
     *              "subType"=null,
     *              "required"=true,
     *              "description"="This is the new description",
     *              "readonly"=false,
     *              "sinceVersion"="v3.0",
     *              "untilVersion"="v4.0"
     *          },
     *          {
     *              "name"="arr",
     *              "dataType"="object (ArrayCollection)"
     *          },
     *          {
     *              "name"="nested",
     *              "dataType"="object (JmsNested)",
     *              "children": {
     *                  "bar": {
     *                      "dataType"="integer",
     *                      "format"="d+"
     *                  }
     *              }
     *          }
     *     }
     * )
     */
```

willdurand:
great. if tests are all green, could you please squash your commits?

fstr:
This patch would enable people to overwrite the parameters in the docs in their controller method ApiDoc annotation. The parameters only represent the "input".

Would it be interesting to have the same feature for the "output"?
Something like responseParameters or returnParameters, which enables users the overwrite the Return parameters that are listed in the generated API docs.

Right now I am able to tell developers that use my API that I expect an array in my PUT request (which is internally a CustomFooCollection), but for my GET request where the developers fetch the same resource, I have no possibility to tell them I return a simple array (which is internally a CustomFooCollection) and currently represented in my Return parameter list as CustomFooCollection.

If this would be interesting for the project, I'm willing to provide a separate patch for it.

willdurand:
> Would it be interesting to have the same feature for the "output"?

I'm not sure about it. Let's postpone this idea.

fstr:
Alright!

I squashed the commits into one feature commit.

willdurand:
thanks!

PR #743, opened 2015-10-28
Add: Possibility to override parameters (#739)

### LINKED ISSUE #739: All parameters are merged from bottom to top, but parameters option in ApiDoc annotation is ignored

BenjaminPaap:
This fixes #739 and enables `parameters` and `input` option to be used simultaneously.

### DIFF
--- a/Extractor/ApiDocExtractor.php
+++ b/Extractor/ApiDocExtractor.php
@@ -322,6 +322,11 @@ class ApiDocExtractor
                 });
             }
 
+            $parameters = $this->mergeParameters(
+                $parameters,
+                $annotation->getParameters()
+            );
+
             $annotation->setParameters($parameters);
         }
 


BenjaminPaap:
Had this done yesterday too but did not see the already opened pr for this.

PR #748, opened 2015-11-09
Empty string is now a valid default parameter value

fstr:
Empty string is now a valid default parameter value that is successfully merged with other parameters
Preventing access of uninitialized array key 'default'
Added functional test case for mergeParameters that covers this issue


willdurand:
Thanks!

PR #749, opened 2015-11-09
Introduced default-value parameter to JsonSerializableParser

fstr:
JSP is now able to set a default value for scalar types
This makes merging the parser result with other parser results easier
and gives some additional info when creating docs

### DIFF
--- a/Parser/JsonSerializableParser.php
+++ b/Parser/JsonSerializableParser.php
@@ -59,7 +59,8 @@ class JsonSerializableParser implements ParserInterface
             'subType' => null,
             'required' => null,
             'description' => null,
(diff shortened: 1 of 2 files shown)


Parser/JsonSerializableParser.php
```diff
@@ -49,6 +49,10 @@ public function parse(array $input)
         return $parsed['children'];
     }
 
+    /**
```
willdurand: No need for this comment block

willdurand:
@fstr perfect! could you please squash your commits now?

fstr:
Done :)

willdurand:
perfect, thanks!

PR #750, opened 2015-11-10
Fix route host with placeholder

piotrantosik:
I have a routing file:
```yml
api_v1:
    resource: "@ApiBundle/Resources/config/routing.yml"
    prefix:   /v1
    host:     api.{domain}
    defaults:
        domain: "%domain%"
    requirements:
        domain: "%domain%"
```

Currently documentation looks like this:
route

After this PR placeholder is replacement by default parameter:

route2

### DIFF
--- a/Annotation/ApiDoc.php
+++ b/Annotation/ApiDoc.php
@@ -476,6 +476,11 @@ class ApiDoc
 
         if (method_exists($route, 'getHost')) {
             $this->host = $route->getHost() ? : null;
+
+            //replace route placeholders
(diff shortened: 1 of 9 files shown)


willdurand:
Does not work well on all supported Symfony versions

piotrantosik:
Just noticed, I'll try to fix it asap

piotrantosik:
@willdurand fixed, also remove duplicated assertions

willdurand:
Perfect! Thank you.

PR #768, opened 2015-12-15
Check for deprecated method enterScope() in DumpCommand

SHyx0rmZ:
The dump command uses ContainerInterface::enterScope, which was deprecated in Symfony 2.8 and removed in 3.0.

### DIFF
--- a/Command/DumpCommand.php
+++ b/Command/DumpCommand.php
@@ -61,7 +61,7 @@ class DumpCommand extends ContainerAwareCommand
             $formatter->setEnableSandbox(false);
         }
 
-        if ('html' === $format) {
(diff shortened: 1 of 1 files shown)


gcds:
+1

da-wen:
+1

willdurand:
Does the command still work with this fix?

da-wen:
Yes, it works. I fixed it yesterday in my vendor code and the html will be generated as usual.

willdurand:
perfect then. thanks!

willdurand:
This fix will be shipped in 2.11.2.

da-wen:
thx a lot.

PR #781, opened 2016-01-08
Add textarea type for sandbox

AlmogBaku:
Adding option to send data using textarea.

* This is very useful when you want to send long text.


willdurand:
thank you

PR #792, opened 2016-01-26
Fix #565

### LINKED ISSUE #565: Required form fields for PUT method always false?
Is there a reason why for the PUT request the required form fields are always false, while for the POST / PATCH methods they reflect what is specified in the form class?

If this has to do with the idempotence of the PUT method, my inclination was to do it the other way round (POST/PATCH does not require all the form fields while PUT does)... 

PUT request:
put

POST / PATCH requests:
post

patch

rodrigm:
Fix #565 

---

Required fields will be false for `PATCH` routes instead of `PUT`

### DIFF
--- a/Extractor/ApiDocExtractor.php
+++ b/Extractor/ApiDocExtractor.php
@@ -322,8 +322,8 @@ class ApiDocExtractor
             $parameters = $this->clearClasses($parameters);
             $parameters = $this->generateHumanReadableTypes($parameters);
 
-            if ('PUT' === $annotation->getMethod()) {
-                // All parameters are optional with PUT (update)
+            if ('PATCH' === $annotation->getMethod()) {
+                // All parameters are optional with PATCH (update)
                 array_walk($parameters, function ($val, $key) use (&$parameters) {
                     $parameters[$key]['required'] = false;
                 });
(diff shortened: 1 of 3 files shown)


Padam87:
:+1:

rodrigm:
Sorry, can you give feedback to this?

willdurand:
Could you please squash your commits? thank you

rodrigm:
Done!

willdurand:
thank you!

rodrigm:
:+1

joserobleda:
:heart_eyes:

devantoine:
Doesn't seem to work as expected:

nelmiopatch

rodrigm:
Are you in the correct version?

I've tested and worked fine

devantoine:
I'm in 2.11 and I can see your edit in my vendors.

rodrigm:
It's strange, be sure to delete cache, etc.

devantoine:
Already did.

The input parameter is set to a form type, is it working with form types too?

rodrigm:
Maybe, the form fields have priority

devantoine:
That's it. If I use the entity as the input parameter all the fields requirement are set to false. But it's not possible with a FormType then?

rodrigm:
I think also it should be well.

devantoine:
SO what am I missing here? Do you have to set all my form's fields to required => false?

rodrigm:
Yes, I think being a PATCH should ignore the form and put all `required => false`

devantoine:
So a new PR is needed right?

rodrigm:
Yeah, it's a new issue/PR

PR #794, opened 2016-01-29
Remove incorrect usage of hasOption

mcfedr:
`hasOption` is confusingly named, it actually checks that the option has been created, not if the user has passed the option. So in this case its always true.
Also changed a weird usage of `in_array`.


willdurand:
thank you

PR #795, opened 2016-01-29
Add option to DumpCommand to filter parsed routes by keyword

InputOutput:
This is minor change to allow control over which routes are to have their annotations parsed when running the DumpCommand. A simple keyword can be passed using the route-filter option. This will result in a subset of routes to be extracted.

This change is requested for two reasons:
- the dumpCommand runs out of memory when trying to generate HTML documentation on our codebase
- we'd like the option to automatically split the generated documentation into smaller chunks

NB: I did not find any tests for the DumpCommand so I haven't updated any tests.


willdurand:
I cannot accept new features for now, sorry :(

PR #796, opened 2016-02-01
Fix Array to string conversion error on array requirements

### LINKED ISSUE #564: Notice: Array to string conversion when using array in @QueryParam requirements option
Hi guys,

I found a bug. While I'm trying to visit documentation pages I get an exception: 
```
Notice: Array to string conversion in /var/www/vhosts/localhost.crkl/httpdocs/vendor/nelmio/api-doc-bundle/Nelmio/ApiDocBundle/Extractor/Handler/FosRestHandler.php line 86

500 Internal Server Error - ContextErrorException 
```

My annotations:

```php
/**
 * @Rest\RouteResource("Venue")
 */
class VenueController extends FOSRestController
{
    /**
     * @param ParamFetcher $paramFetcher
     * @return array
     *
     * @Rest\View()
     * @Rest\QueryParam(name="lat", strict=true)
     * @Rest\QueryParam(name="lng", strict=true)
     * @Rest\QueryParam(name="range", strict=true, requirements={
     *     @Assert\GreaterThan("0"),
     *     @Assert\Regex("/\d+/")
     * }, default="10")
     * @Rest\QueryParam(name="product", strict=true, requirements={
     *     @Assert\Regex("/\d+/")
     * }, nullable=true)
     * @ApiDoc()
     */
    public function cgetAction(ParamFetcher $paramFetcher)
    {
        // [...]
    }
```

So, the reason of this error is an array in ```requirements``` option. After removing ```range``` QueryParam and change ```product``` QueryParam to:

```
@Rest\QueryParam(name="product", strict=true, requirements=@Assert\Regex("/\d+/"), nullable=true)
```

everyting works fine.

zhukovra:
It is completely fixes #564

### DIFF
--- a/Extractor/Handler/FosRestHandler.php
+++ b/Extractor/Handler/FosRestHandler.php
@@ -87,6 +87,34 @@ class FosRestHandler implements HandlerInterface
             return (string) $requirements['rule'];
         }
 
+        if (is_array($requirements) && array_key_exists(0, $requirements)) {
+
+            $output = array();
+
+            foreach ($requirements as $req) {
+
+                if (is_object($req) && $req instanceof Constraint) {
+                    if ($req instanceof Regex) {
+                        $output[] = $req->getHtmlPattern();
+                    } else {
+                        $class = get_class($req);
+                        $output[] = substr($class, strrpos($class, '\\')+1);
+                    }
+
+                }
+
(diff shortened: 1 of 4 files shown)


willdurand:
thank you

Issue #798, opened 2016-02-03
Wrong behaviour for "input" property when form has several inner custom types.


debesha:
Given I use next annotation
<pre>     *  input={
     *      "class"="CustomApiBundle\Form\Type\CustomOrderType",
     *      "name"=""
     *  }</pre>

and this CustomOrderType has two inner fields of custom types

<pre>        $builder->add("price", "custom_price");
        $builder->add("address", "custom_address");
</pre>

then it becomes parsed as a same type

<pre>
price	object (PriceType)	false		
price[paid]	string	true		
price[service_fee]	string	true		
price[deposit]	string	true		
price[total]	string	true		
address	object (CustomerAddressType)	false		
address[paid]	string	true		
address[service_fee]	string	true		
address[deposit]	string	true		
address[total]	string	true
</pre>

And if I change the order of fields to 

<pre>
        $builder->add("address", "custom_address");
        $builder->add("price", "custom_price");
</pre>

then 
<pre>address	object (CustomerAddressType)	false		
address[id]	string	true		
address[prefix]	string	true		
address[first_name]	string	true		
address[last_name]	string	true		
address[phone]	string	true		
price	object (PriceType)	false		
price[id]	string	true		
price[prefix]	string	true		
price[first_name]	string	true		
price[last_name]	string	true		
price[phone]	string	true</pre>

debesha:
PR with a fix https://github.com/nelmio/NelmioApiDocBundle/pull/799

dzubchik:
close this issue?

PR #799, opened 2016-02-03
Fixed bug with wrong form parsing when several custom inner types exist

debesha:
Fix to issue
https://github.com/nelmio/NelmioApiDocBundle/issues/798

There is a loop, and if there are few iterations, the previously defined $subForm variable causes wrong behaviour


dragosprotung:
+1 for this fix

cezarystepkowski:
+1 
Will this be merged soon?

willdurand:
thank you

PR #805, opened 2016-02-09
Fix Collection Type options - Symfony 2.8 Support

obense:
Bug has appeared when i migrate from symfony 2.8.1 to 2.8.2.
Must be related to https://github.com/symfony/symfony/pull/17162


spdionis:
:+1: I have the same problem while migrating to symfony 2.8.

willdurand:
thank you

Issue #801, opened 2016-02-04
"input" and "filters" annotations mutually exclusive


InputOutput:
When documenting POST/PUT operations with both GET query filters and input entities, the filters do not get parsed.

E.g.:
```
* @ApiDoc(
     *   resource=true,
     *   input="cruisetravel\BookingBundle\Entity\Cruise\CabinBooking",
     *   return="cruisetravel\BookingBundle\Entity\Cruise\CabinBooking",
     *   filters={
     *     {"name"="autoselect", "dataType"="boolean", "required"="false", "description"="Request a stepless booking"},
     *     {"name"="autoconfirm", "dataType"="boolean", "required"="false", "description"="If true, stepless booking will be automatically confirmed with provider, if false stepless booking will stop at priced booking."},
     *     {"name"="view", "dataType"="string", "required"="false", "description"="Basic, extended and special view providerPricing will attempt to price the requested selections before actual booking flow starts."}
     *   },
     *   parameters={
     *     {"name"="itineraryId", "dataType"="integer", "required"="true", "description"="itinerary id"},
     *     {"name"="cruiseSegmentId", "dataType"="integer", "required"="true", "description"="cruise segment id"}
     *   },
     *   statusCodes={
     *    200="Request received not stored, review and POST selection from returned options",
     *    201="Request stored, continue to PUT if bookingflow not finished",
     *    202="Request passed by proxy to external provider, review and POST or PUT selection from returned options",
     *    400="Bad Request",
     *    405="Method Not Allowed",
     *    422="Unprocessable Entity"
     *  }
     * )
     */
``` 

will not result in a "filters" section in the generated documentation.

I tracked this down to Annotation/ApiDoc.php:176 where changing 
```
if (isset($data['input'])) {
    $this->input = $data['input'];
} elseif (isset($data['filters'])) {
```

into

```
if (isset($data['input'])) {
    $this->input = $data['input'];
}

if (isset($data['filters']))
```

will fix the problem.

I am however uncertain why this restriction is even in place, and can find nothing in the comments or the documentation as to the reason why.

Please advise if this is a bug, if so I will prepare a PR. If not, please explain this behaviour.

PR #811, opened 2016-02-25
Allow filter descriptions to be used in conjunction with POST/PUT inp…

InputOutput:
…ut descriptions.

Fixes bug (?) as described in https://github.com/nelmio/NelmioApiDocBundle/issues/801


willdurand:
thanks!

PR #825, opened 2016-03-21
Mark symfony ^2.7.7 as conflict

willdurand:


stof:
some explanations about this conflict rule may be good for people using Symfony 2.7

willdurand:
I don't have much information unfortunately :cry:  The test suite fails for those versions, and I am not able to fix it.

PR #826, opened 2016-03-21
Fixed link to documentation

crazyball:
Link to ApiDoc documentation is now with ReST format instead of Markdown


willdurand:
ah good catch, thanks!

Issue #416, opened 2014-06-23
Not encode Json in filter Description


ivan1986:
*    filters= {
     *      {"name"="from", "dataType"="integer", "description" = "Таймштамп начала"},
     *      {"name"="to", "dataType"="integer", "description" = "Таймштамп окончания"}
     *    }
![1](https://cloud.githubusercontent.com/assets/156418/3356339/08e2a2ea-fac0-11e3-8b42-aaea12308bcb.jpg)
![2](https://cloud.githubusercontent.com/assets/156418/3356340/0ae12f08-fac0-11e3-8f5a-11574356f7ae.jpg)

VBelozyorov:
So... Is 
```
{{ value|json_encode(constant('JSON_UNESCAPED_UNICODE'))|replace({'\\\\': '\\'})|trim('"') }}
```
will be ok? @willdurand @stof

PR #775, opened 2015-12-28
Fix problem with cyrillic chars in docs

DonPrus:
You could use JSON_UNESCAPED_UNICODE, to fix problem with cyrillic chars in docs

PR #828, opened 2016-03-22
fix unicode

ivaaaan:
https://github.com/nelmio/NelmioApiDocBundle/issues/416


DonPrus:
Just the same as https://github.com/nelmio/NelmioApiDocBundle/pull/775, but all tests passed :)

PR #832, opened 2016-04-08
Allow sending empty file data

ismailbaskin:
When using Vich uploader bundle null file value is used for preventing existing resource. If there is no file value, resource will be deleted.


ismailbaskin:
@willdurand ping

willdurand:
thank you

Issue #525, opened 2014-10-04
Entities with validations ignore JMS serializer groups


alanhartless:
It seems that entity properties that have validations have their JMS serialization group ignored when filtered in the output parameter.

For example, here's a property in my entity

```
    /**
     * @ORM\Column(type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     * @Serializer\Expose
     * @Serializer\Since("1.0")
     * @Serializer\Groups({"userDetails", "userList"})
     */
    protected $id;

    /**
     * @ORM\Column(type="string", length=25, unique=true)
     * @Serializer\Expose
     * @Serializer\Since("1.0")
     * @Serializer\Groups({"userDetails", "userList"})
     */
    protected $username;

    /**
     * @ORM\Column(type="string", length=60, unique=true)
     * @Serializer\Expose
     * @Serializer\Since("1.0")
     * @Serializer\Groups({"userDetails"})
     */
    private $email;
```

In my controller's ApiDoc annotation, I have

```
    * ...
     * output={
     *      "class"="Acme\UserBundle\Entity\User",
     *      "groups"={"userList"}
     *   }
     *  ...
```

In the docs, email still shows up because I have in my loadValidatorMetadata function

```
$metadata->addPropertyConstraint('email',     new Assert\Email(
            array(
                'message' => 'Email invalid,
                'groups'  => array('SecondPass')
            )
        ));

        $metadata->addConstraint(new UniqueEntity(
            array(
                'fields'           => array('email'),
                'message'          => 'Email invalid',
                'repositoryMethod' => 'checkUniqueUsernameEmail'
            )
        ));
```

So....any suggestions on how I can get the api doc to not include the properties that match this scenario?

Thanks,
Alan

syzop:
Unproper patch (see comments at end):
```
--- a/Parser/JmsMetadataParser.php
+++ b/Parser/JmsMetadataParser.php
@@ -85,7 +85,7 @@ class JmsMetadataParser implements ParserInterface, PostParserInterface
     public function parse(array $input)
     {
         $className = $input['class'];
-        $groups    = $input['groups'];
+        $groups    = !empty($input['groups']) ? $input['groups'] : array("Default");

         return $this->doParse($className, array(), $groups);
     }
@@ -133,6 +133,7 @@ class JmsMetadataParser implements ParserInterface, PostParserInterface
                 // apply exclusion strategies
[...]

PR #835, opened 2016-04-14
Fix parameters return when JMS @Groups are used.

AlexandreHagen:
In special cases parameters returns do not match with JMS groups setting. 
See https://github.com/nelmio/NelmioApiDocBundle/issues/525#issuecomment-210201397

### DIFF
--- a/Parser/ValidationParser.php
+++ b/Parser/ValidationParser.php
@@ -61,6 +61,10 @@ class ValidationParser implements ParserInterface, PostParserInterface
      */
     public function supports(array $input)
     {
+        if ($input['groups']){
+            return false;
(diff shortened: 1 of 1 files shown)


willdurand:
Hey! would you be able to reveal the bug you're addressing with a test case?

AlexandreHagen:
Yes ! How do you want I do this ? I have to build a exemple in Symfony Bundle or just copy exemple code here ?

willdurand:
Could you write a test case in the test suite of this bundle?

AlexandreHagen:
Okay ! I don't know well Phpunit. But I will try.

PR #837, opened 2016-04-19
Added translation to description

timhovius:
The labels of a field are usually translated. But this does not happen in the documentation.


willdurand:
thank you!

timhovius:
Maybe you can create a new version? :smile:

PR #839, opened 2016-04-22
[RFR] Added cdn to prevent unexpected char

RomainMarecat:
api:doc:dump --format=html > api.html

console:

Uncaught SyntaxError: Unexpected token ILLEGAL
api.html:34363 Uncaught ReferenceError: $ is not defined

### DIFF
--- a/Resources/views/layout.html.twig
+++ b/Resources/views/layout.html.twig
@@ -8,9 +8,7 @@
         <style type="text/css">
             {{ css|raw }}
         </style>
-        <script type="text/javascript">
-            {{ js|raw }}
-        </script>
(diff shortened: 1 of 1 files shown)


Ma27:
:+1: 
but this file (https://github.com/nelmio/NelmioApiDocBundle/blob/2a0f95eac0ab2aa466545d0b21d15739636811b7/Resources/public/js/all.js) and this line (https://github.com/nelmio/NelmioApiDocBundle/blob/2a0f95eac0ab2aa466545d0b21d15739636811b7/Formatter/HtmlFormatter.php#L240) should be removed as they are not needed anymore, right?

RomainMarecat:
maybe... don't know what this file is doing with js file setted in globalvars ..

GuilhemN:
Feel free to reopen but know that 2.x is not the latest version anymore.

PR #851, opened 2016-05-16
Allow custom ApiDoc annotation

MartinLG:
When we override templates, we sometimes need project related informations to appear in the documentation. So the **ApiDoc** annotation is not sufficient : we need to add variables to this class.

Solution : create a child class of ApiDoc. Then we also need to create a child class of the **ApiDocExtractor** to override the **ANNOTATION_CLASS** constant. And here comes the problem.

`self::ANNOTATION_CLASS` does not support inheritance
`static:ANNOTATION_CLASS` does

I think it could be a good improvement to help creating more complete documentations, with more project related details.

Thanks.


willdurand:
thank you

Issue #842, opened 2016-05-02
@Security annotation on a controller doesn't appear in the generated doc


devantoine:
Following https://github.com/nelmio/NelmioApiDocBundle/pull/497 I'm wondering if the `@Security` annotation on a class is supposed to work? Because currently it does not.

If I had the `@Security` annotation directly on the method, it almost works, meaning the key icon appears next to the route with the "Needs authentication" alt text. I say almost, because when I had the role inside the annotation (`@Security("has_role('ROLE_FOO')")`), the role should appear in the alt of the key icon too.

I could manually set `authentication` and `authenticationRoles` on every method but it would be much easier to take into account the `@Security` annotation on the class.

What do you think? Is it already suppose to work? Is it doable? If so I could try to make a PR for that.

maxpou:
Hi @devantoine,
me too. It's only work on methods.

> If so I could try to make a PR for that.

:+1:

devantoine:
I've made this PR https://github.com/nelmio/NelmioApiDocBundle/pull/852 but I fail to pass the tests with some PHP versions and I don't know why.

ahungry:
Same here (local tests passed after I made a PR, but the remote tests fail).  Pretty sure its a misconfiguration on the CI

dbu:
version 3 is a complete rewrite. please check if this is still a problem with version 3 and a recent symfony version.

PR #852, opened 2016-05-18
Enable @Security controller annotation and parse roles from @Security

devantoine:
Related to https://github.com/nelmio/NelmioApiDocBundle/issues/842 this PR enable `@Security` annotation on a Controller and add authenticationRoles from `@Security` on a method.


devantoine:
I don't know why the build fails PHP 5.4, 5.5, 7 and hhvm. Any clue?

PR #853, opened 2016-05-19
Headers support

miholeus:
Sometimes it is neccessary to specify which headers will be accepted. These pull request solves that issue.
Headers are shown below parameters in their own section.
It is neccessary to specify only name for header.
``Description``, ``default``, ``required`` are optional parameters.
``Default`` parameter is used for sandbox, it will substitute value for header.
Screenshots:
https://www.evernote.com/l/AZCopjR4j4VDFLoXMmxY8HC9IBv0C5gJ7xY
https://www.evernote.com/l/AZB8tJgHiidCUJ0elPyYyctwtJssP4lTyPc

### DIFF
--- a/Annotation/ApiDoc.php
+++ b/Annotation/ApiDoc.php
@@ -47,6 +47,12 @@ class ApiDoc
      * @var array
      */
     private $parameters = array();
+    /**
+     * Headers that client can send.
+     *
+     * @var array
+     */
+    private $headers = array();
 
     /**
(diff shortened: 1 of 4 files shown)


willdurand:
Hi! Well done, I like it. Would you be able to set up a test case though?

miholeus:
Added tests

willdurand:
thank you!

PR #857, opened 2016-05-24
Add navigation index for resources

zanardigit:
When I have a very rich API with many methods, the navigation between sections involves a lot of scrolling.

So I propose to add an index at the top of the page. Of course this will also require some styling (in my case, I fixed it at the bottom right of the page).

Thank you for any comments that may improve this PR.


willdurand:
👍

PR #860, opened 2016-06-02
Fix wrapping of format table row

jupeter:
For parameter types "array of choices" when we have very long format values, it's visible like that: 
Before change

After this PR, it will be looks like that:
After change


willdurand:
Thanks

PR #861, opened 2016-06-02
Links from index.rst to the other doc files

Uplink03:
I've added links to index.rst so the other doc files can be accessible directly, instead of hunting them down by hand in the doc directory.

This should also fix http://symfony.com/doc/current/bundles/NelmioApiDocBundle/index.html which, without the links, is a bit useless.


willdurand:
thanks

PR #865, opened 2016-06-09
Fixed the links to some doc articles

javiereguiluz:
Some users have reported us that the links to other articles in this section are broken: http://symfony.com/doc/master/bundles/NelmioApiDocBundle/index.html#usage  I recommend to use a `toctree` directive because it's easier to maintain (for example the titles of the articles are updated automatically if they change).


willdurand:
thanks

Issue #864, opened 2016-06-07
Dump in html format will cause jquery issues


da-wen:
Before i create a pull request i want others to confirm #on that issues.
I think it happens only with newer versions of console component.

When you are dumping you api docs like this 
`bin/console api:doc:dump --env=dev --format=html > web/content-api.html`
 and open the generated html, the javascript is broken and gives you errors like this:
`content-api.html:646 Uncaught SyntaxError: Unexpected token ILLEGAL
content-api.html:4023 Uncaught ReferenceError: $ is not defined`

I checked the file located under _Resources/public/js/all.js_ and compared it with the generated file and figured out, that all **<<** were ripped out.

Here an example from online compare editor:
Base file: `X[V << 1] = Z++;`
Generated file: `X[V\1] = Z++;`

I went through the Formatter and everything was fine, but i recognized that the console->writeln is causing that issue.
I think console component has update some of their formatters.

A fix for that is to make the output raw.
Current: 
`$output->writeln($formattedDoc);`

Fixed:
`$output->writeln($formattedDoc, OutputInterface::OUTPUT_RAW);`

Gregoire-M:
Hi,
Same problem here with 2.12.0
Your fix works fine for me.

Gregoire-M:
@willdurand Could you tag a new version please?

willdurand:
I will but test suite is failing right now. Must be fixed first.

PR #866, opened 2016-06-10
Usage of OUTPUT_RAW to avoid javascript syntax error when dumping HTML

Gregoire-M:
A javascript syntax error occurs when an HTML doc is dumped, in the 2.12.0 tag. The syntax error is located in jQuery. It is seems to be caused by the console component.

It is fixed by specifying the `OutputInterface::OUTPUT_RAW` option when calling the `$output->writeln()` method.

More details in issue #864.


willdurand:
thanks!

Issue #502, opened 2014-08-31
Do not show "Requirement" when empty


JeroenDeDauw:
For the third filter here, no requirement has been defined. Can in such a case the requirement line be omitted?

![requirement](https://cloud.githubusercontent.com/assets/146040/4102728/a102c8d4-3139-11e4-8cc2-ddeadc7a2d56.jpg)

The filters are defines as follows:

```
 * @QueryParam(name="page", default="1", requirements="[1-9][0-9]{0,1}", description="Specifies the page in the result set")
 * @QueryParam(name="per_page", default="100", requirements="[1-9][0-9]{0,2}", description="The maximum number of items to receive per page")
 * @QueryParam(name="type", nullable=true, description="Only show items of this type.")
```

dbu:
version 3 is a complete rewrite. please check if this is still a problem with version 3 and a recent symfony version.

PR #868, opened 2016-06-12
#502 hide requirement when empty

dpcat237:
Patch for https://github.com/nelmio/NelmioApiDocBundle/issues/502 and a test for this case.


dpcat237:
Tests no pass in PHP5.5, PHP7 and PHPhhvm because of `no environment variable set` but it pass fine in PHP5.4 and PHP5.6.

willdurand:
thanks!

PR #871, opened 2016-06-15
[#870] Allow prioritization of extractor handlers

### LINKED ISSUE #870: Allow prioritization of extractor handlers
I have a case where I"m using one of [my own bundles](https://github.com/mickadoo/YamlApiDocAnnotationBundle) to load api doc configuration from YAML files. I want to then add another handler that will automatically add other parameters / requirements based on the annotation. My problem is that because handlers ignore priority my handler is always called _before_ the handler that loads annotation data is. 

I'll create a pull request that just copies that handling of priority from [symfony's framework bundle](https://github.com/symfony/symfony/blob/v3.1.1/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/AddCacheWarmerPass.php) although I see in the master branch that [there will be a trait for just this](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/DependencyInjection/Compiler/PriorityTaggedServiceTrait.php) as soon as it's tagged and released. For now though I'd just like to add support for prioritization

mickadoo:
Fixes #870

### DIFF
--- a/DependencyInjection/ExtractorHandlerCompilerPass.php
+++ b/DependencyInjection/ExtractorHandlerCompilerPass.php
@@ -22,13 +22,26 @@ class ExtractorHandlerCompilerPass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container)
     {
-        $handlers = array();
-        foreach ($container->findTaggedServiceIds('nelmio_api_doc.extractor.handler') as $id => $attributes) {
-            $handlers[] = new Reference($id);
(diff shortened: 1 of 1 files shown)


GuilhemN:
Feel free to reopen but know that 2.x is not the latest version anymore.

mickadoo:
Sure, it might be a nice feature to have an a lot of Symfony stuff they allow prioritization. Might be useful for someone else in the future, but it's not blocking me right now.

PR #880, opened 2016-07-06
Fix typo

pasdeloup:
Could also be replaced by _resource description_ or **resource description**


GuilhemN:
Fixed in https://github.com/nelmio/NelmioApiDocBundle/pull/915, thanks !

PR #907, opened 2016-09-08
Fixes not working anchors on website.

muchar:
On api/doc page, the anchors on top of the screen are not working on Safari/Chrome/Firefox if names of sections have spaces. Therefore, a number identifier should do the work and is independent from naming conventions used by users.


muchar:
I was thinking of using `{{ section|url_encode }}`, which also does the trick though.

PR #917, opened 2016-09-28
Add integer type to sandbox api

TIARG:
add integer type to sandbox api


Mislop:
@willdurand

Sir-G:
Okay, so I had a cheeky look as I planned to use the fork till the pull request issues are solved, and it seems that there's an issue in the method.html.twig. I seem to get the following error just as it tries to generate an html page when there are specified members in 'filters'.

```
  [Twig_Error_Runtime]                                                                                                            
  Key "format" for array with keys "dataType, description" does not exist in "NelmioApiDocBundle::method.html.twig" at line 274.  
```

PR #920, opened 2016-09-30
Ignore non reflectable methods

gpenverne:
When controller is generated dynamically by a factory , apidocbundle can't reflect method and throws exception.


Extractor/ApiDocExtractor.php
```diff
@@ -144,14 +144,17 @@ public function extractAnnotations(array $routes, $view = ApiDoc::DEFAULT_VIEW)
         foreach ($this->annotationsProviders as $annotationProvider) {
             foreach ($annotationProvider->getAnnotations() as $annotation) {
                 $route = $annotation->getRoute();
-                $array[] = array('annotation' => $this->extractData($annotation, $route, $this->getReflectionMethod($route->getDefault('_controller'))));
+                $reflectionMethod = $this->getReflectionMethod($route->getDefault('_controller'));
+                if (null !== $reflectionMethod) {
```
gpenverne: Here, check if method can be reflected.

kbsali:
:+1:

PR #935, opened 2016-12-20
Hotfix readme changed

jlpoveda:


GuilhemN:
What is this pr for ?

jlpoveda:
I'am adding HMAC authentication support in the sandbox. But in that moment is only a beta.

PR #937, opened 2016-12-29
[3.0] Add models support

GuilhemN:
This PR adds model support based on a ModelRegistry. 
The registry is injected in the describers implementing `ModelRegistryAwareInterface`. Then the describers register a schema in the registry, this schema will then be replaced by a reference to the model generated by the model describers:
```php
class MyDescriber implements DescriberInterface, ModelRegistryAwareInterface
{
    use ModelRegistryAwareTrait;

    public function describe(Swagger $api)
    {
        $schema = $api->getPaths()->get('/my-route/')
            ->getOperation('get')->getResponses()->get('200')->getSchema()
            ->setRef(
                $this->modelRegistry->register(new Model(new Type(Type::BUILTIN_TYPE_OBJECT, false, Dummy::class)))
            );
    }
}
```

will result to:
```yml
paths:
    /my-route/:
        get:
            responses:
                200: { $ref: '#/definitions/Dummy' }
definitions:
    Dummy:
        // ...
```

### DIFF
--- a/Model/ModelRegistry.php
+++ b/Model/ModelRegistry.php
@@ -20,103 +20,79 @@ use Symfony\Component\PropertyInfo\Type;
 
 final class ModelRegistry
 {
-    private $modelDescribers = [];
-    private $options;
     private $unregistered = [];
-    private $hashes = [];
+    private $models = [];
+    private $names = [];
+    private $modelDescribers = [];
+    private $api;
 
     /**
      * @param ModelDescriberInterface[] $modelDescribers
+     *
+     * @internal
      */
-    public function __construct(array $modelDescribers = [])
+    public function __construct(array $modelDescribers = [], Swagger $api)
     {
-        $this->options = new \SplObjectStorage();
         $this->modelDescribers = $modelDescribers;
+        $this->api = $api;
     }
 
-    /**
-     * @param Schema|Items $schema
-     */
-    public function register($schema): ModelOptions
+    public function register(Model $model): string
     {
-        if (!$schema instanceof Schema && !$schema instanceof Items) {
-            throw new \LogicException(sprintf('Expected %s or %s, got %s', Schema::class, Items::class, get_class($schema)));
+        $hash = $model->getHash();
+        if (isset($this->names[$hash])) {
+            return '#/definitions/'.$this->names[$hash];
         }
 
-        if (!isset($this->options[$schema])) {
-            $this->unregistered[] = $schema;
-            $this->options[$schema] = new ModelOptions();
-        }
+        $this->names[$hash] = $name = $this->generateModelName($model);
+        $this->models[$hash] = $model;
+        $this->unregistered[] = $hash;
 
-        return $this->options[$schema];
+        // Reserve the name
+        $this->api->getDefinitions()->get($name);
+
+        return '#/definitions/'.$name;
     }
 
     /**
      * @internal
      */
-    public function registerModelsIn(Swagger $api)
+    public function registerDefinitions()
     {
         while (count($this->unregistered)) {
             $tmp = [];
-            foreach ($this->unregistered as $schema) {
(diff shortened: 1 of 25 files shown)


Model/ModelRegistry.php
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        return $type->getBuiltinType();
+    }
+
+    private function typeToString(Type $type): string
+    {
+        if (Type::BUILTIN_TYPE_OBJECT === $type->getBuiltinType()) {
+            return $type->getClassName();
+        } elseif (Type::BUILTIN_TYPE_ARRAY === $type->getBuiltinType()) {
```
joelwurtz: maybe better to use $type->isCollection in this case ?

ModelDescriber/ObjectModelDescriber.php
```diff
@@ -0,0 +1,54 @@
...
+        $schema->setType('object');
+        $properties = $schema->getProperties();
+
+        $class = $options->getType()->getClassName();
+        foreach ($this->propertyInfo->getProperties($class) as $propertyName) {
+            $types = $this->propertyInfo->getTypes($class, $propertyName);
+            if (0 === count($types)) {
+                throw new \LogicException(sprintf('The PropertyInfo component was not able to guess the type of %s::$%s', $class, $propertyName));
+            }
+            if (count($types) > 1) {
```
joelwurtz: This can happens often, maybe not the place right now, but should be implemented in the future IMO

GuilhemN: I didn't find how to document multiple types with swagger and for objects [only the first type is actually supported by the Symfony serializer](https://github.com/symfony/symfony/blob/40280f203c2e1db2eb8ad74c178bc6840a7b93b9/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php#L245-L277) (as the ObjectNormalizer supports everything) so unless we find a better solution, I think it should be fine.
Maybe we can also just ignore the other types like the symfony serializer does.

Model/ModelRegistry.php
GuilhemN: yes indeed, done :+1:

GuilhemN:
@joelwurtz btw do you think we should keep jms serializer support ? AFAIK it is not well maintained but I think it is still used by some. Maybe we should allow configuring this at the project level (e.g. choose either the jms or the symfony serializer but not both)?

GuilhemN:
Updated to be make `ModelOptions` immutable and permit https://github.com/nelmio/NelmioApiDocBundle/issues/940.

ModelDescriber/ObjectModelDescriber.php
joelwurtz: you can pass an array of types normally : 

```
{
    "definitions": {
        "Foo": {
            "type": "object",
            "properties": {
                 "bar": { "type": ["string", "null", "numeric"] }
            }
        }
    }
}
```

joelwurtz:
> btw do you think we should keep jms serializer support ? AFAIK it is not well maintained but I think it is still used by some. Maybe we should allow configuring this at the project level (e.g. choose either the jms or the symfony serializer but not both)?

It's better maintained for a few months now as new people have taken the lead over it

ModelDescriber/ObjectModelDescriber.php
GuilhemN: @joelwurtz but it does only work for scalars, right ?

GuilhemN:
@joelwurtz then we'll have to re-add its support I guess, shouldn't be too complicated by adding a `PropertyTypeExtractorInterface`.

GuilhemN:
closing in favor of https://github.com/nelmio/NelmioApiDocBundle/pull/943 :)

PR #938, opened 2016-12-30
[3.0] Allow implicit operations with SwaggerPhp annotations

GuilhemN:
As @dbu suggested in [his article](https://blog.liip.ch/archive/2016/05/11/convert-nelmioapidocbundle-to-swagger-php.html), this PR allows to use SwaggerPhp annotations without specifying the path.
An example, before:
```php
/**
 * @Route("/foo", method="GET")
 * @SWG\Get(path="/foo")
 */
public function myAction()
{
}
```

After:
```php
/**
 * @Route("/foo", method="GET")
 * @SWG\Get()
 */
public function myAction()
{
}
```

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -2,7 +2,7 @@
 
 $finder = PhpCsFixer\Finder::create()
     ->in(__DIR__)
-    ->exclude('tests/Fixtures/app/cache')
+    ->exclude('Tests/Functional/cache')
 ;
 
 return PhpCsFixer\Config::create()
@@ -11,14 +11,14 @@ return PhpCsFixer\Config::create()
         'ordered_imports' => true,
         'phpdoc_order' => true,
         'header_comment' => [
-            'header' => <<<COMMENT
+            'header' => <<<HEADER
 This file is part of the NelmioApiDocBundle package.
 
 (c) Nelmio
 
 For the full copyright and license information, please view the LICENSE
 file that was distributed with this source code.
-COMMENT
+HEADER
         ],
     ])
     ->setFinder($finder)
--- a/DependencyInjection/Compiler/AddRouteDescribersPass.php
+++ b/DependencyInjection/Compiler/AddRouteDescribersPass.php
@@ -22,6 +22,6 @@ class AddRouteDescribersPass implements CompilerPassInterface
     {
         $routeDescribers = $this->findAndSortTaggedServices('nelmio_api_doc.route_describer', $container);
 
-        $container->getDefinition('nelmio_api_doc.describers.route')->replaceArgument(3, $routeDescribers);
+        $container->getDefinition('nelmio_api_doc.describers.route')->replaceArgument(2, $routeDescribers);
     }
 }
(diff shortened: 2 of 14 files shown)


GuilhemN:
I wonder whether we should allow something like:
```php
/**
 * @Response("200", description="foo")
 */
public function myAction();
```

It would be a good alternative to the `@ApiDoc` annotation.

dbu:
great @GuilhemN that you started working on this!

i did not review the code in detail, but i wonder if the ApiDoc annotations should even be kept if SWG is used for that in the future. having redundant annotations to do the same thing is confusing (and could create ambiguities if both are defined and contradicting). 

is the `@SWG\Response` annotation coming from swagger php? can there be more than one Response annotation on a method to list different status? is the idea that as soon as there is *any* SWG annotation, the resource is created with all methods that the symfony route configuration allows? if both answers are yes, i like the solution. if not we might need to think some more.

GuilhemN:
> i wonder if the ApiDoc annotations should even be kept if SWG is used for that in the future.

@dbu i'd prefer to not complicate too much the migration to `3.0` and keep `@ApiDoc` (but it can be not recommended and eventually be removed in 4.0).

> is the @SWG\Response annotation coming from swagger php?

Yes `@SWG\Response` is coming from swagger-php (it is the syntax used in its examples).

> can there be more than one Response annotation on a method to list different status?

Sure, that's the idea :)

> is the idea that as soon as there is any SWG annotation, the resource is created with all methods that the symfony route configuration allows?

`@Operation`, `@Response`, `@Parameter` and `@ExternalDocumentation` are detected and transmitted to an operation automatically created by the bundle based on the symfony routes. The other annotations are sub annotations of these annotations (they can't be directly in the method's doc block).

BTW I'm not a big fan of the management of the models by Swagger-php as you have to put an annotation in all of them. Maybe having our own annotation for `@Output` and `@Input` would be better.

dbu:
i wonder if the ApiDoc annotations should even be kept if SWG is used for
that in the future.

@dbu <https://github.com/dbu> i'd prefer to not complicate too much the
migration to 3.0 and keep @ApiDoc (but it can be not recommended and
eventually be removed in 4.0).


i would use the major version change to get rid of the unneeded
annotations. they are easy to batch replace in user code. having 2 for the
same is really confusing

BTW I'm not a big fan of the management of the models by Swagger-php as you
have to put an annotation in all of them. Maybe having our own annotation
for @Output and @Input would be better.

we should discuss this in a separate thread. maybe swaggerphp itself can be
improved too...

GuilhemN:
> i would use the major version change to get rid of the unneeded annotations. they are easy to batch replace in user code. having 2 for the same is really confusing

The question is whether `@ApiDoc` is not needed. Swagger-php has no mechanism to detect how a model is structured (that's normal as it's not framework-oriented) while `@ApiDoc` can: when defining `output` or `input`, the rest of the bundle will guess the model. 
For the rest, Swagger-php is verbose (maybe we could provide aliases such as `@QueryParameter`, `@PathParameter`, ...) but I think it is good enough to be adapted to the framework.

GuilhemN:
I opened https://github.com/nelmio/NelmioApiDocBundle/issues/939 to discuss whether we should remove `@ApiDoc`, it doesn't block us here ;)

GuilhemN:
I merge this to advance, you can still open an issue/PR if you see something to change.

Issue #947, opened 2017-01-18
[3.0]The SwaggerPhpDescriber only scan the kernel root dir.


magnetik:
Hi,

Line 8 of the `NelmioApiDocBundle/Resources/config/swagger_php.xml`, the project path is set as `%kernel.root_dir%` 

If I'm not wrong, if should scan the whole source folders, no?

GuilhemN:
Indeed, it should be changed :)
I'd like to make it configurable and default to `%kernel.root_dir%/../src`.

GuilhemN:
Fixed in https://github.com/nelmio/NelmioApiDocBundle/pull/949.

PR #949, opened 2017-01-19
Make source folder configurable

magnetik:
Fix for #947

The fact that the test are still passing with this change are the only test included :D

Maybe I should start testing the Extension?

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -22,6 +22,7 @@ final class Configuration implements ConfigurationInterface
         $treeBuilder
             ->root('nelmio_api_doc')
             ->children()
+                ->scalarNode('source_folder')->defaultValue('%kernel.root_dir%/../src')->end()
                 ->arrayNode('routes')
                     ->info('Filter the routes that are documented')
                     ->addDefaultsIfNotSet()
--- a/Tests/Functional/TestKernel.php
+++ b/Tests/Functional/TestKernel.php
@@ -62,6 +62,7 @@ class TestKernel extends Kernel
 
         // Filter routes
         $c->loadFromExtension('nelmio_api_doc', [
+            'source_folder' => '%kernel.root_dir%',
             'routes' => [
                 'path_patterns' => ['^/api(?!/admin)'],
             ],
--- a/DependencyInjection/NelmioApiDocExtension.php
+++ b/DependencyInjection/NelmioApiDocExtension.php
@@ -41,6 +41,9 @@ final class NelmioApiDocExtension extends Extension
         }
         if (class_exists(Swagger::class)) {
             $loader->load('swagger_php.xml');
+
+            $swaggerPHPDescriber = $container->getDefinition('nelmio_api_doc.describers.swagger_php');
+            $swaggerPHPDescriber->replaceArgument(0, $config['source_folder']);
         }
         if (interface_exists(ParamInterface::class)) {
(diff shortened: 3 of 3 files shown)


Tests/Functional/TestKernel.php
```diff
@@ -62,6 +62,7 @@ protected function configureContainer(ContainerBuilder $c, LoaderInterface $load
 
         // Filter routes
         $c->loadFromExtension('nelmio_api_doc', [
+            'source_folder' => '%kernel.root_dir%',
```
GuilhemN: `%kernel.root_dir%/../src`?

GuilhemN: I'd like the bundle to work without config in a standard symfony app.

magnetik: The default value is indeed `%kernel.root_dir%/../src` (set in `Configuration`), but the testing app/kernel in `src/Test/Functional` does not have a src folder. Should I create one?

GuilhemN: Indeed reviewed too quickly :)

GuilhemN:
Great, thanks !

DependencyInjection/Configuration.php
```diff
@@ -22,6 +22,7 @@ public function getConfigTreeBuilder()
         $treeBuilder
             ->root('nelmio_api_doc')
             ->children()
```
GuilhemN: BTW i wonder, shouldn't we always require swagger-php? Or otherwise maybe it is better to put this in a sub section `swagger_php`

magnetik: I was under the impression that the source_folder option might be used for multiple bundle, but I might be wrong.

GuilhemN: I don't see other use cases. 
In fact we could maybe even get ride of this option if we use our own parser with Swagger-Php annotations that would focus on controllers but that would mean losing support of Swagger-Php annotations for entities and only support `@Model`. Wdyt @dbu?

dbu: if we can use what swagger-php does, we should hard require that imho.

whether to use our own parser or not, i don't know. i was really surprised to see swagger-php having its own php parser code instead of using something like the doctrine annotations tool to do that... not a fan of "sharing" annotations with swagger-php but parsing them differently. maybe its easier to fork swagger-php and make the library clean and extensible to work well together with the bundle? and then maybe the swagger-php people are ok to take that library as a next major version so it does not have to be a permanent fork...

GuilhemN: > if we can use what swagger-php does, we should hard require that imho.

Will do.

> i was really surprised to see swagger-php having its own php parser code instead of using something like the doctrine annotations tool to do that...

Actually they use the doctrine parser to parse docblocks but they don't want to execute the code (which is forced by reflection) so they added a layer on top of it to make it static.

> maybe its easier to fork swagger-php and make the library clean and extensible to work well together with the bundle?

I don't think it's worth it, I prefer merging our efforts :)

> make the library clean and extensible to work well together with the bundle

It would be great but I don't think it's doable without loosing its flexibility to work with any project.

dbu: did not want to promote a split. all the better if you can get it working with swagger-php ;-)

GuilhemN:
What do you think of [this](https://github.com/nelmio/NelmioApiDocBundle/pull/952) workaround?

PR #951, opened 2017-01-20
Add support for bool type in ScalarModelDescriber

magnetik:
No sure if there is a reason why it was not there in the first place. 

It's useful to me =)


GuilhemN:
It looks like I forgot it. Thanks !

PR #952, opened 2017-01-23
[3.0][SwaggerPhp] Only parse controllers

GuilhemN:
Uses https://github.com/zircote/swagger-php/commit/fca06ec2b745d4eedabd41e5e0e0ebd8c4ac29dd to be able to parse constants. 
To avoid unexpected errors (annotations not existing), only controllers are parsed, that means that `@Schema` is no longer supported on entities, this is fine imo as there is `@Model`.
People also won't be able anymore to use root annotations (`@Info`, etc.) but I'd like to replace them by an array in the config that would be merged in the doc.

This also means that we no longer need `source_folder` as the files are detected using reflection.

Wdyt?


Tests/Functional/Controller/ApiController.php
```diff
@@ -41,15 +39,15 @@ public function swaggerAction()
      * @SWG\Response(
      *     response="201",
      *     description="Operation automatically detected",
-     *     @Model(type="Nelmio\ApiDocBundle\Tests\Functional\Entity\User")
+     *     @Model(type=User::class)
```
dbu: oh, does this work in annotations? very cool if it does!

Describer/SwaggerPhpDescriber.php
```diff
@@ -14,48 +14,68 @@
...
-        $this->operationResolver = $operationResolver;
+        $files = [];
+        foreach ($this->routeCollection->all() as $route) {
+            if (!$route->hasDefault('_controller')) {
+                continue;
+            }
+
+            // if able to resolve the controller
+            $controller = $route->getDefault('_controller');
+            if ($callable = $this->controllerReflector->getReflectionClassAndMethod($controller)) {
```
dbu: does this work for controllers as services as well?

Tests/Functional/Controller/ApiController.php
GuilhemN: Yes, that's what this pr fixes :)

Describer/SwaggerPhpDescriber.php
GuilhemN: Yes !

dbu:
cool! does this cover all controllers? it does handle controller as service you said. does it also handle autodetected controllers in AppBundle/Controller/XyzController that symfony autoregisters? it seems like a very good improvement to me to only scan controllers instead of the whole codebase.

what are the "root annotations"? tried to find them in the current doc, but did not find them. if they are on the controller class, you could use reflection to also get those? and keep track of which controller classes have already been seen... if its completely global then yes, sounds like it should go to the config.

GuilhemN:
> cool! does this cover all controllers? it does handle controller as service you said. does it also handle autodetected controllers in AppBundle/Controller/XyzController that symfony autoregisters? it seems like a very good improvement to me to only scan controllers instead of the whole codebase.

It is the same logic that is [used currently](https://github.com/nelmio/NelmioApiDocBundle/blob/dev/Util/ControllerReflector.php), so normally all cases are supported.

> what are the "root annotations"? tried to find them in the current doc, but did not find them. if they are on the controller class, you could use reflection to also get those? and keep track of which controller classes have already been seen... if its completely global then yes, sounds like it should go to the config.

I'm not sure how to call them. Swagger-php allows to have annotations anywhere in your codebase and you could have for example a `@SWG\Info(title="My First API", version="0.1")` annotation that would lead to something like:
```yml
info:
   title: My First API
   version: 0.1
```

with this PR, this annotation will only be parsed if it is in a controller (but it should not be done, I just can't prevent it).

dbu:
> I'm not sure how to call them. Swagger-php allows to have annotations
> anywhere in your codebase and you could have for example a
> |@SWG\Info(title="My First API", version="0.1")| annotation that would
> lead to something like:
> 
> info:
>    title: My First API
>    version: 0.1
> 
> with this PR, this annotation will only be parsed if it is in a
> controller (but it should not be done, I just can't prevent it).

okay, that should indeed be configuration and not an annotation. maybe
lets recommend using the configuration, and if both annotation and
config is present, let the annotation win. or even ignore the annotation
- though that would be unexpected for people used to swagger-php.

GuilhemN:
> okay, that should indeed be configuration and not an annotation. maybe lets recommend using the configuration, and if both annotation and config is present, let the annotation win. or even ignore the annotation

I think the configuration will be the base on which everything else is merged but will win against annotations.
Using these annotations should be discouraged anyway imo as they will only be parsed if the controller in which they are is registered somewhere.

PR #966, opened 2017-03-03
Only pass class to visited array in nested fields

stefanbraspenning:
When a class has multiple nested fields of the same class, the parser should not skip those. It should only skip a class if the class is a parent to avoid infinite recursion


GuilhemN:
Thanks for contributing!

PR #971, opened 2017-03-15
Do not use nullable result directly in ObjectModelDescriber

magnetik:
Not sure about the solution though: I'm just skipping the declaration but maybe we want to do something else.

### DIFF
--- a/ModelDescriber/ObjectModelDescriber.php
+++ b/ModelDescriber/ObjectModelDescriber.php
@@ -22,6 +22,8 @@ class ObjectModelDescriber implements ModelDescriberInterface, ModelRegistryAwar
 {
     use ModelRegistryAwareTrait;
 
+    private $propertyInfo;
+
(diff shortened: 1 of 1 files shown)


ModelDescriber/ObjectModelDescriber.php
```diff
@@ -33,7 +35,11 @@ public function describe(Model $model, Schema $schema)
         $properties = $schema->getProperties();
 
         $class = $model->getType()->getClassName();
-        foreach ($this->propertyInfo->getProperties($class) as $propertyName) {
+        $propertyInfoProperties = $this->propertyInfo->getProperties($class);
+        if (null === $propertyInfoProperties) {
+            return;
+        }
+        foreach ($propertyInfoProperties as $propertyName) {
```
GuilhemN: I feel like a space is missing between the two lines above

GuilhemN:
Thank you @magnetik !

Issue #965, opened 2017-03-02
[3.0] Support swagger tag


magnetik:
Hi,

If I'm not wrong, it seems that that the tags are added to the swagger document, but not added on routes.

Seems like a good addition

GuilhemN:
What do you mean?
You can add tags on routes using
```php
/**
 * @SWG\Get(
 *     tags={"foo", "bar"}
 * )
 */
public function myAction() {}
```

magnetik:
Since I'm using FOSRestBundle, I was not adding `@SWG\Get`, but I was adding `@SWG\Tag` on my action directly. 
Obviously at the moment this does not work, but repeating the method to declare the tag feels wrong to me.

GuilhemN:
@magnetik it doesn't seem to be meant for this so it would first require a PR on Swagger-php repo to support `@Tag` in `@Operation`.
But otherwise I like it :+1: maybe we'll be able to get ride of `@Operation`

GuilhemN:
In fact, it is logical [according to the spec](http://swagger.io/specification/).
`Operation::$tags` is an array of strings while `Swagger::$tags` is an array of `Tag`, so I don't think we can change something here.

magnetik:
Yeah so I have to redefine the method using `@SWG\Get` or `@SWG\Post`, not a big fan of this.

Don't we need our own annotation to define things like that?

I was also wondering about supporting view (as defined in the v2), does it make sense to you?

GuilhemN:
> Yeah so I have to redefine the method using @SWG\Get or @SWG\Post, not a big fan of this.
me neither

> Don't we need our own annotation to define things like that?
We could make `@Tag` work at the root of methods only for this bundle, we already have a very different semantic than Swagger-Php anyway.

> I was also wondering about supporting view (as defined in the v2), does it make sense to you?
I'm not sure of the benefit, do you have an use case?

daevidvincent:
[...]

PR #975, opened 2017-03-17
Support @SWG\Tag

GuilhemN:
Fix https://github.com/nelmio/NelmioApiDocBundle/issues/965

Theorically this works fine but I wonder whether `exsyst/swagger` prints a valid file for operation tags as they must not be arrays.


magnetik:
Looks fine to me. Not sure about your concern though.

GuilhemN:
Let's merge it as is :) I'll have to make sure the documentation dumped is valid though.

PR #976, opened 2017-03-20
[3.0] Handle FOSRestBundle QueryParam required

magnetik:
If the QueryParam is not nullable and strict, a 400 error will be raised if the query param is not raised, so it's required.


GuilhemN:
Thank you @magnetik !

PR #980, opened 2017-03-30
Fix deprecated

LordWeedlle:
I got some deprecated using NelmioApiDocBundle.
It appears that only non-legacy symfony requires formType class instead of formType instance.

Also, I couldn't load vendors to launch unit tests, any time I tried resulted in the following error:
```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - dunglas/api-bundle 1.x-dev requires phpdocumentor/reflection ^1.0.7 -> satisfiable by phpdocumentor/reflection[1.0.7].
    - dunglas/api-bundle v1.0.0 requires phpdocumentor/reflection ^1.0.7 -> satisfiable by phpdocumentor/reflection[1.0.7].
    - dunglas/api-bundle v1.0.0-beta requires phpdocumentor/reflection ^1.0.7 -> satisfiable by phpdocumentor/reflection[1.0.7].
    - dunglas/api-bundle v1.0.0-beta.2 requires phpdocumentor/reflection ^1.0.7 -> satisfiable by phpdocumentor/reflection[1.0.7].
    - dunglas/api-bundle v1.0.0-beta.3 requires phpdocumentor/reflection ^1.0.7 -> satisfiable by phpdocumentor/reflection[1.0.7].
    - dunglas/api-bundle v1.0.1 requires phpdocumentor/reflection ^1.0.7 -> satisfiable by phpdocumentor/reflection[1.0.7].
    - dunglas/api-bundle v1.1.0 requires phpdocumentor/reflection ^1.0.7 -> satisfiable by phpdocumentor/reflection[1.0.7].
    - dunglas/api-bundle v1.1.0-beta.1 requires phpdocumentor/reflection ^1.0.7 -> satisfiable by phpdocumentor/reflection[1.0.7].
    - dunglas/api-bundle v1.1.0-beta.2 requires phpdocumentor/reflection ^1.0.7 -> satisfiable by phpdocumentor/reflection[1.0.7].
    - dunglas/api-bundle v1.1.1 requires phpdocumentor/reflection ^1.0.7 -> satisfiable by phpdocumentor/reflection[1.0.7].
    - Conclusion: don't install phpdocumentor/reflection 1.0.7
    - Installation request for dunglas/api-bundle ~1.0@dev -> satisfiable by dunglas/api-bundle[1.x-dev, v1.0.0, v1.0.0-beta, v1.0.0-beta.2, v1.0.0-beta.3, v1.0.1, v1.1.0, v1.1.0-beta.1, v1.1.0-beta.2, v1.1.1].
```


Parser/FormTypeParser.php
```diff
@@ -268,7 +268,7 @@ private function parseForm($form)
                                 unset($subForm);
                             }
 
-                            if (LegacyFormHelper::hasBCBreaks()) {
```
LordWeedlle: There might be a reason I'm not aware of for this if statement, I can change my modification if needed to 
```php
if (LegacyFormHelper::hasBCBreaks() || !LegacyFormHelper::isLegacy()) {
```

GuilhemN: This detects symfony 3.0, when you use sf 2.8 we don't know whether your code works with the new system so we can't get rid of the deprecation without potentially introducing a bc break.
The best way to get rid of it is to upgrade to 3.0 :)

LordWeedlle: According to https://github.com/symfony/symfony/blob/2.8/UPGRADE-2.8.md#form (check section talking about ```method_exists(AbstractType::class, 'getBlockPrefix')```) we can use ```LegacyFormHelper::isLegacy()``` to detect if we should use fully-qualified class name or instance of forms.
So if I change my modification to keep previous if and only add my modification this shouldn't introduce bc break and keep the old code working =)

GuilhemN:
Thanks for contributing!

Issue #867, opened 2016-06-10
Confusion with naming strategies (underscore vs. camelCase)


Gregoire-M:
I have a model class called `Address`, with a `$zipCode` property (camelCase):

``` php
    /**
     * @Type("string")
     */
    private $zipCode;
```

For this object, my API expects a JSON property `zip_code` (underscore). I think that's the default behaviour of the JMS serializer.

The generated API doc parameters look like this:

| Parameter | Type | Required ? | Format | Description |
| --- | --- | --- | --- | --- |
| address[zip_code] | string | false |  |  |

When I add validation annotations to populate the "Format" column, here is the result:

``` php
    /**
     * @Assert\NotNull()
     * @Assert\Length(max="20")
     * @Type("string")
     */
    private $zipCode;
```

| Parameter | Type | Required ? | Format | Description |
| --- | --- | --- | --- | --- |
| address[zip_code] | string | false |  |  |
| address[zipCode] |  | true | {length: max: 20} |  |

As you can see, there's a duplication of the property, with 2 different naming strategies.

If I underscore the property in my model class, it works fine. But I don't want to overrule my naming conventions for that.

Is it a bug, or something I'm doing wrong?

Nanakii:
I have the same issue, can we have some feedback on this ?

Mairu:
You can use `@SerializedName("zipCode")` annotation from JMS to use only camelCase notation as workaround.

CaptainQuirk:
Having the same difficulty here. I don't want to use the camelCase version and using the underscore version in the ```@SerializedName``` annotation doesn't work : It get's dupplicated.

CaptainQuirk:
Sorry ! You have to configure the parsers ! However, if you want the jms serializer to be respected and validation assertion to be taken into account, you're in trouble !

sfarkas1988:
Is this a new issue or was it always this way? I've exactly the same problem and don't see a valid reason for it.

jmaheux:
I used a custom parser for this. See my comment on [this other issue](https://github.com/nelmio/NelmioApiDocBundle/issues/529#issuecomment-254338664).

develth:
I made a pull request where you can configure your validation parsers through config.yml

Here is an example of an SnakeCaseValidationParser

https://gist.github.com/develth/9b4e738538e27e1b6e117b4d2a4c7c03

Thanks to @jmaheux  for the hint.

JustBlackBird:
[...]

PR #985, opened 2017-04-13
Added possibility to use another ValidationParser through the config

develth:
The current ValidationParser adds property duplicates with camelCase & underscore naming. 

Added possibility to configure an own ValidationParser to change naming conventions.

refs #867


GuilhemN:
Thanks for contributing!

PR #988, opened 2017-04-28
Fix filters always as query parameters

lsmith77:


GuilhemN:
Thank you @lsmith77 !

PR #989, opened 2017-04-28
ensure that boolean fields are submited as boolean values when using json

lsmith77:


GuilhemN:
@willdurand @Seldaek is it possible to add @lsmith77 as a contributor please ? ☺

LGTM at first sight but i don't really know this part of the code.

lsmith77:
I am just not much of a javascript coder also it does like a very hacky direct approach to fix this (like I didn't even try to address XML).

Seldaek:
Added as collaborator. I haven't touched or used this bundle in years so I don't think my opinion matters much here :p

GuilhemN:
@Seldaek thanks :)

3.0 is in progress so imo it's ok to merge something a bit hacky if it works fine, we won't have maintainabily issues anyway.

lsmith77:
and I guess the test setup is also known as broken atm?

GuilhemN:
Yes it is since a patch release if I remember well, I need to take some time to fix them.

PR #995, opened 2017-05-16
Add ability to add a little description in Output definition

Osin:
Maybe is not the best way to do it, but i look for a way to support description or another anotation type without modify a lot the core.
Exemple:
I created an annotation class

```
class Description extends Constraint
{
    public $value;

    public function __construct(array $data)
    {
        $this->value = $data['value'];
    }
}
```

I create for this example a definition class
```

final class ProgramDefintion extends AbstractType

    /**
     * @Type("Int")
     * @Description("Identifiant en bdd front du programme - AutoIncrement sur 8 chiffres")
     */
    public $program_ai_id;

    /**
     * @Type("String")
     * @Description("Identifiant en bdd front du programme - Guid sur 36 caractères")
     */
    public $program_id_programme;

    /**
     * @Type("String")
     * @Description("Référence du programme")
     */
    public $program_ref_programme;

```

That's it, in "@ApiDoc",

```
     *     responseMap={
     *        "programs" = {"class"=AppBundle\Program\Definition\ProgramDefinition::class},
     *     }
```

I hope it helps you
capture du 2017-05-16 16-14-06


GuilhemN:
Feel free to reopen but know that 2.x is not the latest version anymore, so not the highest priority.

Osin:
Thank you, i will try first to update my NelmioApiDoc installation. After that i will send another pull-request about documentation on output

PR #996, opened 2017-05-23
Add support for ExpressionProperty

cmodijk:
If the property is a `ExpressionPropertyMetadata` don't try to extract a description as there is no property or method named they way of the item.

Without this the extractor will give a error `Property <property_name> does not exist`. 

I'm not sure if we should add a other way to extract it but in my case I get a item which is a string which will not allow me to extract a description.


GuilhemN:
Thanks for contributing!
I just set 3.0 as master so this is not valid anymore and I'm closing this.

PR #1007, opened 2017-06-07
Change the behavior of path_patterns filter to use OR instead of AND

bgarel:
The rule of path_patterns config parameter is not obvious. This pull request changes the filter behavior of the route path_patterns. It makes a OR between pattern elements instead of a AND.


GuilhemN:
Thank you @bgarel!

PR #1024, opened 2017-06-25
Add JMS serializer support

GuilhemN:
Adds support for jms metadata.

This PR needs tests but I don't really know how to make them as I don't really use the jms serializer, could one of you help me on this?

### DIFF
--- a/DependencyInjection/NelmioApiDocExtension.php
+++ b/DependencyInjection/NelmioApiDocExtension.php
@@ -13,6 +13,7 @@ namespace Nelmio\ApiDocBundle\DependencyInjection;
 
 use FOS\RestBundle\Controller\Annotations\ParamInterface;
 use Nelmio\ApiDocBundle\ModelDescriber\FormModelDescriber;
+use Nelmio\ApiDocBundle\ModelDescriber\JMSModelDescriber;
 use Nelmio\ApiDocBundle\Routing\FilteredRouteCollectionBuilder;
 use phpDocumentor\Reflection\DocBlockFactory;
 use Symfony\Component\Config\FileLocator;
@@ -33,6 +34,12 @@ final class NelmioApiDocExtension extends Extension implements PrependExtensionI
     public function prepend(ContainerBuilder $container)
     {
         $container->prependExtensionConfig('framework', ['property_info' => ['enabled' => true]]);
+
+        // JMS Serializer support
+        $bundles = $container->getParameter('kernel.bundles');
+        if (isset($bundles['JMSSerializerBundle'])) {
+            $container->prependExtensionConfig('nelmio_api_doc', ['models' => ['use_jms' => true]]);
+        }
     }
 
     /**
@@ -49,7 +56,7 @@ final class NelmioApiDocExtension extends Extension implements PrependExtensionI
             $container->register('nelmio_api_doc.model_describers.form', FormModelDescriber::class)
                 ->setPublic(false)
                 ->addArgument(new Reference('form.factory'))
-                ->addTag('nelmio_api_doc.model_describer', ['priority' => 10]);
+                ->addTag('nelmio_api_doc.model_describer', ['priority' => 100]);
         }
 
         // Filter routes
@@ -79,10 +86,18 @@ final class NelmioApiDocExtension extends Extension implements PrependExtensionI
             $loader->load('fos_rest.xml');
         }
 
-        $bundles = $container->getParameter('kernel.bundles');
         // ApiPlatform support
+        $bundles = $container->getParameter('kernel.bundles');
         if (isset($bundles['ApiPlatformBundle']) && class_exists('ApiPlatform\Core\Documentation\Documentation')) {
             $loader->load('api_platform.xml');
+        }
+
+        // JMS metadata support
+        if ($config['models']['use_jms']) {
+            $container->register('nelmio_api_doc.model_describers.jms', JMSModelDescriber::class)
+                ->setPublic(false)
+                ->setArguments([new Reference('jms_serializer.metadata_factory'), new Reference('jms_serializer.naming_strategy')])
+                ->addTag('nelmio_api_doc.model_describer', ['priority' => 50]);
         }
 
         // Import the base configuration
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -1,0 +1,128 @@
+<?php
+
+/*
+ * This file is part of the NelmioApiDocBundle package.
+ *
+ * (c) Nelmio
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Nelmio\ApiDocBundle\ModelDescriber;
+
+use EXSyst\Component\Swagger\Schema;
+use JMS\Serializer\Exclusion\GroupsExclusionStrategy;
+use JMS\Serializer\Metadata\PropertyMetadata;
+use JMS\Serializer\Naming\PropertyNamingStrategyInterface;
+use JMS\Serializer\SerializationContext;
+use Metadata\MetadataFactoryInterface;
+use Nelmio\ApiDocBundle\Describer\ModelRegistryAwareInterface;
+use Nelmio\ApiDocBundle\Describer\ModelRegistryAwareTrait;
+use Nelmio\ApiDocBundle\Model\Model;
+use Symfony\Component\PropertyInfo\Type;
+
+/**
+ * Uses the JMS metadata factory to extract input/output model information.
+ */
+class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareInterface
+{
+    use ModelRegistryAwareTrait;
+
+    private $factory;
+    private $namingStrategy;
+
+    public function __construct(MetadataFactoryInterface $factory, PropertyNamingStrategyInterface $namingStrategy)
+    {
+        $this->factory = $factory;
+        $this->namingStrategy = $namingStrategy;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function describe(Model $model, Schema $schema)
+    {
+        $className = $model->getType()->getClassName();
+        $metadata = $this->factory->getMetadataForClass($className);
+        if (null === $metadata) {
+            throw new \InvalidArgumentException(sprintf('No metadata found for class %s.', $className));
+        }
+
+        $groupsExclusion = null !== $model->getGroups() ? new GroupsExclusionStrategy($model->getGroups()) : null;
+
+        $schema->setType('object');
+        $properties = $schema->getProperties();
+        foreach ($metadata->propertyMetadata as $item) {
+            if (null === $item->type) {
+                continue;
+            }
+
+            // filter groups
+            if (null !== $groupsExclusion && $groupsExclusion->shouldSkipProperty($item, SerializationContext::create())) {
+                continue;
+            }
+
+            $name = $this->namingStrategy->translateName($item);
+            $property = $properties->get($name);
+
+            if ($type = $this->getNestedTypeInArray($item)) {
+                $property->setType('array');
+                $property = $property->getItems();
+            } else {
+                $type = $item->type['name'];
+            }
+
+            if (in_array($type, array('boolean', 'integer', 'string', ' float', 'array'))) {
+                $property->setType($type);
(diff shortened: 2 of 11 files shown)


Alex-D:
JMS Serializer seems to not be Symfony 4 ready :/

Alex-D:
Something like that should cover a lot of cases:

```php
<?php

namespace Nelmio\ApiDocBundle\Tests\Functional\Entity;

/**
 * User
 *
 * @ORM\Entity()
 * @Serializer\ExclusionPolicy("all")
 */
class User
{
    /**
     * @var int
     *
     * @ORM\Column(name="id", type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     * @Serializer\Expose
     */
    private $id;

    /**
     * @var string
     *
     * @ORM\Column(name="email", type="string", length=255, unique=true)
     * @Serializer\Expose
     */
    private $email;

    /**
     * @var string[]
     *
     * @ORM\Column(name="roles", type="json_array")
     * @Serializer\Type("array<string>")
     * @Serializer\Accessor(getter="getRoles",setter="setRoles")
     * @Serializer\Expose
     */
    private $roles;

    /**
     * @var string
     *
     * @ORM\Column(name="password", type="string", length=255)
     */
    private $password;

    /**
     * @var \DateTime
     *
     * @ORM\Column(name="created_at", type="datetime")
     * @Serializer\Expose
     */
    private $createdAt;

    /**
     * Set roles
     *
     * @param string[] $roles
     *
     * @return User
     */
    public function setRoles($roles)
    {
    }

    /**
     * Get roles
     *
     * @return string[]
     */
    public function getRoles()
    {
    }

    public function addUsers(User $user)
    {
    }

    public function setDummy(Dummy $dummy)
    {
    }
}

```

ModelDescriber/JMSModelDescriber.php
```diff
@@ -0,0 +1,141 @@
...
+
+        $params = array();
+        $reflection = new \ReflectionClass($className);
+        $defaultProperties = array_map(function ($default) {
+            if (is_array($default) && count($default) === 0) {
+                return null;
+            }
+
+            return $default;
+        }, $reflection->getDefaultProperties());
```
Alex-D: `$defaultProperties` seems to be unused :/

ModelDescriber/JMSModelDescriber.php
```diff
@@ -0,0 +1,141 @@
...
+
+            // filter groups
+            if (null !== $groupsExclusion && $groupsExclusion->shouldSkipProperty($item, SerializationContext::create())) {
+                continue;
+            }
+
+            $name = $this->namingStrategy->translateName($item);
+            $property = $properties->get($name);
+
+            $dataType = $this->processDataType($item);
```
Alex-D: `processDataType` methods does not exists, I think it should be something like this:

```php
<?php
    private function processDataType(PropertyMetadata $item)
    {
        return $item->type['name'];
    }
```

GuilhemN: I forgot to replace this call, thanks!

ModelDescriber/JMSModelDescriber.php
GuilhemN: Indeed, must be removed. This comes from nelmio 2.x but I decided to not reuse this logic for now as I don't think it is that useful.

GuilhemN:
Tests added with your sample :) However, I had to specify a type for `$id` and `$email`. Doesn't JMS support php doc blocks?

Alex-D:
Nope. But in my own class, it's based on `@ORM\Column` annotation :)

Tests/Functional/Entity/JMSDummy.php
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * User
+ *
+ * @Serializer\ExclusionPolicy("all")
+ */
+class JMSDummy
+{
+    /**
+     * @var int
```
Alex-D: I think we can remove all `@var` since there are not used by JMS

Tests/Functional/Entity/JMSDummy.php
```diff
@@ -0,0 +1,78 @@
...
+     */
+    private $createdAt;
+
+    /**
+     * Set roles
+     *
+     * @param string[] $roles
+     *
+     * @return User
+     */
```
Alex-D: Useless PHP doc can be removed

Tests/Functional/Entity/JMSDummy.php
```diff
@@ -0,0 +1,78 @@
...
+    public function setRoles($roles)
+    {
+    }
+
+    /**
+     * Get roles
+     *
+     * @return string[]
+     */
```
Alex-D: Same here

README.md
```diff
@@ -126,14 +126,41 @@ class UserController
...
+The metadata of the JMS serializer are used by default to describe your
+models. Note that PHP doc blocks aren't supported in this case.
+
+In case you prefer using the [Symfony PropertyInfo component](https://symfony.com/doc/current/components/property_info.html) (you
+won't be able to use JMS serialization groups), you can disable JMS serializer
+support in your config:
+
+```yml
+nelmio_api_doc:
+    models: { use_jms: false }
```
Alex-D: `false` is default value, we should turn that to `true` when we use JMS, no?

.travis.yml
```diff
@@ -29,6 +29,7 @@ before_install:
     - phpenv config-rm xdebug.ini
     - if [ "$SF_4" == "true" ]; then composer config minimum-stability dev; fi
     - if [ "$SF_4" == "true" ]; then composer require symfony/symfony:4.0.* --dev --no-update; fi
+    - if [ "$SF_4" == "true" ]; then composer remove --dev --no-update jms/serializer-bundle; fi
```
Alex-D: Seems not working :/

Alex-D:
We should add something like that to cover the part of code which load another model as items of the array:

```php
<?php
    /**
     * @var User[]
     *
     * @Serializer\Type("array<User>")
     * @Serializer\Expose
     */
    private $friends;
```

README.md
GuilhemN: It's done automatically by the extension (see https://github.com/nelmio/NelmioApiDocBundle/pull/1024/files#diff-11c1d6d50f3adeb4ab15baf6d54e707dR41).

.travis.yml
GuilhemN: yeah... I'm investigating

GuilhemN:
> We should add something like that to cover the part of code which load another model as items of the array

done

README.md
Alex-D: Oooh OK :)

Tests/Functional/Entity/JMSDummy.php
```diff
@@ -0,0 +1,63 @@
...
+     * @Serializer\Type("array<string>")
+     * @Serializer\Accessor(getter="getRoles", setter="setRoles")
+     * @Serializer\Expose
+     */
+    private $roles;
+
+    /**
+     * @Serializer\Type("string")
+     */
+    private $password;
```
Alex-D: We should not use Serializer here. It's just to have an example of an hidden property.

Tests/Functional/Entity/JMSDummy.php
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Ignored as the JMS serializer can't detect its type.
+     *
+     * @Serializer\Expose
+     */
```
Alex-D: And add the

`@Serializer\Type("DateTime")`

To test Object case.

GuilhemN: In fact I wanted to make sure that the property is not documented when the type is not guessable.

Tests/Functional/Entity/JMSDummy.php
GuilhemN: Just here to make sure this field is not documented :)

Tests/Functional/Entity/JMSDummy.php
Alex-D: So, we should add a property to test Object case?

GuilhemN: What do you mean ? `$users` already tests object support.

Tests/Functional/Entity/JMSDummy.php
Alex-D: Hum. OK :)

Tests/Functional/Entity/JMSDummy.php
Alex-D: Nope, `$users` cover array of Object, not Object :)

DependencyInjection/NelmioApiDocExtension.php
```diff
@@ -33,6 +34,12 @@
     public function prepend(ContainerBuilder $container)
     {
         $container->prependExtensionConfig('framework', ['property_info' => ['enabled' => true]]);
+
+        // ApiPlatform support
```
Alex-D: This comment is an mistake? It should be `JMS Serializer support`, no?

pdugas:
Should there be a test for the alternative approach using the default ExclusionPolicy("none") and Exclude()?  I personally use this approach since I typically am only looking to exclude a few problematic properties.

goetas:
@Alex-D 
> JMS Serializer seems to not be Symfony 4 ready :/

Any special feature from sf4 that is required here? (willing to help)

Alex-D:
Since JMSSerializer has added 4.0 to his versions: https://github.com/schmittjoh/JMSSerializerBundle/blob/master/composer.json#L22

@GuilhemN You can re-add it to test suite :)

I try to use this branch of apidocbundle on my project, but it seems not using JMS annotations to generate swagger models for my models in Models folder, but it works for entities in Entity folder :/

Am I missing something?

DependencyInjection/NelmioApiDocExtension.php
GuilhemN: indeed

ModelDescriber/JMSModelDescriber.php
```diff
@@ -0,0 +1,128 @@
...
+            // filter groups
+            if (null !== $groupsExclusion && $groupsExclusion->shouldSkipProperty($item, SerializationContext::create())) {
+                continue;
+            }
+
+            $name = $this->namingStrategy->translateName($item);
+            $property = $properties->get($name);
+
+            if ($type = $this->getNestedTypeInArray($item)) {
+                $property->setType('array');
```
GuilhemN: If we use `object`, we have to specify which keys will be used. That seems complicated in this case...

ModelDescriber/JMSModelDescriber.php
```diff
@@ -0,0 +1,128 @@
...
+                $type = $item->type['name'];
+            }
+
+            if (in_array($type, array('boolean', 'integer', 'string', ' float', 'array'))) {
+                $property->setType($type);
+            } elseif ('double' === $type) {
+                $property->setType('float');
+            } elseif ('DateTime' === $type) {
+                $property->setType('string');
+                $property->setFormat('date');
```
GuilhemN: AFAIK we can't change that, dates must follow ISO 8601 (see https://xml2rfc.tools.ietf.org/public/rfc/html/rfc3339.html#anchor14).

ModelDescriber/JMSModelDescriber.php
```diff
@@ -0,0 +1,128 @@
...
+                $property = $property->getItems();
+            } else {
+                $type = $item->type['name'];
+            }
+
+            if (in_array($type, array('boolean', 'integer', 'string', ' float', 'array'))) {
+                $property->setType($type);
+            } elseif ('double' === $type) {
+                $property->setType('float');
+            } elseif ('DateTime' === $type) {
```
GuilhemN: sure

GuilhemN:
> Should there be a test for the alternative approach using the default ExclusionPolicy("none") and Exclude()? I personally use this approach since I typically am only looking to exclude a few problematic properties.

I don't think so, it's up to jms serializer internals.

> Any special feature from sf4 that is required here? (willing to help)

No, it's just to run the tests.

> One question about groups and exclusion strategies... in the jms serializer, recently, exclusion strategies become dynamic... this means that a field might be exposed in certain situation and in other not. Is there a way to mark a filed as "maybe" exposed or "optionally" available?

I guess not marking them as `required` is enough.

GuilhemN:
Thank you for all your reviews!

PR #1031, opened 2017-07-03
Bugfix/no overrides if already set

ToGo101:
- required for PathVariables is no longer overwritten, if already set
- type is no longer overwritten, if already set

Small Example: 

```
/**
 * @Method("GET")
 * @Route("/api/{optional}", name="optionalRoute", defaults={"optional" = 1})
 * @SWG\Tag(name="Optional")
 * @SWG\Response(
 *     response=200,
 *     description="TBD",
 *     @SWG\Schema(
 *          title="Test",
 *          example="{httpCode: 200, success:true}"
 *
 *     )
 * )
 * @SWG\Parameter(in="path", name="optional", type="integer", description="The Optional Value", required=false, default=1)
 * @return JsonResponse
 */
public function optionalParam(int $optional = 1)
{
	return new JsonResponse([
		'httpCode' => 200,
		'success'  => true,
	]);
}
```
Before:
grafik

After:
grafik

### DIFF
--- a/RouteDescriber/RouteMetadataDescriber.php
+++ b/RouteDescriber/RouteMetadataDescriber.php
@@ -33,9 +33,8 @@ final class RouteMetadataDescriber implements RouteDescriberInterface
                 }
 
                 $parameter = $operation->getParameters()->get($pathVariable, 'path');
-                if (null === $parameter->getRequired()) {
-                    $parameter->setRequired(true);
-                }
+                $parameter->setRequired(true);
+
                 if (null === $parameter->getType()) {
                     $parameter->setType('string');
                 }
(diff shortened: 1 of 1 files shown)


RouteDescriber/RouteMetadataDescriber.php
```diff
@@ -33,8 +33,12 @@ public function describe(Swagger $api, Route $route, \ReflectionMethod $reflecti
                 }
 
                 $parameter = $operation->getParameters()->get($pathVariable, 'path');
-                $parameter->setRequired(true);
-                $parameter->setType('string');
+                if (null === $parameter->getRequired()) {
+                    $parameter->setRequired(true);
```
GuilhemN: Swagger doesn't support optional path parameters (see https://swagger.io/specification/#parameterObject), so this can be reverted.
I'm not sure what we could do here.

ToGo101: Hmm. Yeah, thats true.
I've spoken to a few people, its better to make two routes, so its more clear.
As soon as i'm home, i'll change this.

GuilhemN:
Thank you @ToGo101 :)

Issue #1032, opened 2017-07-04
[3.0] app_dev.php is ignored


coolmic:
Hi,

I just migrated from 2.x to 3.0 , and I have an issue when I used app_dev.php in the url.

My api doc url is 
`http://domain.dev/app_dev.php/api/doc/`

When I do a request, it target

`http://domain.dev/auth/login`

instead of

`http://domain.dev/api_dev.php/auth/login` .

Did I miss something? It worked well in 2.x .

GuilhemN:
see https://github.com/nelmio/NelmioApiDocBundle/pull/1035

PR #1035, opened 2017-07-05
Use request base url

GuilhemN:
Fix https://github.com/nelmio/NelmioApiDocBundle/issues/1032


tarlepp:
Hmm, when is this merged ?

GuilhemN:
There's an issue with Api-Platform, I need to find a solution first but i don't have the time right now.

PR #1040, opened 2017-07-22
Added Authorization header to Curl block

aurimasniekis:
Added Authorization header to Curl block, fixed Curl url to have origin, fixed `escapeShell` method to not add escape char for space element as it fucks header values.

screen shot 2017-07-22 at 6 48 44 pm

### DIFF
--- a/Resources/views/layout.html.twig
+++ b/Resources/views/layout.html.twig
@@ -85,6 +85,22 @@
                 }
             };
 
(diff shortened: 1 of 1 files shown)


aurimasniekis:
I don't see a correlation between my changes and that Travis failed...

GuilhemN:
2.0 is no longer supported, for new features, please target 3.0.

PR #1049, opened 2017-07-31
Add missing svg definitions for icons

HaraldVelner:
The SVG definitions for icons are missing in the nelmio index twig


GuilhemN:
Imo it can be put in `index.html.twig` :)

GuilhemN:
Thank you @HaraldVelner!

Issue #1051, opened 2017-08-02
3.0.0-beta FR: Could we add useAttributeAsKey() to documentation in Configuration?


discordier:
I have a rather large config specified in yml that I wanted to split up and include into the main config as usual.
However, the array node `documentation` lacks the call to `->useAttributeAsKey()` and therefore the configuration breaks (all configs loose their keys except for the first one.

I added the following on my side and it works pretty nice (I am unsure about the best name for the key attribute though).
```diff
    public function getConfigTreeBuilder()
    {
        $treeBuilder = new TreeBuilder();
        $treeBuilder
            ->root('nelmio_api_doc')
            ->children()
                ->arrayNode('documentation')
+               ->useAttributeAsKey('name')
                    ->info('The documentation used as base')
                    ->example(['info' => ['title' => 'My App']])
                    ->prototype('variable')->end()
                ->end()
                ->arrayNode('routes')
                    ->info('Filter the routes that are documented')
                    ->addDefaultsIfNotSet()
                    ->children()
                        ->arrayNode('path_patterns')
                            ->example(['^/api', '^/api(?!/admin)'])
                            ->prototype('scalar')->end()
                        ->end()
                    ->end()
                ->end()
                ->arrayNode('models')
                    ->addDefaultsIfNotSet()
                    ->children()
                        ->booleanNode('use_jms')->defaultFalse()->end()
                    ->end()
                ->end()
            ->end();

        return $treeBuilder;
    }
```
In the end it could be anything not in use by swagger, therefore "name" might not be the best pick.

GuilhemN:
:+1:, what about `key`? Could you submit a PR with some tests please?

discordier:
Will do.

discordier:
See #1061

PR #1061, opened 2017-08-08
Allow configuration to be split on root key level.

discordier:
We now allow to split the configuration on multiple files, each
containing one of the various root keys of the "documentation" entry.

This means, you may now split "documentation/*" out to own yml files and
import them.

NOTE: this only allows splitting of entries being a direct child of the
documentation entry, so splitting i.e. the info key over multiple files
will not work.

This is the implementation of #1051.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -23,6 +23,7 @@ final class Configuration implements ConfigurationInterface
             ->root('nelmio_api_doc')
             ->children()
                 ->arrayNode('documentation')
+                    ->useAttributeAsKey('key')
                     ->info('The documentation used as base')
                     ->example(['info' => ['title' => 'My App']])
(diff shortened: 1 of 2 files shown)


GuilhemN:
Thank you @discordier!

discordier:
@GuilhemN How shall we proceed about documenting the stuff in this repo in future?

GuilhemN:
@discordier it'd be great to recreate docs for https://symfony.com/doc/current/bundles/NelmioApiDocBundle/index.html. 
For now as it's light I did put everything in the README but it would probably be better if we dedicated pages to each important feature, especially explaining`@SWG` annotations in our repo instead of linking to `zircote-php` docs.

I won't have much time to dedicate to docs during the next two years, I'll do my best but I can promise anything. If you could submit PRs that'd be awesome :)

PR #1062, opened 2017-08-09
Parsing default values from routes

qkdreyer:


Extractor/Handler/PhpDocHandler.php
```diff
@@ -107,6 +107,14 @@ public function handle(ApiDoc $annotation, array $annotations, Route $route, \Re
             }
         }
 
+        foreach ($route->getDefaults() as $var => $default) {
+            if (strpos($var, '_') !== 0) {
```
BigZ: yoda conditions are adivsed

Extractor/Handler/PhpDocHandler.php
```diff
@@ -107,6 +107,14 @@ public function handle(ApiDoc $annotation, array $annotations, Route $route, \Re
...
+                if (!isset($annotationRequirements[$var]['default'])) {
```
BigZ: you could use empty instead of !isset
that wouls be more easy to read

GuilhemN:
2.x is no longer maintained, please consider upgrading to 3.x.

BigZ:
thanks for updating. i checked out 3.x, good job :)

PR #1067, opened 2017-08-21
Fix fatal error when using a Constraint as requirement

sandermarechal:


GuilhemN:
Thank you @sandermarechal!

GuilhemN:
Thank you @sandermarechal!

Issue #1065, opened 2017-08-16
$ref not found 


MassiveHiggsField:
Action looks like this:

    /**
     * @Operation(
     *     summary="Gets an agent for a given id.",
     *     @SWG\Parameter(
     *         name="id",
     *         in="query",
     *         description="Which agent to return.",
     *         required=false,
     *         type="string"
     *     ),
     *     @SWG\Response(
     *         response="200",
     *         description="Returned when successful.",
    *          @Model(type=AppBundle\Entity\Agent::class, groups={"agent.show"})
     *     ),
     *     @SWG\Response(
     *         response="404",
     *         description="Returned when the agent is not found."
     *     )
     * )
     *
     * @Rest\Get("/agent")
     * @Rest\QueryParam(name="id", requirements="\d+", description="Which agent to return.")
     */

Results in: 

    "User Notice: $ref "#/definitions/Agent" not found for @SWG\Schema() in \Nelmio\ApiDocBundle\SwaggerPhp\ModelRegister->__invoke() in /vagrant/api/vendor/nelmio/api-doc-bundle/SwaggerPhp/ModelRegister.php on line 74"

Any help appreciated

7thcubic:
For some reason i am encountering the same issue.
I tried to install a fresh copy of symfony + jms serializer + nelmio api only to test and i also hit the same issue

mathielen:
Same here

GuilhemN:
Could you provide us the stack trace please? I think because we don't register the models in zircote-php, it raises a notice when it validates its output.

7thcubic:
```php
namespace AppBundle\Model;
use JMS\Serializer\Annotation as Serializer;

class Testing
{

    /**
     * @var
     * @Serializer\Type("string")
     */
    public $test;

}
```

On default controller
```php
    /**
     * @SWG\Response(
     *     response="200",
     *     description="User successfully retrieved",
     *     @Model(type="AppBundle\Model\Testing")
     * )
```

```
Symfony\Component\Debug\Exception\ContextErrorException:
User Notice: $ref "#/definitions/Testing" not found for @SWG\Schema() in \Nelmio\ApiDocBundle\SwaggerPhp\ModelRegister->__invoke() in D:\Work\tnelm\vendor\nelmio\api-doc-bundle\SwaggerPhp\ModelRegister.php on line 71

  at vendor\zircote\swagger-php\src\Logger.php:38
[...]

PR #1069, opened 2017-08-22
Update SwaggerPhpDescriber.php

betzholz:
$ref not found #1065 Fixed


GuilhemN:
That's like ignoring all errors from PHP :) That's a no go for me...

PR #1076, opened 2017-08-28
Syntax update

Nyholm:
`~7.0|~7.1` is the same as `~7.0`. I think you mean to do `7.0.*|7.1.*` since PHP does not do Semver.


GuilhemN:
Indeed, thank you @Nyholm!

PR #1077, opened 2017-08-28
Fix type for float / double for JMS models

wesleylancel:


ModelDescriber/JMSModelDescriber.php
```diff
@@ -73,10 +73,10 @@ public function describe(Model $model, Schema $schema)
                 $type = $item->type['name'];
             }
 
-            if (in_array($type, array('boolean', 'integer', 'string', ' float', 'array'))) {
+            if (in_array($type, array('boolean', 'integer', 'string', 'array'))) {
                 $property->setType($type);
-            } elseif ('double' === $type) {
-                $property->setType('float');
+            } elseif ('double' === $type || 'float' === $type) {
+                $property->setType('number');
```
GuilhemN: Imo we should also add `$property->setFormat($type);`.

wesleylancel: Added!

GuilhemN:
Thank you @wesleylancel!

PR #1081, opened 2017-09-11
fix: map route requirements to parameter pattern instead of format

Jasperras:
Currently symfony route requirements are mapped to swaggers' 'format' field on parameters. As described in the following links, the format is meant to be an extension upon the specified 'type' field. The requirements should be mapped to the 'pattern' field.

see:
[OpenAPI Parameter Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject)
[JSON Validation RFC](https://tools.ietf.org/html/draft-fge-json-schema-validation-00#section-5.2.3)
[OpenAPI Data Type Format](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#dataTypeFormat)

### DIFF
--- a/RouteDescriber/RouteMetadataDescriber.php
+++ b/RouteDescriber/RouteMetadataDescriber.php
@@ -40,7 +40,7 @@ final class RouteMetadataDescriber implements RouteDescriberInterface
                 }
 
                 if (isset($requirements[$pathVariable])) {
-                    $parameter->setFormat($requirements[$pathVariable]);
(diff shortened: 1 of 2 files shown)


haroldiedema:
Can this be merged anytime soon?

We're currently using a forked version of this in our applications, but we'd ideally want to use the composer package from this repository.

GuilhemN:
Thank you @Jasperras!

PR #1086, opened 2017-09-15
Added Doctrine support, get some troubble with ObjectModelDescriber.p…

betzholz:
…hp so deactived him.
also i added the possibility to update a schema by a second model disscriber.

### DIFF
--- a/DependencyInjection/NelmioApiDocExtension.php
+++ b/DependencyInjection/NelmioApiDocExtension.php
@@ -12,6 +12,7 @@
 namespace Nelmio\ApiDocBundle\DependencyInjection;
 
 use FOS\RestBundle\Controller\Annotations\ParamInterface;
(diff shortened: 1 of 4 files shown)


GuilhemN:
If the issue is that you have models named `int`, etc. this could be solved easily by inlining built-in types in `ObjectModelDescriber`.

BigZ:
Let us know more about the problems your facing @betzholz

betzholz:
Hey sorry for the late response.
I got for all primitives somthing like:  "string: { type: "string"}, int: {type: "integer"}", so the definition is valid but not nice...
If you dont use the ObjectModelDescriber everything in my cases is fine... but i bet this is not an option ;)
I think this s a result of overwriting... in my version the modelregestry allowes the decriber to overwrite the other descripter or rather complement the schema, so that many sources can bee used.

GuilhemN:
Should be fixed by https://github.com/nelmio/NelmioApiDocBundle/pull/1125 :)

PR #1093, opened 2017-09-26
Enclose form type fields in an object

BigZ:
When serializing a form type, its fields are actually enclosed in an object which name is the one of the entity the form takes account for.

To achieve this, we register one more form wich is actually an extension of the original formType in the model registery to have a second pass on the describer.
On the first pass, we had a ref to the fields definition on the original form type, and on the second one we actually register the fields.

Something like
```
@SWG\Parameter(
     *     name="body",
     *     in="body",
     *     description="Artist to add",
     *     required=true,
     *     @Model(type=ArtistType::class),
     * )
```

With a form like this one

```
        $builder
            ->add('name', TextType::class, ['required' => true])
            ->add('slug', TextType::class, ['required' => true])
```

Should generate those definitions

```
        "ArtistType": {
            "required": [
                "artist"
            ],
            "properties": {
                "artist": {
                    "$ref": "#/definitions/ArtistTypeProperties"
                }
            },
            "type": "object"
        },
        "ArtistTypeProperties": {
            "required": [
                "name",
                "slug"
            ],
            "properties": {
                "name": {
                    "type": "string"
                },
                "slug": {
                    "type": "string"
                }
            },
            "type": "object"
        }
```

### DIFF
--- a/ModelDescriber/FormModelDescriber.php
+++ b/ModelDescriber/FormModelDescriber.php
@@ -45,12 +45,10 @@ final class FormModelDescriber implements ModelDescriberInterface, ModelRegistry
         }
 
         $schema->setType('object');
-        $properties = $schema->getProperties();
-
         $class = $model->getType()->getClassName();
 
         $form = $this->formFactory->create($class, null, []);
-        $this->parseForm($schema, $form);
+        $this->parseForm($schema, $form, $model);
     }
 
     public function supports(Model $model): bool
@@ -58,7 +56,16 @@ final class FormModelDescriber implements ModelDescriberInterface, ModelRegistry
         return is_a($model->getType()->getClassName(), FormTypeInterface::class, true);
     }
 
-    private function parseForm(Schema $schema, FormInterface $form)
+    private function parseForm(Schema $schema, FormInterface $form, Model $model)
+    {
+        if ('Properties' === substr($model->getType()->getClassName(), -strlen('Properties'))) {
+            return $this->addProperties($schema, $form);
+        }
+
+        $this->addEnclosingDefinition($schema, $model);
+    }
+
+    private function addProperties(Schema $schema, FormInterface $form)
     {
         $properties = $schema->getProperties();
 
@@ -131,6 +138,26 @@ final class FormModelDescriber implements ModelDescriberInterface, ModelRegistry
                 $schema->setRequired($required);
             }
         }
+    }
+
+    private function addEnclosingDefinition(Schema $schema, Model $model)
+    {
+        $formClass = $model->getType()->getClassName();
+        $properties = $schema->getProperties();
+
+        $propertiesFormClass = substr($formClass, strrpos($formClass, '\\') + 1) . 'Properties';
+        $entityName = strtolower(str_replace('TypeProperties', '', $propertiesFormClass));
+
+        $property = $properties->get($entityName);
+        $schema->setRequired([$entityName]);
+        $property->setRef('#/definitions/'.$propertiesFormClass);
+
+        // register the properties for next pass
+        if (!class_exists($propertiesFormClass)) {
+            eval("class $propertiesFormClass extends $formClass {}");
+        }
+        $model = new Model(new Type(Type::BUILTIN_TYPE_OBJECT, false, $propertiesFormClass));
+        $this->modelRegistry->register($model);
     }
 
     private function isBuiltinType(string $type): bool
--- a/Tests/Functional/FunctionalTest.php
+++ b/Tests/Functional/FunctionalTest.php
@@ -177,26 +177,37 @@ class FunctionalTest extends WebTestCase
 
     public function testFormSupport()
     {
+        // User Object
         $this->assertEquals([
             'type' => 'object',
             'properties' => [
-                'dummy' => ['$ref' => '#/definitions/DummyType'],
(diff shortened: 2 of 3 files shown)


BigZ:
i'd be willing to know if that is different in other setups than mine,
but this is the actual behavior of symfony 2/3 + jms serializer + fosrestbundle + form
which is quite common i guess

kopaygorodsky:
@BigZ  Hey. 
Could you pls explain what is the advantage of enclosing fields in an object? 
In current version we define form type as a model and do parsing of fields, in the result you will see an object DummyType with properties.  
Your proposal to create DummyProperties object with same properties makes no sense for me or I've got your idea wrong.

BigZ:
Hi @kopaygorodsky and thanks a lot for your time !
the advantage is that is fits the actual implementation of symfony + fosrestbundle + form, afaik.
let's look at those 2 examples, which are really simple POST actions of an entity:
https://github.com/BigZ/ftvenapi/blob/master/src/ApiBundle/Controller/ArticleController.php#L49
https://github.com/BigZ/promoteapi/blob/master/src/AppBundle/Controller/ArtistController.php#L94
it will expect something like
```
{ "entity" -> {
    "field1" => "value1",
    "field2" => "value2"
}}

```
to work.
submitting 
```
{
    "field1" => "value1",
    "field2" => "value2"
}
```
would fail and return something like
```
{
    "children": {
        "field1": {},
        "field2": {}
    }
}
```
Both are very simple & minimal implementations. 

I'm willing to know any behavior that would be different than what i get. I'm pretty sure that i will have to fix this PR to fit to different setups.
And maybe i just got it wrong :)
Cheers

kopaygorodsky:
@BigZ 
You use 2.* version of ApiDoc in the second example, right? 

Do you talk about 3.* version? 

[3.*]
You can name your property in annotation

```
     *  @SWG\Parameter(
     *      name="vacancy", -> can be any name you want
     *      in="body",
     *      description="Do something",
     *      required=true,
     *      @Model(type=VacancyType::class)
     *  )
```
image
What is actually 
```
{ "vacancy" -> {
    "field1" => "value1",
    "field2" => "value2"
}}

```

Am I right?

BigZ:
With what you said, when you press the button "try it out" it will generate this CURL request
`curl -X POST "http://localhost/vacancies" -H "accept: application/json" -H "Content-Type: application/json" -d "{ \"field1\": \"string\", \"field2\": \"string\"}"`
which would fail as the correct request would be
`curl -X POST "http://localhost/vacancies" -H "accept: application/json" -H "Content-Type: application/json" -d "{ \"vacancy\": { \"field1\": \"string\", \"field2\": \"string\" }}"`
My PR fixes that :)

It's the way the swagger spec works. i tried it with swagger-ui, dredd and nelmioapidoc - it's how we are supposed to do it, from what i did read around.
This comment actually pointed me towards this: https://github.com/apiaryio/dredd/issues/637#issuecomment-269322420

kopaygorodsky:
Yes, I see that it doesn't work correct. 
On my prntscreen(comment above)  I define a parameter `vacancy` inside the  `body`.
It's clear that you have a parameter `vacancy` and  serialized model as a value, it should work, but it's not.  
You PR fixes this, but not in a right way, because then you will have parameter `someName` + enclosed fields in `someName`. It's like a duplicate (check prntscreen)
image

Let's ask @GuilhemN to be sure.

kopaygorodsky:
And if you try to define 2 parameters with ` @SWG\Parameter` in `body`  - only one will be submitted, what is not correct. (just checked and confirmed)

BigZ:
it's not a bug, it's the way swagger works.
Look at the spec here https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-4
`A list of parameters that are applicable for all the operations described under this path. These parameters can be overridden at the operation level, but cannot be removed there. The list MUST NOT include duplicated parameters. A unique parameter is defined by a combination of a name and location. The list can use the Reference Object to link to parameters that are defined at the Swagger Object's parameters. **There can be one "body" parameter at most**.`

BigZ:
@GuilhemN can we have your feedback on this one please ?

kopaygorodsky:
@BigZ I found the way how to enclose it
```
     *  @SWG\Parameter(
     *      name="request", //any name
     *      in="body",
     *      description="updated entity information",
     *      required=true,
     *     @SWG\Schema(
     *          required={"someName"},
     *          @SWG\Property(property="vacancy", ref="#/definitions/SomeNameType"),
     *          @Model(type= SomeNameType::class)
     *      )
     *  ),
```
and the result is:
```
{
  "someName": {
  ...fields of SomeNameType
    },
}
```
@Model is used to define SomeNameType definition.

kopaygorodsky:
@BigZ hey, am I correct?

GuilhemN:
Sorry I missed your PR...  I'm not a big fan of this solution, if I remember well, I coded it this way because I think it's more usual to remove forms name in an API. Maybe we could add a config option or a new option in `@Model` to change this behavior instead ?

kopaygorodsky:
@GuilhemN yes, I fully agree. Will try to fix it.

GuilhemN:
> btw this pr can be closed, I think. 

I'll wait for @BigZ feedback before :)

GuilhemN:
Closing for now, feel free to reopen if you want to continue it.

PR #1094, opened 2017-09-26
Support for PHP 7.2

tarlepp:


composer.json
```diff
@@ -15,7 +15,7 @@
         }
     ],
     "require": {
-        "php": "7.0.*|7.1.*",
+        "php": "7.0.*|7.1.*|7.2.*",
```
BigZ: why not 
`"php": "^7"`
?

tarlepp: just following existing pattern

BigZ: i'm questioning the existing pattern here ^^

GuilhemN: I guess we can change it to ^7.0 :)

GuilhemN:
Thank you @tarlepp!

tamcy:
Um.. In PHP 7.2 this line will throw a user notice error **Warning: count(): Parameter must be an array or an object that implements Countable**, because `\EXSyst\Component\Swagger\Collections\Responses` is not an array and doesn't implements `Countable`. 

https://github.com/nelmio/NelmioApiDocBundle/blob/87f40feefcb3ac95d64d9620011b5b3f4292a957/Describer/DefaultDescriber.php#L41

Given it implements `Traversable`, changing `count()` to `iterator_count()` shall fix the problem.

PR #1098, opened 2017-10-03
Do not test on symfony 4

BigZ:
I'm not sure it's very wise to test on dev dependencies... Especially when it breaks the build.
Let's wait a bit to bring it back in :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,14 +22,9 @@ matrix:
   include:
     - php: 7.0
       env: COMPOSER_FLAGS="--prefer-lowest"
-    - php: 7.1
-      env: SF_4="true"
 
 before_install:
     - phpenv config-rm xdebug.ini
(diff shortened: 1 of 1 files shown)


hboomsma:
Is it not better to add `allow_failure`? I guess it is a good thing to support Symfony 4 in the future?

GuilhemN:
It's not complicated to fix, I just don't have the time right now to fix it :)
I'd rather fix the issue than ignore it as we'll very likely just forget it.

BigZ:
hardcoded dev versions sounds unstable anyway. what is the benefit of unreliable test results ?
fixing it + allow failure can somehow be ok while waiting for the stable release, but we'll have to change something on the build at release date anyway

GuilhemN:
Tests are fixed now :) Thanks!

PR #1104, opened 2017-10-12
PHP 7.2 compatibility fix

tamcy:
Starting from PHP 7.2 a warning will be thrown [when calling count() against non-countable objects](https://wiki.php.net/rfc/counting_non_countables).

This pull request attempts to fix a case where count() is called on a non-countable object - 
 `\EXSyst\Component\Swagger\Collections\Responses` is not an array and doesn't implements `Countable`. Given it implements `Traversable`, the counting code is changed from `count()` to `iterator_count()`.

While the title states it as a compatibility fix, I suspect this also fixes a hidden bug.


GuilhemN:
Thank you @tamcy!

PR #1105, opened 2017-10-13
Supporting different sets of documentations for different domains

gam6itko:
The feature allows to configure displayed API function for different domains.
Backward compatibility was saved.
Tests work ok.

```
nelmio_api_doc:
    routes:
        api:
            host: api.example.com
            path_patterns: # an array of regexps
                - ^/api
        rest:
            host: rest.example.com
            path_patterns:
                - ^/rest
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 /.phpunit
 /Tests/Functional/cache
 /Tests/Functional/logs
+/.idea/
--- a/composer.json
+++ b/composer.json
@@ -36,11 +36,11 @@
         "symfony/phpunit-bridge": "^3.3",
         "sensio/framework-extra-bundle": "^3.0",
         "doctrine/annotations": "^1.2",
-
         "phpdocumentor/reflection-docblock": "^3.1",
         "api-platform/core": "^2.0.3",
         "friendsofsymfony/rest-bundle": "^2.0",
-        "jms/serializer-bundle": "^1.0|^2.0"
+        "jms/serializer-bundle": "^1.0|^2.0",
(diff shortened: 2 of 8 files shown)


GuilhemN:
I wonder if that's worth being supported, it impacts a lot our code and i'm not sure it is that useful. Wdyt @dbu?

.gitignore
```diff
@@ -7,3 +7,4 @@
 /.phpunit
 /Tests/Functional/cache
 /Tests/Functional/logs
+/.idea/
```
dbu: this is dev setup specific and should be in each devs global gitignore, not in projects

composer.json
```diff
@@ -36,11 +36,11 @@
         "symfony/phpunit-bridge": "^3.3",
         "sensio/framework-extra-bundle": "^3.0",
         "doctrine/annotations": "^1.2",
-
         "phpdocumentor/reflection-docblock": "^3.1",
         "api-platform/core": "^2.0.3",
         "friendsofsymfony/rest-bundle": "^2.0",
-        "jms/serializer-bundle": "^1.0|^2.0"
+        "jms/serializer-bundle": "^1.0|^2.0",
+        "phpunit/phpunit": "^6.3"
```
dbu: i don't think that this should be added here, it has nothing to do with the feature.

GuilhemN:
@dbu I think what you actually want is areas support (see https://github.com/nelmio/NelmioApiDocBundle/pull/1169). What this PR proposes is more specific, it only allows to filter routes differently based on their host, I think this should be done in 3rd party code by replacing the routes filter.

PR #1119, opened 2017-10-26
Support the swagger-php Property annotation for Models

juchi:
Hi,

When using the Model annotation in order to reference an Entity class in a doc block, the properties from the object are parsed using the PropertyInfo component to get the variable type an create a $ref for this property.
This behavior prevents to add more information to the property. For instance, it's not possible to set an example value.

This PR adds the ability to annotate object properties with `Swagger\Annotations\Property` and fallback to the current behavior (using $ref) if none is set.

Thanks for your comments and advices.

### DIFF
--- a/ModelDescriber/ObjectModelDescriber.php
+++ b/ModelDescriber/ObjectModelDescriber.php
@@ -11,10 +11,12 @@
 
 namespace Nelmio\ApiDocBundle\ModelDescriber;
 
+use Doctrine\Common\Annotations\Reader;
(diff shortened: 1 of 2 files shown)


Vowow:
Good idea. +1

GuilhemN:
Sorry to answer so lately... Could you test your changes please?

GuilhemN:
Closing in favor of https://github.com/nelmio/NelmioApiDocBundle/pull/1125.

PR #1129, opened 2017-11-24
Mention/example of output collection annotation

woofiewilly:
I spent way too much effort trying to find this feature. I only found it by seeing it in the source code and experimenting a couple times to see how it worked. I just added a simple mention and example.


GuilhemN:
Thank you @woofiewilly!

Issue #1112, opened 2017-10-19
[3.0] jms child item field serializing


7thcubic:
I think the describer is not parsing the child group for jms serializer properly.

Example:

```php
class Foo {
//Groups("content")
//Reference Bar
public $bar;
//Groups("list')
public $id;
}

class Bar {
//Groups("list")
public $id;
//Groups("full")
public $text;
//Groups("say")
public $say;
}
```

with the serialization group of
["list","content", "bar" => "full"]

the serialization registry should handle

```php
$property->setRef(
                    $this->modelRegistry->register(new Model(new Type(Type::BUILTIN_TYPE_OBJECT, false, $type), ["full"]))
                );
```

instead of

```php
$property->setRef(
                    $this->modelRegistry->register(new Model(new Type(Type::BUILTIN_TYPE_OBJECT, false, $type), ["list","content", "bar" => "full"]))
                );
```

Some how or rather it isn't processing correctly right now because we are re-registering the child object with the entire group string which does not generate the correct expected format

https://jmsyst.com/libs/serializer/master/cookbook/exclusion_strategies#overriding-groups-of-deeper-branches-of-the-graph

basically i fix sorta fixed it by doing this though, anyone got a better way of handling it?

```php

                $groups = array_key_exists($name, $model->getGroups()) && is_array($model->getGroups()[$name])? $model->getGroups()[$name] : $model->getGroups();

        $property->setRef(
            $this->modelRegistry->register(new Model(new Type(Type::BUILTIN_TYPE_OBJECT, false, $type), $groups))
         );
```

GuilhemN:
Fixed in https://github.com/nelmio/NelmioApiDocBundle/pull/1132

PR #1132, opened 2017-11-27
Fix complex groups context for jms models

7thcubic:
#1112

### DIFF
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -95,8 +95,14 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
                     continue;
                 }
 
+                if(empty($model->getGroups()) || !array_key_exists($name, $model->getGroups()) || !is_array($model->getGroups()[$name]))
+                    $groups = $model->getGroups();
+                else
+                    $groups = $model->getGroups()[$name];
+
+
                 $property->setRef(
-                    $this->modelRegistry->register(new Model(new Type(Type::BUILTIN_TYPE_OBJECT, false, $type), $model->getGroups()))
+                    $this->modelRegistry->register(new Model(new Type(Type::BUILTIN_TYPE_OBJECT, false, $type), $groups))
                 );
             }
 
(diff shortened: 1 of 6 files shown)


ModelDescriber/JMSModelDescriber.php
```diff
@@ -87,8 +87,10 @@ public function describe(Model $model, Schema $schema)
                     continue;
                 }
 
+                $groups = array_key_exists($name, $model->getGroups()) && is_array($model->getGroups()[$name])? $model->getGroups()[$name] : $model->getGroups();
```
GuilhemN: There's a space missing after the `is_array` call.

7thcubic: space??? no its correct it is to make sure that it is an array

7thcubic:
not sure if it is me but at least for my use case like what is set on the jms controller the group context doesn't seem to be correct. 

because when serializing a child object from a parent object, the context used seems to be the context of the parent object, which means you would not have control of child object context groups like what JMS depicts, meaning if child and parent has the same group name the configuration will bleed over.

ModelDescriber/JMSModelDescriber.php
GuilhemN: I'm talking about a CS issue.

ModelDescriber/JMSModelDescriber.php
```diff
@@ -95,8 +95,14 @@ public function describe(Model $model, Schema $schema)
                     continue;
                 }
 
+                if(empty($model->getGroups()) || !array_key_exists($name, $model->getGroups()) || !is_array($model->getGroups()[$name]))
```
GuilhemN: The `empty` check is useless

ModelDescriber/JMSModelDescriber.php
```diff
@@ -95,8 +95,14 @@ public function describe(Model $model, Schema $schema)
...
+                    $groups = $model->getGroups();
+                else
```
GuilhemN: Please use brackets

GuilhemN:
> not sure if it is me but at least for my use case like what is set on the jms controller the group context doesn't seem to be correct. 

Indeed I did not think about this feature when I created this class.

GuilhemN:
Thank you @7thcubic!

PR #1133, opened 2017-11-29
Fixing missing documentation on Markdown output

Smolli:
There is a little bug with the markdown generation by missing documentation. 

Adding the documentation section only when it's the same as the description does not make sense. So the documentation is missing when you create the markdown output.

This little tweak fixes it.


GuilhemN:
Thank you @Smolli!

PR #1134, opened 2017-12-04
[2.x] Fix compatibility with Symfony 3.4 and 4.0

dunglas:
Tests will be red because they already are, but the API Platform test suite pass with those changes. I suggest to merge this patch as is.


GuilhemN:
Thank you @dunglas! :)

PR #1139, opened 2017-12-10
fix route parameter not required

scasei:
Route  parameters are set to always required, but there are use cases and in symfony is it possible to declare them not required. This should fix it.


GuilhemN:
This is unfortunately not supported by swagger 2.0, they must be required with it.

scasei:
Ok, I see, sorry

PR #1145, opened 2017-12-15
[3.0] Pass real property to swagger annotation reader

discordier:
The type invariation (changing from the real property to the sub items
definition) causes the swagger annotations to end up on the created
`Items` entry.
This changes the behaviour to update the real property entry with the
swagger annotation.

Additionally this ensures that "description", "title" and "example" are
only updated on `Schema` instances.


GuilhemN:
Thank you @discordier!

PR #1148, opened 2017-12-18
make upgrade instructions depend on stable version

dbu:
now that you have a stable release, people do not need to depend on the beta version anymore.

(and afaik composer recommends to use ^ rather than ~ when working with libraries that respect semver)


GuilhemN:
Indeed, thank you @dbu!

Issue #1149, opened 2017-12-18
error when symfony form component is not activated


dbu:
if the form component is not activated in the symfony framework bundle, but the code is available, we get an error `The service "nelmio_api_doc.generator" has a dependency on a non-existent service "form.factory".`

the correct thing would be to do a compiler pass that checks the container for form.factory and only adds the form model describer if forms is enabled.

dbu:
i can do a pull request for this if you agree that this is the correct approach.

GuilhemN:
That's indeed a problem... See https://github.com/nelmio/NelmioApiDocBundle/pull/1154 :)

dbu:
will be fixed by #1154

PR #1150, opened 2017-12-18
Require the Symfony/form component

thePanz:
The Symfony Form is required for this bundle to work, require it directly.


dbu:
i think the problem is what i reported as #1149 - it does not matter whether the form component is available or not.

thePanz:
As discussed with @dbu , this MR is not fixing the issue :)
Sorry for reporting!

PR #1151, opened 2017-12-18
Support model field required information from swagger-php class annotation

dbu:
the swg definition annotation on the class level can specify whether a property of a model is required or not. e.g.

```
/**
 * @SWG\Definition(required={"id", "displayName", "name"})
 */
```


GuilhemN:
Is it possible to have a test to ensure it works fine ? :)

Mairu:
Should be closed in favor of https://github.com/nelmio/NelmioApiDocBundle/pull/1196

GuilhemN:
Fixed in https://github.com/nelmio/NelmioApiDocBundle/pull/1196 🎉

PR #1152, opened 2017-12-18
parse phpdoc annotations in jms models

dbu:
The jms metadata model seems to not pick up descriptions. Imho we should pick them up, i played around with this code and hope for feedback. If the maintainers agree, i am happy to wrap up the pull request. Potentially it could also make sense to use the PhpdocAnnotationReader together with the form model generator, in case it does not pick all of the phpdoc.

### DIFF
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -103,6 +103,8 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
 
             // read property options from Swagger Property annotation if it exists
             if (null !== $item->reflection) {
+                $phpdocReader = new PhpdocAnnotationReader();
+                $phpdocReader->updateWithPhpdoc($item->reflection, $realProp);
                 $this->swaggerPropertyAnnotationReader->updateWithSwaggerPropertyAnnotation($item->reflection, $realProp);
             }
         }
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -35,14 +35,18 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
 
     private $swaggerPropertyAnnotationReader;
 
+    private $phpdocPropertyAnnotationsReader;
+
     public function __construct(
         MetadataFactoryInterface $factory,
         PropertyNamingStrategyInterface $namingStrategy,
-        SwaggerPropertyAnnotationReader $swaggerPropertyAnnotationReader
+        SwaggerPropertyAnnotationReader $swaggerPropertyAnnotationReader,
+        PhpdocPropertyAnnotationReader $phpdocPropertyAnnotationReader
     ) {
         $this->factory = $factory;
         $this->namingStrategy = $namingStrategy;
         $this->swaggerPropertyAnnotationReader = $swaggerPropertyAnnotationReader;
+        $this->phpdocPropertyAnnotationsReader = $phpdocPropertyAnnotationReader;
     }
 
     /**
@@ -103,8 +107,7 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
 
             // read property options from Swagger Property annotation if it exists
             if (null !== $item->reflection) {
-                $phpdocReader = new PhpdocAnnotationReader();
-                $phpdocReader->updateWithPhpdoc($item->reflection, $realProp);
+                $this->phpdocPropertyAnnotationsReader->updateWithPhpdoc($item->reflection, $realProp);
                 $this->swaggerPropertyAnnotationReader->updateWithSwaggerPropertyAnnotation($item->reflection, $realProp);
             }
         }
--- a/ModelDescriber/PhpdocAnnotationReader.php
+++ b/ModelDescriber/PhpdocAnnotationReader.php
@@ -1,0 +1,61 @@
+<?php
+
+/*
+ * This file is part of the NelmioApiDocBundle package.
+ *
+ * (c) Nelmio
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Nelmio\ApiDocBundle\ModelDescriber;
+
+use EXSyst\Component\Swagger\Schema;
+use EXSyst\Component\Swagger\Items;
+use phpDocumentor\Reflection\DocBlock\Tags\Var_;
+use phpDocumentor\Reflection\DocBlockFactory;
+use phpDocumentor\Reflection\DocBlockFactoryInterface;
+
+/**
+ * @internal
+ */
+class PhpdocAnnotationReader
+{
+    private $docBlockFactory;
+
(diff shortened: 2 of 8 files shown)


ModelDescriber/JMSModelDescriber.php
```diff
@@ -103,6 +103,8 @@ public function describe(Model $model, Schema $schema)
 
             // read property options from Swagger Property annotation if it exists
             if (null !== $item->reflection) {
+                $phpdocReader = new PhpdocAnnotationReader();
```
dbu: this would need to be injected instead. i wait for feedback if the general approach is ok and if so will do that part.

ModelDescriber/PhpdocAnnotationReader.php
```diff
@@ -0,0 +1,61 @@
...
+     * @param \ReflectionProperty $reflectionProperty
+     * @param Items|Schema        $property
+     */
+    public function updateWithPhpdoc(\ReflectionProperty $reflectionProperty, $property)
+    {
+        try {
+            $docBlock = $this->docBlockFactory->create($reflectionProperty);
+            if (!$title = $docBlock->getSummary()) {
+                /** @var Var_ $var */
+                foreach ($docBlock->getTagsByName('var') as $var) {
```
dbu: technically there could be more than one `@var` in a property docblock. i chose to pick the first one that has a description.

ModelDescriber/PhpdocAnnotationReader.php
```diff
@@ -0,0 +1,61 @@
...
+                    if ($title) break;
+                }
+            }
+            if ($property->getTitle() === null && $title) {
+                $property->setTitle($title);
+            }
+            if ($property->getDescription() === null && $docBlock->getDescription()) {
+                $property->setDescription($docBlock->getDescription()->render());
+            }
+        } catch (\Exception $e) {
```
GuilhemN: We only need to catch errors when reading the php doc block, right ?

README.md
```diff
@@ -151,7 +151,7 @@ serialization groups when using the Symfony serializer.
 ### If you're using the JMS Serializer
 
 The metadata of the JMS serializer are used by default to describe your
-models. Note that PHP doc blocks aren't supported in this case.
```
GuilhemN: I think we should keep this comment, it is about the fact that properties type is not detected based on php docs when using the jms serializer (should probably be reworded though).

ModelDescriber/PhpdocAnnotationReader.php
dbu: indeed, this risks hiding errors. fixing

README.md
dbu: i added a check to use the type from the `@var` if not already specified by jms.

dbu:
okay, here we go, added dependency injection and adressed your feedback.

dbu:
shall i squash the commits?

also, i feel like the best approach would be to have tagged description extractors with priorities and then run all active extractors in order. rather than put extra tasks into specific extractors. but that would be a massive refactoring that would end up with code duplication to preserve BC.

Resources/config/services.xml
```diff
@@ -54,6 +54,12 @@
             <argument type="service" id="annotation_reader" />
         </service>
 
+        <service
+                id="nelmio_api_doc.model_describers.phpdoc_property_annotation_reader"
+                class="Nelmio\ApiDocBundle\ModelDescriber\PhpdocPropertyAnnotationReader"
```
GuilhemN: As it's used in only one place I'd rather just instantiate it in the describer's constructor.

ModelDescriber/PhpdocPropertyAnnotationReader.php
```diff
@@ -0,0 +1,76 @@
...
+            $property->setTitle($title);
+        }
+        if ($property->getDescription() === null && $docBlock->getDescription() && $docBlock->getDescription()->render()) {
+            $property->setDescription($docBlock->getDescription()->render());
+        }
+        if ($property->getType() === null) {
+            /** @var Var_ $var */
+            foreach ($docBlock->getTagsByName('var') as $var) {
+                if ($var->getType()) {
+                    $property->setType($var->getType());
```
GuilhemN: This is misleading imo, when using symfony's PropertyInfo component it supports far more cases. I think supporting this is up to the jms serializer

ModelDescriber/PhpdocPropertyAnnotationReader.php
```diff
@@ -0,0 +1,76 @@
...
+            $docBlock = $this->docBlockFactory->create($reflectionProperty);
+        } catch (\Exception $e) {
+            // ignore
+            return;
+        }
+
+        if (!$title = $docBlock->getSummary()) {
+            /** @var Var_ $var */
+            foreach ($docBlock->getTagsByName('var') as $var) {
+                if (!$description = $var->getDescription()) continue;
```
GuilhemN: Please use brackets to keep a consistent code :)

GuilhemN:
> shall i squash the commits?

No need, github does it automatically :)

> also, i feel like the best approach would be to have tagged description extractors with priorities and then run all active extractors in order. rather than put extra tasks into specific extractors. but that would be a massive refactoring that would end up with code duplication to preserve BC.

Yeah not sure how to deal with this :/ Probably having a DescriptionParser for all describers would be better.

Resources/config/services.xml
dbu: why that? its a private service, so no harm done declaring this here. i find it much clearer to have all services defined in  service configuration files, rather than in php code.

the best solution would probably be to define the services inline in the main service definition of the jms describer which itself could be a service definition in the config file. then we could either split that into its own file and only load when enabled, or remove it in the extension if not needed.

wdyt?

ModelDescriber/PhpdocPropertyAnnotationReader.php
dbu: indeed, willdo

ModelDescriber/PhpdocPropertyAnnotationReader.php
dbu: it could also majorly mess up if the var is a php class or just a random typo... removed again.

dbu:
> Yeah not sure how to deal with this :/ Probably having a DescriptionParser for all describers would be better.

yeah some main class that creates the swagger object and allows to register describers and then runs them in the right order to augment the swagger information. i'll see if i find time to propose a deeper refactoring.

Resources/config/services.xml
GuilhemN: Imo public classes should be instantiable by the user but my point is kind of broken as we already disrespect it... Not sure we should disrespect it even more though but if you think it's better this way I won't insist, let's keep it as is :)

> the best solution would probably be to define the services inline in the main service definition of the jms describer which itself could be a service definition in the config file. then we could either split that into its own file and only load when enabled, or remove it in the extension if not needed.

Why not, could indeed be clearer :)

GuilhemN:
Ok, thanks! :) I'll try to spend some time fixing issues as well next two weeks.

GuilhemN:
Thank you @dbu! :)

PR #1153, opened 2017-12-18
fix date type detection with jms model describer

dbu:
when i debugged, jms seems to have transformed an annotation of `\DateTime` into `'date'`.


GuilhemN:
That's weird, I wasn't able to find an information about this... In the [`DateHandler`](https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/Handler/DateHandler.php), `date` is not supported. 
Could it be an error in your code ?

dbu:
cannot reproduce myself anymore. sorry for the noise

PR #1154, opened 2017-12-18
Enable the FormModelDescriber only if forms are enabled

GuilhemN:
fix https://github.com/nelmio/NelmioApiDocBundle/issues/1149


GuilhemN:
Comment added, thanks :)

PR #1155, opened 2017-12-18
[3.0] Add integer support to FormModelDescriber

discordier:
Integer support was missing so far.


GuilhemN:
Thank you @discordier!

PR #1162, opened 2017-12-28
add changelog listing changes since 3.0

dbu:
i suggest we also ask people who do pull requests to add their notes to the changelog to make it more manageable for people to know what is going on.


GuilhemN:
Thank you @dbu! And sure, it would indeed be better to fulfill the changelog at the same time we add features.

PR #1166, opened 2017-12-31
Fix undefined variable  in ApiPlatformDescriber

kubaceg:
When I run fresh installation of NelmioApiDocBundle I have such error:

"Notice: Undefined variable: paths" in vendor/nelmio/api-doc-bundle/Describer/ApiPlatformDescriber.php:39

I look at the file and I saw that `$paths` variable is not defined, so I define it as empty array. Bundle with my fix works :)


GuilhemN:
Thank you @kubaceg, and happy new year! :)

PR #1169, opened 2018-01-03
Add areas support

GuilhemN:
Should fix https://github.com/nelmio/NelmioApiDocBundle/pull/1105#pullrequestreview-86355879.


composer.json
```diff
@@ -16,27 +16,26 @@
     ],
     "require": {
         "php": "^7.0",
-        "symfony/framework-bundle": "^3.2.5|^4.0",
-        "symfony/property-info": "^3.1|^4.0",
+        "symfony/framework-bundle": "^3.4|^4.0",
```
dbu: i would do this cleanup in a separate pull request that does not add new features.

Tests/Functional/Entity/JMSComplex.php
```diff
@@ -1,31 +1,32 @@
...
  */
 
 namespace Nelmio\ApiDocBundle\Tests\Functional\Entity;
 
 use JMS\Serializer\Annotation as Serializer;
 
 /**
- * Class JMSComplex
- * @package Nelmio\ApiDocBundle\Tests\Functional\Entity
+ * Class JMSComplex.
```
dbu: this does not tell anything the code does not already say. i'd remove the text line completely. (and these things part of the cleanup pull request rather than the new feature.)

DependencyInjection/NelmioApiDocExtension.php
```diff
@@ -54,22 +58,49 @@ public function load(array $configs, ContainerBuilder $container)
...
+            $container->register(sprintf('nelmio_api_doc.describers.route.%s', $area), RouteDescriber::class)
+                ->setPublic(false)
+                ->setArguments([
+                    new Reference(sprintf('nelmio_api_doc.routes.%s', $area)),
+                    new Reference('nelmio_api_doc.controller_reflector'),
+                    new TaggedIteratorArgument('nelmio_api_doc.route_describer'),
+                ])
+                ->addTag(sprintf('nelmio_api_doc.describer.%s', $area), ['priority' => -400]);
+
+            $container->register('nelmio_api_doc.describers.swagger_php', SwaggerPhpDescriber::class)
```
dbu: i find it an antipattern to define services in the DI extension instead of in the services xml files. its much harder to read what is going on imho. i would rather have additional services xml files for specific features that are only loaded if that feature is to be enabled. but i guess its a question of taste as well...

DependencyInjection/Configuration.php
```diff
@@ -21,20 +21,49 @@ public function getConfigTreeBuilder()
...
+                })
+            ->end()
             ->children()
                 ->arrayNode('documentation')
                     ->useAttributeAsKey('key')
                     ->info('The documentation used as base')
                     ->example(['info' => ['title' => 'My App']])
                     ->prototype('variable')->end()
                 ->end()
-                ->arrayNode('routes')
```
dbu: this will need to be documented in the changelog

DependencyInjection/Configuration.php
```diff
@@ -21,20 +21,49 @@ public function getConfigTreeBuilder()
         $treeBuilder = new TreeBuilder();
         $treeBuilder
             ->root('nelmio_api_doc')
+            ->beforeNormalization()
+                ->ifTrue(function ($v) {
+                    return !isset($v['areas']) && isset($v['routes']);
+                })
+                ->then(function ($v) {
+                    $v['areas'] = $v['routes'];
+                    unset($v['routes']);
```
dbu: this should also trigger a deprecation warning

DependencyInjection/Configuration.php
```diff
@@ -21,20 +21,49 @@ public function getConfigTreeBuilder()
         $treeBuilder = new TreeBuilder();
         $treeBuilder
             ->root('nelmio_api_doc')
+            ->beforeNormalization()
+                ->ifTrue(function ($v) {
+                    return !isset($v['areas']) && isset($v['routes']);
```
dbu: if both areas and routes are set, we should throw an invalid configuration exception explaining to use only the new `areas` section. otherwise the user will just see an error that routes is not known, which will be confusing when before, routes did work.

README.md
```diff
@@ -77,9 +77,9 @@ routes that are documented by configuring the bundle:
 ```yml
 # app/config/config.yml
 nelmio_api_doc:
-    routes:
+    areas:
```
dbu: does the areas model allow to specify other criteria than path patterns? and how to specify more than one distinct area?

GuilhemN: > does the areas model allow to specify other criteria than path patterns?

For now it's only based on path patterns but we could easily add more criteria if we want to, I'd rather do this later though to not complicate this PR. The best imo would be to allow to add a custom routes filter.

> and how to specify more than one distinct area?

You can define as many areas as you want:
```yml
nelmio_api_doc:
    areas:
        default: [ path_patterns: [ ^/api ] ]
        internal: [ path_patterns: [ ^/internal ] ]
        commercial: [ path_patterns: [ ^/commercial ] ]
        # ...
```

DependencyInjection/NelmioApiDocExtension.php
GuilhemN: I tend to agree but this services are dynamically generated and we can't put them in a file unfortunately... Maybe using the php syntax would be clearer though (see https://github.com/symfony/symfony/pull/23834).

dbu: its ok, i can live with it as is, if we can't move it to the xml :-)

README.md
dbu: can you add a doc section for configuring several areas? so if you specify what we have here, you implicitly create a default area, and if you would explicitly create a list of areas, you get several?

DependencyInjection/Configuration.php
GuilhemN: added

GuilhemN:
I added tests and docs. Much bigger diff :/

ApiDocGenerator.php
```diff
@@ -46,7 +46,7 @@ public function generate(): Swagger
         }
 
         if ($this->cacheItemPool) {
-            $item = $this->cacheItemPool->getItem('swagger_doc');
+            $item = $this->cacheItemPool->getItem($cacheItemId ?? 'swagger_doc');
```
dbu: i think this will never do anything. ?? is like isset, not telling you that the variable does not exist. but cacheItemId will never exist in this context unless i am missing something. did you mean to store the constructor parameter andin the class and use $this->cacheItemId?

Controller/DocumentationController.php
```diff
@@ -12,21 +12,37 @@
...
+    /**
+     * @param ContainerInterface $generatorLocator
+     */
+    public function __construct($generatorLocator)
     {
-        $this->apiDocGenerator = $apiDocGenerator;
+        if (!$generatorLocator instanceof ContainerInterface) {
+            @trigger_error(sprintf('Providing an instance of "%s" to "%s()" is deprecated since version 3.1. Provide it an instance of "%s" instead.', ApiDocGenerator::class, __METHOD__, ContainerInterface::class), E_USER_DEPRECATED);
+            $generatorLocator = new ServiceLocator(['default' => function () use ($generatorLocator) {
+                return $generatorLocator;
```
dbu: this will only work if $generatorLocator is the ApiDocGenerator, right? i would typehint both and explain that the generator is still supported for BC. and do a typecheck before the deprecation to fail loudly if its not something we can work with.

ApiDocGenerator.php
GuilhemN: Indeed, I added tests to make sure it works. Will have to use it one day to have better perfs...

Controller/DocumentationController.php
GuilhemN: type hint added :)

GuilhemN:
Thanks for all your reviews! :)

Issue #1140, opened 2017-12-11
[3.0] API Platform + NelmioApiDocBundle prefix issues


simshaun:
I'm using API Platform + NelmioApiDocBundle.

The API Platform's built-in docs are correct and look like this:
![image](https://user-images.githubusercontent.com/379519/33824698-3d2854ba-de2d-11e7-91e2-4eec8cb51430.png)

And this is what the Nelmio API docs look like. It's adding an extra `/api`.
![image](https://user-images.githubusercontent.com/379519/33824720-50b599d4-de2d-11e7-8aef-e3d4209143c2.png)

My API Platform routing is:

```yaml
api_platform:
    resource: .
    type: api_platform
    prefix: /api
```

MJBGO:
I have the same issue :(

https://github.com/nelmio/NelmioApiDocBundle/issues/1128

GuilhemN:
Not sure what to do here... I have to investigate why this is happening but I do not have much time these days :/
Would greatly appreciate help here :)

7thcubic:
not sure exactly what is happening but from my understanding.

routing prefix: /foo
controller route: /bar
action route: /action

will generate /foo/bar/action which is correct.

so not entirely sure what was being generated by the api platform bridge, need to investigate

GuilhemN:
Should be fixed by #1172

Issue #1128, opened 2017-11-13
[3.0] BasePath doubled


MJBGO:
Hi,

I've been searching for a solution but could not find one ...

I've set the basePath parameter to "/api/v1/" and my routing.yml looks like this : 

![image](https://user-images.githubusercontent.com/8344487/32705844-a00e843a-c818-11e7-8ff2-7bb5ecd6d384.png)

But all my paths still have the "/api/v1" prefix and when I execute them, the path is doubled "/api/v1/api/v1" ...

![image](https://user-images.githubusercontent.com/8344487/32705892-09579346-c819-11e7-817f-194e5d932220.png)

Is it a bug or am I doing something wrong ?

Thanks

GuilhemN:
You did set it manually ? If so this bundle does not take it into account.

7thcubic:
can you post the annotation on the controller actions?

GuilhemN:
Should be fixed by https://github.com/nelmio/NelmioApiDocBundle/pull/1172

Bcow7:
Hi,
I have the exact same issue with commit `279d20b`.

Here is my configuration in /nelmio_api_doc.yaml
```
nelmio_api_doc:
    documentation:
        host: my-api.example.org
        basePath: /api/v2/
        schemes: [https]
        info:
            title: API Doc test
            description: This is an awesome app!
            version: 1.0.0
    areas: # to filter documented areas
        default:
            path_patterns:
                - ^/api/v2(?!/doc*) # Accepts routes under /api except /api/doc

```

and Controller
![image](https://user-images.githubusercontent.com/2751121/55243120-de261b00-523e-11e9-83d2-3ea0a1615bad.png)

![image](https://user-images.githubusercontent.com/2751121/55242932-8d162700-523e-11e9-80d7-ff4e46c63177.png)

Am I doing something wrong?

Thank you very much!

7thcubic:
pretty sure it is settings error, because u are setting the base path to be

my-api.example.org/api/v2

and using a controller that has a base route of 

/api/v2

which means the current swagger generated is correct.

Try setting your base path to be: my-api.example.org

GuilhemN:
@Bcow7 the basePath you configured manually isn't changing how routes are registered, so you get the path doubled.
Following @7thcubic instruction should solve your issue.

PR #1172, opened 2018-01-04
Fix Api-Platform support

GuilhemN:
Fixes https://github.com/nelmio/NelmioApiDocBundle/issues/1140, https://github.com/nelmio/NelmioApiDocBundle/issues/1128


dbu:
can this not be fixed in the 2.0 branch so that we can release it immediately?

GuilhemN:
Well the 2.x branch is the 2.0 :) We would need a 3.0 branch but it doesn't exist, I guess we can just release 3.1 sooner?

Issue #1168, opened 2018-01-03
How to create array model?


Gemorroj:
```php
    /**
     * @Route("/api/orders", methods={"GET"})
     * @SWG\Response(
     *     response=200,
     *     description="Returns all orders",
     *     @SWG\Schema(type="array", @Model(type=Order::class)),
     * )
     * @SWG\Tag(name="orders")
     */
    public function listAction()
    {
        $repository = $this->getDoctrine()->getRepository(Order::class);
        return $repository->findAll();
    }
```

I have exception `Call to a member function getBuiltinType() on null at vendor\nelmio\api-doc-bundle\ModelDescriber\ObjectModelDescriber.php:71`

bliszka:
I got the same exception for my model class

```php
    public function getBuckets(): array
    {
        return BucketTypes::getBucketsByStrategy(
            $this->getBucketsStrategy()
        );
    }
```

Propably in `ObjectModelDescriber.php`  `else` statment is missed.
Right now is: 
```php
  if (Type::BUILTIN_TYPE_ARRAY === $type->getBuiltinType()) {
      $type = $type->getCollectionValueType();
      $property->setType('array');
      $property = $property->getItems();
  }

  if (Type::BUILTIN_TYPE_STRING === $type->getBuiltinType()) {
      $property->setType('string');
  } elseif ......
```

Maybe it should be: 
```php
  if (Type::BUILTIN_TYPE_ARRAY === $type->getBuiltinType()) {
      $type = $type->getCollectionValueType();
      $property->setType('array');
      $property = $property->getItems();
  } elseif (Type::BUILTIN_TYPE_STRING === $type->getBuiltinType()) {
      $property->setType('string');
  } elseif ......
```

Gemorroj:
@GuilhemN any ideas? 🤔

GuilhemN:
Well the problem is that the bundle can't deduce the sub type of your property so it breaks (it understands it's an array but doesn't know an array of what). You have to update your code to be more precise but we also definitelly need a proper exception here.

PR #1174, opened 2018-01-08
Add an exception

GuilhemN:
Add an exception to fail properly.

Fixes https://github.com/nelmio/NelmioApiDocBundle/issues/1168#issuecomment-355964204

### DIFF
--- a/ModelDescriber/ObjectModelDescriber.php
+++ b/ModelDescriber/ObjectModelDescriber.php
@@ -51,6 +51,21 @@ class ObjectModelDescriber implements ModelDescriberInterface, ModelRegistryAwar
         }
 
         foreach ($propertyInfoProperties as $propertyName) {
+            $property = $properties->get($propertyName);
+
+            // read property options from Swagger Property annotation if it exists
+            if (property_exists($class, $propertyName)) {
+                $this->swaggerPropertyAnnotationReader->updateWithSwaggerPropertyAnnotation(
+                    new \ReflectionProperty($class, $propertyName),
+                    $property
+                );
+            }
+
+            // If type manually defined
+            if (null !== $property->getType()) {
+                continue;
+            }
+
             $types = $this->propertyInfo->getTypes($class, $propertyName);
             if (0 === count($types)) {
                 throw new \LogicException(sprintf('The PropertyInfo component was not able to guess the type of %s::$%s', $class, $propertyName));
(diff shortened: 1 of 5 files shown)


ModelDescriber/ObjectModelDescriber.php
```diff
@@ -64,6 +64,9 @@ public function describe(Model $model, Schema $schema)
 
             if (Type::BUILTIN_TYPE_ARRAY === $type->getBuiltinType()) {
                 $type = $type->getCollectionValueType();
+                if (null === $type) {
+                    throw new \LogicException(sprintf('The sub type of the property "%s::$%s" is not detected by the PropertyInfo Component. Please update your code to make sure it is.', $class, $propertyName));
```
dbu: ...to make sure it is. => can we be more specific? i would say 'Property "%s:%s" is an array, but no indication of the array elements are made. Use e.g. string[] for an array of string.' (if i understand correctly, this is the thing to do in this context, right? could we additionally suggest to use a swagger annotation?)

Mairu: It will not be possible to create an associative array this way (json object in doc). It would really be better to check for swagger annotation before throwing an error when "array" is used.

dbu: then we should probably validate later, after all describers have been run?

Mairu: Yeah that is the thing i would suggest too. Because the internal type/doc block used in an application that is parsed by the PropertyInfo Component could be different to type you want to use for your documentation / external api.

GuilhemN: I guess we could instead execute https://github.com/nelmio/NelmioApiDocBundle/blob/7af3210b6f173abb959ddeaf46b162fa1b3ae122/ModelDescriber/ObjectModelDescriber.php#L96-L102 earlier and try to guess the type only if it's not manually defined.

dbu: the fancy thing could be if each describer can register warnings on a field, and at the end there is a validation step that checks if properties are correctly described, and if not prints all warnings for that property...

i am a bit worried when order of execution for the describers becomes too important (its of course relevant for priority about which description wins if several are provided).

GuilhemN: Well model describers must register fully a model. I don't think such system would be very useful right now, maybe it would be if we added the possibility to describe properties individually though (for validation constraints for instance).

GuilhemN:
> i think we should untangle those describers and not make one call the other, but just chain them. 

I'm not sure how we could do this but it could be interesting indeed :)

PR #1177, opened 2018-01-09
Provide example for action summary and description

### LINKED ISSUE #1176: [3.0] Can't describe action in controller
there is no way to describe what endpoint actually do. in version 2 there was `description` property in `@ApiDoc` annotation
```
...
* @ApiDoc(
*  resource=true,
*  description="This is a description of your API method",
...
```

how can it be done now?

dbu:
fix #1176

@GuilhemN is this correct?

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -123,6 +123,10 @@ To document your routes, you can use the SwaggerPHP annotations and the
     class UserController
     {
         /*
+         * List the rewards of the specified user.
+         *
+         * This call takes into account all confirmed awards, but not pending or refused awards.
+         *
          * @Route("/api/{user}/rewards", methods={"GET"})
          * @SWG\Response(
          *     response=200,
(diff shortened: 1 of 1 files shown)


Resources/doc/index.rst
```diff
@@ -123,6 +123,10 @@ To document your routes, you can use the SwaggerPHP annotations and the
     class UserController
     {
         /*
+         * List the rewards of the specified user.
+         *
+         * This call takes into account all confirmed awards, but not pending or refused awards.
+         *
          * @Route("/api/{user}/rewards", methods={"GET"})
```
dbu: @GuilhemN does this actually work? the `user` parameter is not explained in the annotation.

GuilhemN: It is automatically added by the bundle to the documentation (as a string parameter).

dbu: okay. we should maybe say that, or even add an example for the in=path parameter annotation.

GuilhemN: Indeed, the docs are very minimal right now :)

GuilhemN:
Thank you @dbu :)

PR #1185, opened 2018-01-12
Make the docs clearer about models

GuilhemN:


Resources/doc/index.rst
```diff
@@ -168,19 +168,29 @@ Use models
 As shown in the example above, the bundle provides the ``@Model`` annotation.
 When you use it, the bundle will deduce your model properties.
 
+.. note::
+
+    A model can be a form type, an entity or any object.
```
dbu: We could be more specific while we are at it. ... a Symfony form type, a Doctrine ORM entity or a general PHP object.

(Am i correct assuming that the doctrine odm annotations are not used?)

Resources/doc/index.rst
```diff
@@ -191,8 +201,8 @@ It has two options:
 
         /**
          * @SWG\Response(
-         *   response="200",
-         *   description="Success",
+         *   response="200"
```
dbu: Isn't the comma needed here?

Resources/doc/index.rst
GuilhemN: > We could be more specific while we are at it. ... a Symfony form type, a Doctrine ORM entity or a general PHP object.

+1, updated

> (Am i correct assuming that the doctrine odm annotations are not used?)

 You mean ``Doctrine orm annotation``? The extractor is not easy to find but they are extracted, see https://github.com/symfony/doctrine-bridge/blob/master/PropertyInfo/DoctrineExtractor.php.

Resources/doc/index.rst
GuilhemN: indeed :)

dbu:
Thanks.

The bridge extractor phpdoc talks about odm metadata, but i dont think it looks at any of the odm's. Thus it seems correct to only talk about orm.

GuilhemN:
Ok let's keep your reword as is then :)

PR #1188, opened 2018-01-15
Made nelmio_api_doc.generator a public service

michellesanver:
The service was not defined public or private, which threw deprecation notices in the tests where we use it from the container. It is now declared public explicitly to not have deprecations.

### DIFF
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -15,7 +15,7 @@
         </service>
 
         <!-- Swagger Spec Generator -->
-        <service id="nelmio_api_doc.generator" alias="nelmio_api_doc.generator.default" />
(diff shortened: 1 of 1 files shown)


michellesanver:
It's deprecations in the tests indeed, specifically in "WebTestCase" where we have this: 

```
    protected function getSwaggerDefinition()
    {
        static::createClient();

        return static::$kernel->getContainer()->get('nelmio_api_doc.generator')->generate();
    }
```

GuilhemN:
Thank you @michellesanver!

> however, as this is the main entry point to the bundle, i think making it a public service sounds reasonable to me regardless of my above question.

That's why it was made public in the first place :)
Maybe we should update our tests to use dependency injection, but imo it's not worth the effort.

dbu:
thanks michelle!

PR #1198, opened 2018-01-22
add php 7.2

andreybolonin:


.travis.yml
```diff
@@ -19,13 +22,10 @@ branches:
 
 matrix:
   fast_finish: true
-  include:
-    - php: 7.0
-      env: COMPOSER_FLAGS="--prefer-lowest"
```
dbu: why do you remove the lowest version build? this shows if the composer constraints are correct or allow to install with too old versions of a dependency.

.travis.yml
```diff
@@ -19,13 +22,10 @@ branches:
...
 
 before_install:
     - phpenv config-rm xdebug.ini
 
 install: composer update $COMPOSER_FLAGS --prefer-stable
 
-script: ./phpunit
+script: vendor/bin/phpunit
```
dbu: please revert this line so that we continue using the symfony phpunit bridge

composer.json
```diff
@@ -36,6 +36,7 @@
         "symfony/cache": "^3.4|^4.0",
         "symfony/phpunit-bridge": "^3.4",
         "symfony/stopwatch": "^3.4|^4.0",
+        "phpunit/phpunit": "^4.8.35|^5.4.4|^6.0.0",
```
dbu: this is not needed, the phpunit bridge installs phpunit for us.

.travis.yml
```diff
@@ -3,6 +3,9 @@ language: php
 php:
   - 7.0
   - 7.1
+  - 7.2
+  
+env: SYMFONY_DEPRECATIONS_HELPER=weak_vendors
```
dbu: if you do this in the phpunit.xml.dist file, it will also apply when running tests locally.

Issue #1163, opened 2017-12-28
Question: Adding security schemes to an operation / merge fails


jmglsn:
Hi,
I stuck on trying to add a special security scheme to an operation. I defined one at global level inside the bundle configuration and for one operation I wanted to reset it to an empty array - as it has no security. 

But I don't find a way to define the security scheme. I always get an error `Unable to merge @SWG\\Get() in \\UserController->fetchUserRewardsAction()`.

I'll use the example from this project to illustrate my problem:
```php
namespace AppBundle\Controller;

use AppBundle\Entity\User;
use AppBundle\Entity\Reward;
use Nelmio\ApiDocBundle\Annotation\Model;
use Swagger\Annotations as SWG;
use Symfony\Component\Routing\Annotation\Route;

class UserController
{
    /*
     * @Route("/api/{user}/rewards", methods={"GET"})
     * @SWG\Response(
     *     response=200,
     *     description="Returns the rewards of an user",
     *     @SWG\Schema(
     *         type="array",
     *         @Model(type=Reward::class, groups={"full"})
     *     )
     * )
     * @SWG\Parameter(
     *     name="order",
     *     in="query",
     *     type="string",
     *     description="The field used to order rewards"
     * )
     * @SWG\Tag(name="rewards")
     * // now I try to add a special security info to this operation...
     * @SWG\Get(security={})
     */
    public function fetchUserRewardsAction(User $user)
    {
        // ...
    }
}
```

So it seems that's not possible to merge `@SWG\Get(security={})` with the `@Route` but what's the workaround here?
Thanks in advance!

GuilhemN:
What if you use `@Operation` instead of `@Get`?

vgeyer:
@GuilhemN "Cannot instantiate abstract class Swagger\Annotations\Operation" would be the result auf this. Is there no solution to tell a route "hey you need to be authenticated for this?

Personally, i would expect a  "@SWG\Security(name="api_key")" annotation or something like this. 

My config: 
```
    /**
     * @Rest\Get("/{id}", requirements={"id"="^\d{1,}$"})
     * @SWG\Response(
     *     response=200,
     *     description="Returns informations for a specific customer.",
    *      @Model(type=App\Entity\Customer\Customer::class)
     * )
     * @SWG\Parameter(
     *     name="id",
     *     in="query",
     *     type="integer",
     *     description="The customer id."
     * )
     * @SWG\Tag(name="Customer")
[...]

PR #1201, opened 2018-01-23
Added support for `@Security` annotations

dragosprotung:
This addresses #1163

### DIFF
--- a/Describer/SwaggerPhpDescriber.php
+++ b/Describer/SwaggerPhpDescriber.php
@@ -144,13 +144,26 @@ final class SwaggerPhpDescriber extends ExternalDocDescriber implements ModelReg
                 $implicitAnnotations[] = $annotation;
             }
 
-            if (0 === count($implicitAnnotations) && 0 === count($tags)) {
+            if (0 === count($implicitAnnotations) && 0 === count($tags) && 0 === count($security)) {
                 continue;
             }
 
             foreach ($httpMethods as $httpMethod) {
                 $annotationClass = $operationAnnotations[$httpMethod];
-                $operation = new $annotationClass(['_context' => $context, 'path' => $path, 'value' => $implicitAnnotations, 'tags' => $tags, 'security' => $security]);
+                $constructorArg = [
+                    '_context' => $context,
+                    'path' => $path,
+                    'value' => $implicitAnnotations,
+                ];
+
+                if (0 !== count($tags)) {
+                    $constructorArg['tags'] = $tags;
+                }
+                if (0 !== count($security)) {
+                    $constructorArg['security'] = $security;
+                }
+
+                $operation = new $annotationClass($constructorArg);
                 $analysis->addAnnotation($operation, null);
             }
         }
--- a/Describer/SwaggerPhpDescriber.php
+++ b/Describer/SwaggerPhpDescriber.php
@@ -14,6 +14,7 @@ namespace Nelmio\ApiDocBundle\Describer;
 use Doctrine\Common\Annotations\Reader;
 use EXSyst\Component\Swagger\Swagger;
 use Nelmio\ApiDocBundle\Annotation\Operation;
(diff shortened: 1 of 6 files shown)


dragosprotung:
@GuilhemN not sure how to fix the 2 broken tests. some guidance will be appreciated

GuilhemN:
I think it's an issue with `json_decode` somewhere in the code, for which we should use the second parameter to convert json objects to php arrays. Or maybe it's an issue in exsyst/swagger...

I can't spend a lot of time on this right now but I'll try to help you more a bit later this the week. Thanks for your PR :)

dragosprotung:
Well, that solved it :)
Ready for review

Resources/doc/index.rst
```diff
@@ -153,6 +154,7 @@ To document your routes, you can use the SwaggerPHP annotations and the
          *     description="The field used to order rewards"
          * )
          * @SWG\Tag(name="rewards")
+         * @Security(name="api_key")
```
dbu: can you please mention this in the text that follows, and say a few words what "name" is?

dragosprotung:
@dbu I think this should be called `Security` as it refers to the Security object from the operation (https://swagger.io/specification/#operationObject, https://swagger.io/specification/#securityRequirementObject) and not the security schema (top level)

Resources/doc/index.rst
dragosprotung: Added some description

dragosprotung:
This belongs here because when using only zircote the usage is a bit different:

```php
/**
 * @SWG\Get(
 *     path="/test",
 *     security={{"Bearer":{}}}
 * )
 */
```

Using this bundle:
```php
/**
 * @SWG\Response(
 *     response=200,
 *     description="Returned on success"
 * )
 * @Security(name="Bearer")
 */
```

Resources/doc/index.rst
```diff
@@ -168,6 +170,14 @@ Use models
 As shown in the example above, the bundle provides the ``@Model`` annotation.
 When you use it, the bundle will deduce your model properties.
 
+Use security
```
GuilhemN: Imo this should be in a dedicated article about authentication, it is not needed for basic uses.

dragosprotung: True, security is not covered at all. Shall I remove this section ? Adding documentation related to security is for another PR maybe

dbu: actually yes, and also yes that its out of scope for this PR. maybe lets create an issue to document the "security" things and paste the explanation you wrote in there?

should the annotation be used at all in the example above, if its a special case only?

dragosprotung: For basic usage I think this is not needed. I removed it

Describer/SwaggerPhpDescriber.php
```diff
@@ -141,7 +150,7 @@ private function getAnnotations(): Analysis
 
             foreach ($httpMethods as $httpMethod) {
                 $annotationClass = $operationAnnotations[$httpMethod];
-                $operation = new $annotationClass(['_context' => $context, 'path' => $path, 'value' => $implicitAnnotations, 'tags' => $tags]);
+                $operation = new $annotationClass(['_context' => $context, 'path' => $path, 'value' => $implicitAnnotations, 'tags' => $tags, 'security' => $security]);
```
GuilhemN: should be added only if needed imo. (It seems it's always added to the output right now).

dragosprotung: It's the same with tags. Shall I add a check for both ?

GuilhemN: Sure, should be done as well indeed :)

dragosprotung: done

GuilhemN:
Thank you @dragosprotung!

PR #1202, opened 2018-01-24
Added Symfony constraint annotation reader

martijnvankempen:
This PR adds an annotation reader for the Symfony\Component\Validator\Constraint annotation.
Based on the constraint type certain elements of the properties are set.

The model's property results in the following API documentation.
image

screenshot from 2018-01-24 12-01-15


ModelDescriber/SymfonyConstraintAnnotationReader.php
```diff
@@ -0,0 +1,139 @@
...
+
+        $symfonyConstraintDefined = false;
+
+        foreach ($annotations as $annotation) {
+            if ($annotation instanceof Constraint) {
+                $symfonyConstraintDefined = true;
+            }
+
+            if ($annotation instanceof NotBlank || $annotation instanceof NotNull) {
+                $required[] = $reflectionProperty->getName();
```
GuilhemN: We can just call `$property->setRequired(true)`

ModelDescriber/SymfonyConstraintAnnotationReader.php
```diff
@@ -0,0 +1,139 @@
...
+                $this->appendDescription($property, $annotation->message);
+            }
+
+            if ($annotation instanceof Length) {
+                if ($annotation->min > 0) {
+                    $required[] = $reflectionProperty->getName();
```
GuilhemN: Same here

ModelDescriber/SymfonyConstraintAnnotationReader.php
```diff
@@ -0,0 +1,139 @@
...
+        $symfonyConstraintDefined = false;
+
+        foreach ($annotations as $annotation) {
+            if ($annotation instanceof Constraint) {
+                $symfonyConstraintDefined = true;
+            }
+
+            if ($annotation instanceof NotBlank || $annotation instanceof NotNull) {
+                $required[] = $reflectionProperty->getName();
+                $this->appendDescription($property, $annotation->message);
```
GuilhemN: Making the property required is enough, don't you think ?

martijnvankempen: I agree ;)

ModelDescriber/SymfonyConstraintAnnotationReader.php
martijnvankempen: The red * behind the property doesn't show when only $property->setRequired(true) is called. I think it better to add the required properties to the schema so it is easier to see which properties are required. We could also go for both options. Let me know what you think.

ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
```diff
@@ -0,0 +1,153 @@
...
+        } else {
+            $property->setPattern($newPattern);
+        }
+    }
+
+    /**
+     * Append the description from the constraint to the existing description.
+     *
+     * @param Schema $property
+     * @param string $newDescription
```
GuilhemN: These doc blocks can be removed since the information is already in methods signature.

ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
```diff
@@ -0,0 +1,153 @@
...
+                $property->setMinLength($annotation->min);
+                $property->setMaxLength($annotation->max);
+            }
+
+            if ($annotation instanceof Regex) {
+                $this->appendPattern($property, $annotation->getHtmlPattern());
+            }
+
+            if ($annotation instanceof AbstractComparison) {
+                $this->appendDescription($property, $annotation->message);
```
GuilhemN: I don't think we need this, we could instead use the ``minimum`` and ``maximum`` fields for instance.

ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
```diff
@@ -0,0 +1,153 @@
...
+            if ($annotation instanceof Length) {
+                if ($annotation->min > 0) {
+                    $this->updateSchemaDefinitionWithRequiredProperty($reflectionProperty);
+                }
+
+                $property->setMinLength($annotation->min);
+                $property->setMaxLength($annotation->max);
+            }
+
+            if ($annotation instanceof Regex) {
```
GuilhemN: What about using ``Assert\...`` in this file too?

martijnvankempen: Could you clarify this?

GuilhemN: I mean importing the ``Symfony\Component\Validator\Constraints`` namespace (``use Symfony\Component\Validator\Constraints as Assert;``) to avoid having a use statement for each constraint :)

ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
martijnvankempen: I have decided to remove this condition since the AbstractComparison is used by 16  different constraints. So I will leave it out off scope for now.

GuilhemN:
Great, thank you @martijnvankempen! :)

PR #1205, opened 2018-01-25
Add php 7.2 support

GuilhemN:


GuilhemN:
Some library uses each but I don't know which one... ``weak_vendor`` doesn't silence the error. I'll try to again later.

PR #1207, opened 2018-01-25
add ability to check for presence of default parameter in route for area

wodka:
This would add the ability to restrict paths within the same area to only contain routes that have a specific default set.

ie: 
```php
mespace AppBundle\Controller;

use Symfony\Component\Routing\Annotation\Route;

class AdminController
{
    /*
     * will show in the admin area
     * @Route("/api/execute1", defaults={"doc_area" = "admin"})
     */
    public function executeAction()
    {
        // ...
    }

    /*
     * only in default area
     * @Route("/api/execute2")
     */
    public function executeAction()
    {
        // ...
    }
}
```

with the config:
```yaml
nelmio_api_doc:
    areas:
        default:
            path_patterns: [ ^/api ]
        admin:
            path_patterns: [ ^/api ]
            check_default: doc_area
```

would only show one route in admin and both in the default area


GuilhemN:
On the one hand, it feels weird to use route defaults to define to which documentation area a route belongs but on the other hand, this is a very practical feature... What do you think @dbu?

Maybe providing an `@Area` annotation would be more natural?

dbu:
thanks @wodka for proposing this.

afaik the route defaults are also available on the request. i fear that this is indeed mixing things. 

swagger-php has a tags attribute on the `@Operation` (base class of GET, PUT, ...). but with nelmio, we usually don't use the operations annotations, as that would mean to duplicate the url definition... 
but i think we should use the swagger tags concept for such manual attribution of end points to areas.
@GuilhemN do you have an idea how to do that? do we need a new annotation or could it be an optional attribute on a annotation we already have?

note that swagger-php also has the `@Tag` annotation. "A Swagger "Tag Object":  https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#tagObject" => this is to define meta information about a tag, NOT to define a tag on a method, unless i am completely mistaken. (the swagger-php documentation is rather limited)

GuilhemN:
> swagger-php has a tags attribute on the @Operation (base class of GET, PUT, ...). but with nelmio, we usually don't use the operations annotations, as that would mean to duplicate the url definition...

@dbu The bundle actually provides an [``@Operation`` annotation](https://github.com/nelmio/NelmioApiDocBundle/blob/master/Annotation/Operation.php) to avoid duplicating the url.

> but i think we should use the swagger tags concept for such manual attribution of end points to areas.
@GuilhemN do you have an idea how to do that? do we need a new annotation or could it be an optional attribute on a annotation we already have?

I don't think we need anything new, we already support all annotations that would be required by this implementation. The problem with it is that tags are shown in the documentation while they would be used for an internal purpose here.

Imo the best would be to have an ``@Area`` (or ``@Documentation``?) annotation, that would be detected by our routes filter.

> note that swagger-php also has the @Tag annotation. "A Swagger "Tag Object": https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#tagObject" => this is to define meta information about a tag, NOT to define a tag on a method, unless i am completely mistaken. (the swagger-php documentation is rather limited)

That's right but we still added support for ``@Tag`` on methods when using the bundle (see https://github.com/nelmio/NelmioApiDocBundle/blob/master/Describer/SwaggerPhpDescriber.php#L133-L138). Swagger-php does not support many annotations at the root of methods in fact.

dbu:
ah, nice. and the tags field of the Operation can be used in the area definition? imho that would be enough.

if the tag would be exposing too much, then an area annotation, or documentation annotation with areas attribute seems ok, yes.

wodka:
I would prefer the area annotation :) if you like then I'll try to change the pull request accordingly

GuilhemN:
> and the tags field of the Operation can be used in the area definition?

Yes, but it's not the only possibility to define tags and to be consistent,  I think we would have to support them all..

Using tags would bring more complexity than simplicity imo, there are too many possibilities to define tags on an operation and we can't predict them all before generating the docs.

dbu:
i would have expected tags to be the easy thing, but if you say its not.

anyways, if there is an actual use case for an additional area marker that is not mixed with swagger tags, i'd say go for it ;-)

wodka:
From how I understand it the areas that can be defined have no direct relation to swagger - in terms of that each area creates a completely separate swagger file.

If I add the area annotation should those annotations automatically be exposed or should they still require configuration like:
```yaml
nelmio_api_doc:
    areas:
        default:
            path_patterns: [ ^/api ]
            annotation: false # default
        admin:
            path_patterns: [ ^/api ]
            annotation: true
```

dbu:
i would have them always on. a tag should be able to specify more than one area i think, e.g. for a "privileged" area that has all basic calls plus additional privileged (secret) calls.

with a tag-like system, the areas could also specify several tags, but that gets quite complicated so i'd rather not do that, and it starts to overlap too much with the swagger concept of tags.

wodka:
ok, then basically something like:
```php
/** @Area("admin") */
function a() {}

/** @Area({"admin","default"}) */
function b() {}

/** */
function c() {}
```

default would include b() and c() and admin a() and b()

GuilhemN:
Like @dbu, I think `admin` should include `c()` (ie all areas should include routes without `@Area`)

dbu:
should there be a way to tell an area should not include unmarked routes? maybe `show_area_tag_undefined` that defaults to true but can be set to false? (not a fan of the name btw but can't think of a better one)

wodka:
perhaps `include_default` that is set to true per default

my example would be: 
```yaml
nelmio_api_doc:
    areas:
        default:
            path_patterns: [ ^/api ]
        admin:
            path_patterns: [ ^/api ]
            include_default: false
```

GuilhemN:
@wodka `include_default` looks good to me :+1:

GuilhemN:
Closing in favor of https://github.com/nelmio/NelmioApiDocBundle/pull/1460, thanks!

PR #1210, opened 2018-01-26
Nullable patterns no longer throw exceptions

martijnvankempen:
I found out that the Regex patterns are nullable which could trigger an exception. This fix will ignore patterns that are null.


dbu:
and thanks for the pull request!

martijnvankempen:
@dbu appendPattern is called three times in total. I prefer preventing duplicate code.

martijnvankempen:
@dbu I'll make sure to create some unit tests next month or so.

PR #1212, opened 2018-01-26
FormModelDescriber: set type to array with correct example for multiple choices form.

fredrocheron:
When a form containing a ChoiceType field with option multiple set to true is used with the `Nelmio\ApiDocBundle\Annotation\Model` annotation, the field type is set to `string` instead of `array` in the api documentation.

This PR set the type to `array` ~~and add an example value of `[1, 2, 3]`~~.

### DIFF
--- a/ModelDescriber/FormModelDescriber.php
+++ b/ModelDescriber/FormModelDescriber.php
@@ -103,7 +103,6 @@ final class FormModelDescriber implements ModelDescriberInterface, ModelRegistry
                 if ('choice' === $blockPrefix) {
                     if ($config->getOption('multiple')) {
                         $property->setType('array');
-                        $property->setExample('[1, 2, 3]');
                     } else {
                         $property->setType('string');
                     }
@@ -135,7 +134,6 @@ final class FormModelDescriber implements ModelDescriberInterface, ModelRegistry
                     if ($config->getOption('multiple')) {
                         $property->setFormat(sprintf('[%s id]', $entityClass));
                         $property->setType('array');
-                        $property->setExample('[1, 2, 3]');
                     } else {
                         $property->setType('string');
                         $property->setFormat(sprintf('%s id', $entityClass));
--- a/ModelDescriber/FormModelDescriber.php
+++ b/ModelDescriber/FormModelDescriber.php
@@ -101,7 +101,12 @@ final class FormModelDescriber implements ModelDescriberInterface, ModelRegistry
                 }
 
                 if ('choice' === $blockPrefix) {
-                    $property->setType('string');
+                    if ($config->getOption('multiple')) {
+                        $property->setType('array');
+                        $property->setExample('[1, 2, 3]');
+                    } else {
+                        $property->setType('string');
+                    }
                     if (($choices = $config->getOption('choices')) && is_array($choices) && count($choices)) {
                         $property->setEnum(array_values($choices));
                     }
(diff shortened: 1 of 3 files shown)


ModelDescriber/FormModelDescriber.php
```diff
@@ -101,7 +101,12 @@ private function parseForm(Schema $schema, FormInterface $form)
                 }
 
                 if ('choice' === $blockPrefix) {
-                    $property->setType('string');
+                    if ($config->getOption('multiple')) {
+                        $property->setType('array');
+                        $property->setExample('[1, 2, 3]');
```
GuilhemN: I don't think that's accurate, the field won't necessarilly look like this.

fredrocheron: I agree it won't always be accurate but I've just copied this line: https://github.com/fredrocheron/NelmioApiDocBundle/blob/cea91071cb29eb56cbeee2ca747d75218439b9fb/ModelDescriber/FormModelDescriber.php#L138
If you have an idea for a better example,  I'll use it. But I think it's important to set an example even if it's not perfect.

fredrocheron:
@GuilhemN I've looked for a test file to complete/modify before commiting but I was unable to find it. Could you point me in the right direction please ?

ModelDescriber/FormModelDescriber.php
GuilhemN: Imo https://github.com/fredrocheron/NelmioApiDocBundle/blob/cea91071cb29eb56cbeee2ca747d75218439b9fb/ModelDescriber/FormModelDescriber.php#L138 should be removed too. I think an example can only be set by the user; there it only highlights it's an array, while this is already highlighted by the array type.

GuilhemN:
Forms support is tested here: https://github.com/nelmio/NelmioApiDocBundle/blob/master/Tests/Functional/FunctionalTest.php#L206.
(Form types are in https://github.com/nelmio/NelmioApiDocBundle/tree/master/Tests/Functional/Form)

fredrocheron:
Thanks for your help. I've pushed a new commit with arrays examples removed and tests updated for multiple ChoiceType.

Tests/Functional/FunctionalTest.php
```diff
@@ -228,14 +228,18 @@ public function testFormSupport()
                     'type' => 'string',
                     'enum' => ['male', 'female'],
                 ],
+                'foz' => [
+                    'type' => 'array',
+                    'enum' => ['male', 'female'],
```
GuilhemN: This should go under `items` in fact

Tests/Functional/Form/DummyType.php
```diff
@@ -24,6 +24,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $builder->add('bar', TextType::class, ['required' => false]);
         $builder->add('foo', ChoiceType::class, ['choices' => ['male', 'female']]);
+        $builder->add('foz', ChoiceType::class, ['choices' => ['male', 'female'], 'multiple' => true]);
```
GuilhemN: Is there a way to detect the field's sub type, or does it only accept strings ?

fredrocheron: The add method only accepts strings or null.

fredrocheron:
New commit with requested changes and other things: 
- the FormModelDescriber now uses "items" for multiple ChoiceType forms
- the type of the property or items properties is now either "string" or "number" based on the values inside the "choices" option array.
- unused variable declaration has been removed

fredrocheron:
Great. Do I need to squash my commits ?

GuilhemN:
No need, i'm just waiting to have some time to merge this :)

GuilhemN:
Thank you @fredrocheron!

fredrocheron:
Thank you @GuilhemN! :)

PR #1225, opened 2018-02-13
Fix Api-Platform support

GuilhemN:


GuilhemN:
That's a BC break introduced in a patch so there's nothing we could have done about it. The method they added an argument on is public but it's a controller method so I guess they thought it was fine to change its signature, it should have been done in a minor at least though imo (see https://github.com/api-platform/core/commit/e3521932495eea27c4048c7a84e37ad80ae4f977).

ping @dunglas just so you know :)

dunglas:
Good catch. We could make this argument optional to revert the BC break (this class is `final`), but anyway you'll still need this patch to benefit from the bug fix.

dbu:
@dunglas thanks for the quick reaction. also, i guess this is really an edge case of BC, as when the controller is called by symfony it will automatically add the request... would there maybe be a way to extract a proper service we could call in NelmioApiDocBundle, instead of calling a controller?

dunglas:
This controller does almost nothing (https://github.com/api-platform/core/pull/1705/files#diff-7ee972d0500ace9d33ebcd1fec7c650dR48), maybe can you initialize the Documentation object directly in the factory?

dbu:
that would make sense, yes. 

could the documentation be exposed as service by api-platform and get injected all these configuration values for title and so on? that would be even nicer for us, and maybe cleaner on api-platform side as well, the action would just invoke the documentation class. probably should be a lazy service to not add performance overhead.

dunglas:
I'm not comfortable making a value object a service... A factory looks more appropriate.

Issue #1231, opened 2018-02-15
Definitions is found from entities, but not found from controller


petr-buchin:
In my entities I use definitions, as I was advised in https://github.com/nelmio/NelmioApiDocBundle/issues/1229#issuecomment-365691390 - by defining them in config.

This solution works well for entities, but somehow this does not work for controllers: the same ref to the same definition, that works in entity, does not work in controller.
My annotation that does not work:

```php
    /**
     * @Route("/example", methods={"GET"})
     *
     * @SWG\Response(
     *     response=400,
     *     description="Returned when parameter 'mediaSize' is absent or invalid.",
     *     @SWG\Schema(
     *         type="array",
     *         @SWG\Items(ref="#/definitions/Media")
     *     )
     * )
     */
```

And the error:
![Error](https://i.imgur.com/wCSubtJ.png)

P.S. `Media` definition is used for this response only as an example, since I have exactly the same ref to this definition in entity, and it works.

GuilhemN:
Indeed sorry, I forgot this issue due to a BC break in zircote/swagger-php... (see https://github.com/nelmio/NelmioApiDocBundle/issues/1085) I'll submit a patch asap.

petr-buchin:
@GuilhemN  thank you very much for the quick reply and for your work :)

GuilhemN:
Should be fixed by https://github.com/nelmio/NelmioApiDocBundle/pull/1232

Issue #1085, opened 2017-09-13
Could not resolve reference when $ref relative path files. "basePath" undefined. 


padlavoine:
I've actually a problem with $ref when displaying the swagger ui:

Resolver error at paths./private/api/v1/users.get.responses.200.schema.$ref
Could not resolve reference: ../../Entity/User.php#/definitions/User

/Controller/PrivateApi/UserController.php

```
/**
 * @Route("/v1/users")
 *
 */
class UserController extends BaseController
{    
     /**
     * List users.
     *
     * @Route("", methods={"GET"})
     * @SWG\Response(
     *     response=200,
     *     description="Returns the list of users",
     *     @SWG\Schema(ref="../../Entity/User.php#/definitions/User")
     * )
     * @SWG\Parameter(
     *     name="order",
     *     in="query",
     *     type="string",
     *     description="The field used to order users"
     * )
     * @SWG\Tag(name="Users")
     */
    public function getAction(Request $request, ParamFetcherInterface $paramFetcher)
    {
     .......
    } 
}
```
/Entity/User.php

```
/**
 *
 * @SWG\Definition(definition="User", type="object")
 *
 */
class User
{
    /**
     * @var string
     *
     * @ORM\Column()
     *
     * @Serializer\Expose
     * @Serializer\Groups({"UPDATE", "CREATE", "SHOW", "LIST"})
     *
     * @SWG\Property()
     *
     */
    private $name;
     
     .....
}

/**
 *  @SWG\Definition(
 *   definition="UserFull",
 *   type="object",
 *   allOf={
 *       @SWG\Schema(ref="#/definitions/User"),
 *       @SWG\Schema(
 *           required={"name"},
 *           @SWG\Property(property="uuid", format="int64", type="integer")
 *       )
 *   }
 * )
 */
```

config.yml:

```
# Nelmio API Doc (swagger)
nelmio_api_doc:
    models:
      use_jms: true
    routes:
        path_patterns: # an array of regexps
            - ^/api
            - ^/public
            - ^/private
    documentation:
        basePath: /
        info:
            title: ''
            description: '',
            version: 1.0
```
Thanks

GuilhemN:
`@SWG\Definition` is not supported by this bundle, that's why we provide `@Model`.

padlavoine:
Ok, so it's not possible to add custom model ? not an entity linked with database.
Sometimes, the controller get data from many entities and make new data model object with new datas.

   ```
[...]

PR #1232, opened 2018-02-15
Support references to config definitions

GuilhemN:
Fixes https://github.com/nelmio/NelmioApiDocBundle/issues/1231 and https://github.com/nelmio/NelmioApiDocBundle/issues/1085#issuecomment-354111322

### DIFF
--- a/Describer/SwaggerPhpDescriber.php
+++ b/Describer/SwaggerPhpDescriber.php
@@ -24,30 +24,31 @@ use Swagger\Annotations as SWG;
 use Swagger\Context;
 use Symfony\Component\Routing\RouteCollection;
 
-final class SwaggerPhpDescriber extends ExternalDocDescriber implements ModelRegistryAwareInterface
+final class SwaggerPhpDescriber implements ModelRegistryAwareInterface
 {
     use ModelRegistryAwareTrait;
 
     private $routeCollection;
-
     private $controllerReflector;
-
     private $annotationReader;
+    private $overwrite;
 
     public function __construct(RouteCollection $routeCollection, ControllerReflector $controllerReflector, Reader $annotationReader, bool $overwrite = false)
     {
         $this->routeCollection = $routeCollection;
         $this->controllerReflector = $controllerReflector;
         $this->annotationReader = $annotationReader;
+        $this->overwrite = $overwrite;
+    }
 
-        parent::__construct(function () {
-            $analysis = $this->getAnnotations();
+    public function describe(Swagger $api)
+    {
+        $analysis = $this->getAnnotations($api);
 
-            $analysis->process($this->getProcessors());
-            $analysis->validate();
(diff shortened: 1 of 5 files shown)


Describer/SwaggerPhpDescriber.php
```diff
@@ -60,9 +61,30 @@ private function getProcessors(): array
         return array_merge($processors, Analysis::processors());
     }
 
-    private function getAnnotations(): Analysis
+    private function getAnnotations(Swagger $api): Analysis
     {
         $analysis = new Analysis();
+        $analysis->addAnnotation(new class($api) extends SWG\Swagger {
```
dbu: why do you create this as an anonymous class? would a named class that can be unit tested not make more sense?

Describer/SwaggerPhpDescriber.php
```diff
@@ -60,9 +61,30 @@ private function getProcessors(): array
...
+            private $api;
+
+            public function __construct($api)
```
dbu: why no type declaration for $api?

Describer/SwaggerPhpDescriber.php
```diff
@@ -60,9 +61,30 @@ private function getProcessors(): array
...
+                $this->api = $api;
+                parent::__construct([]);
+            }
+
+            /**
+             * Support definitions from the config and reference to models.
+             */
+            public function ref($ref)
+            {
+                if (0 === strpos($ref, '#/definitions/') && $this->api->getDefinitions()->has(substr($ref, 14))) {
```
dbu: what if $this->api is null? (as i think it is, looking at how the class is instantiated below)

Describer/SwaggerPhpDescriber.php
GuilhemN: There is not much to test, the fact that no exception is thrown during the documentation build already proves that it works as expected (I added a reference to a config definition. 

I used an anonymous class because this class is only useful here and does almost nothing but I can put it in its own file if you truly prefer.

dbu: fair enough, just wanted to challenge the design. lets keep it anonymous.

Describer/SwaggerPhpDescriber.php
GuilhemN: added :)

Describer/SwaggerPhpDescriber.php
GuilhemN: It's not, ``parent::_construct([])`` calls ``SWG\Swagger`` constructor which doesn't have an ``$api`` property.

dbu: ah, the `null` i see is for addAnnotation and the instantiation is above before the anonymous declaration. i am not used to anonymous classes. all good then!

PR #1242, opened 2018-02-26
Remove beta tag from upgrade note

Soullivaneuh:


GuilhemN:
Good catch, thank you @Soullivaneuh :)

PR #1243, opened 2018-02-27
Fix #1239: FosRestDescriber properly configures Parameter#pattern property

### LINKED ISSUE #1239: FOSRestBundle's `AbstractScalarParam#requirements` should be used for `SWG\Parameter#pattern`
Hello,

I think that by accident [`Parameter#format`](https://github.com/nelmio/NelmioApiDocBundle/blob/v3.1.0/RouteDescriber/FosRestDescriber.php#L57) is set instead of the `pattern` property. I might be wrong but to me, _format_ should be used for different purposes, and FOSRestBundle's `QueryParam#requirements` and `RequestParam#requirements` should be used to configure `Parameter::setPattern()` property.

Right now input validation is completely ignored. With suggestion above it works fine. However, I can't still see in the UI regex pattern for allowed input.

For example, the following FOSRestBundle `QueryParam` annotation would produce input field like shown in the screenshot, but allowed regex is not mentioned, therefore the clients can't really know what is allowed.

```
@QueryParam(name="itemsPerPage", default="20", requirements="\d{1,2}", description="Items per page. Default is `20`.")
```

missing regex pattern

Thanks!

ManInTheBox:
Minor bugfix that configures `Parameter#pattern` property instead of `Parameter#format`.
Tests provided that prove this implementation works as expected.

### DIFF
--- a/RouteDescriber/FosRestDescriber.php
+++ b/RouteDescriber/FosRestDescriber.php
@@ -54,7 +54,18 @@ final class FosRestDescriber implements RouteDescriberInterface
 
                 $normalizedRequirements = $this->normalizeRequirements($annotation->requirements);
                 if (null !== $normalizedRequirements) {
-                    $parameter->setPattern($normalizedRequirements);
+                    if ($normalizedRequirements instanceof Constraint) {
+                        if ($normalizedRequirements instanceof Regex) {
+                            $format = $normalizedRequirements->getHtmlPattern();
+                        } else {
+                            $reflectionClass = new \ReflectionClass($normalizedRequirements);
+                            $format = $reflectionClass->getShortName();
+                        }
+
+                        $parameter->setFormat($format);
+                    } else {
+                        $parameter->setPattern($normalizedRequirements);
+                    }
                 }
             }
         }
@@ -71,13 +82,7 @@ final class FosRestDescriber implements RouteDescriberInterface
         }
         // if custom constraint
         if ($requirements instanceof Constraint) {
-            if ($requirements instanceof Regex) {
-                return $requirements->getHtmlPattern();
-            }
-
-            $reflectionClass = new \ReflectionClass($requirements);
-
-            return $reflectionClass->getShortName();
+            return $requirements;
         }
     }
 }
--- a/RouteDescriber/FosRestDescriber.php
+++ b/RouteDescriber/FosRestDescriber.php
@@ -54,7 +54,7 @@ final class FosRestDescriber implements RouteDescriberInterface
 
                 $normalizedRequirements = $this->normalizeRequirements($annotation->requirements);
                 if (null !== $normalizedRequirements) {
-                    $parameter->setFormat($normalizedRequirements);
+                    $parameter->setPattern($normalizedRequirements);
                 }
             }
         }
--- a/RouteDescriber/FosRestDescriber.php
+++ b/RouteDescriber/FosRestDescriber.php
@@ -52,37 +52,44 @@ final class FosRestDescriber implements RouteDescriberInterface
                     $parameter->setDescription($annotation->description);
                 }
 
-                $normalizedRequirements = $this->normalizeRequirements($annotation->requirements);
-                if (null !== $normalizedRequirements) {
-                    if ($normalizedRequirements instanceof Constraint) {
-                        if ($normalizedRequirements instanceof Regex) {
-                            $format = $normalizedRequirements->getHtmlPattern();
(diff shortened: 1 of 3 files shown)


RouteDescriber/FosRestDescriber.php
```diff
@@ -54,7 +54,7 @@ public function describe(Swagger $api, Route $route, \ReflectionMethod $reflecti
 
                 $normalizedRequirements = $this->normalizeRequirements($annotation->requirements);
                 if (null !== $normalizedRequirements) {
-                    $parameter->setFormat($normalizedRequirements);
+                    $parameter->setPattern($normalizedRequirements);
```
GuilhemN: We should manage differently requirements when they are a regex or a constraint class imo.
Could you inline the normalizeRequirements() method here and if it's a constraint name, keep using the format field please ?

ManInTheBox: Thanks, will keep you updated.

RouteDescriber/FosRestDescriber.php
```diff
@@ -54,7 +54,18 @@ public function describe(Swagger $api, Route $route, \ReflectionMethod $reflecti
 
                 $normalizedRequirements = $this->normalizeRequirements($annotation->requirements);
                 if (null !== $normalizedRequirements) {
-                    $parameter->setFormat($normalizedRequirements);
+                    if ($normalizedRequirements instanceof Constraint) {
+                        if ($normalizedRequirements instanceof Regex) {
```
GuilhemN: Shouldn't this still be put in ``pattern``?

I think we should only use ``format`` for random constraints.

ManInTheBox: Makes sense. Perhaps I didn't understand the requirements properly.
So we want `Regex` into `pattern` and only other custom classes (the else block) to go into the `format`, right?

GuilhemN: That's right :)

ManInTheBox:
Hey, @GuilhemN I refactored the code a bit and (I believe) fulfilled all the requirements. Could you please review this one more time? Thanks!

RouteDescriber/FosRestDescriber.php
```diff
@@ -52,32 +52,44 @@ public function describe(Swagger $api, Route $route, \ReflectionMethod $reflecti
...
-        if ($requirements instanceof Constraint) {
-            if ($requirements instanceof Regex) {
-                return $requirements->getHtmlPattern();
-            }
 
+        if ($requirements instanceof Regex) {
+            return $requirements->getHtmlPattern();
+        }
+
+        return null;
```
ManInTheBox: I explicitly return `null` and not rely on PHP's default return value which is also `null`. This is more a preference. Feel free to object if you don't like it ;)

Tests/Functional/FunctionalTest.php
```diff
@@ -124,6 +124,22 @@ public function testFOSRestAction()
...
+        $this->assertNull($fooParameter->getFormat());
+
+        $barParameter = $parameters->get('bar', 'formData');
+        $this->assertNotNull($barParameter->getPattern());
+        $this->assertEquals('\d+', $barParameter->getPattern());
+        $this->assertNull($barParameter->getFormat());
+
+        $bazParameter = $parameters->get('baz', 'formData');
+        $this->assertNotNull($bazParameter->getFormat());
+        $this->assertEquals('IsTrue', $bazParameter->getFormat());
```
ManInTheBox: I believe this is what we want, right? If they put custom `Constraint` which also **is not** `Regex` constraint, then we want to use only its class short name.

RouteDescriber/FosRestDescriber.php
```diff
@@ -52,32 +52,44 @@ public function describe(Swagger $api, Route $route, \ReflectionMethod $reflecti
                     $parameter->setDescription($annotation->description);
                 }
 
-                $normalizedRequirements = $this->normalizeRequirements($annotation->requirements);
-                if (null !== $normalizedRequirements) {
-                    $parameter->setFormat($normalizedRequirements);
+                $pattern = $this->getPattern($annotation->requirements);
+                if (null !== $pattern) {
```
ManInTheBox: I also don't support assignments in conditional checks, so I didn't want to assign `$pattern` here but instead on the line above, before the conditional check. Feel free to object if this doesn't fit in this project.

ManInTheBox:
You're welcome! ;)

PR #1257, opened 2018-03-13
Allow to filter routes by host (#1121)

### LINKED ISSUE #1121: Host/hosts parameter
I propose to add the host/hosts parameter to config. 

`
    private function match(Route $route): bool
    {
        if(  !in_array( $route->getHost(), $this->hosts ) ){
            return false;
        }
        foreach ($this->pathPatterns as $pathPattern) {
            if (  preg_match('{'.$pathPattern.'}', $route->getPath())) {
                return true;
            }
        }

        return false;
    }
`

phansys:
|Q            |A     |
|---          |---   |
|Branch       |master|
|Bug fix?     |no    |
|New feature? |yes   |
|BC breaks?   |no    |
|Deprecations?|no    |
|Tests pass?  |yes   |
|Fixed tickets|#1121 |
|License      |MIT   |
|Doc PR       |n/a   |

**TODO:**
- [x] Fix tests;
- [x] Update docs.

Closes #1121.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ CHANGELOG
 * Add a documentation form extension. Use the ``documentation`` option to define how a form field is documented.
 * Allow references to config definitions in controllers.
 
+Config
+* `nelmio_api_doc.areas` added support to filter by host patterns.
+
+  ```yml
+  nelmio_api_doc:
+      routes: [ host_patterns: [ ^api\. ] ]
+  ```
+
+* Added dependency for "symfony/options-resolver:^3.4.4|^4.0"
+
 3.1.0 (2017-01-28)
 ------------------
 
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -48,10 +48,10 @@ final class Configuration implements ConfigurationInterface
                 ->end()
                 ->arrayNode('areas')
                     ->info('Filter the routes that are documented')
-                    ->defaultValue(['default' => ['path_patterns' => []]])
+                    ->defaultValue(['default' => ['path_patterns' => [], 'host_patterns' => []]])
                     ->beforeNormalization()
                         ->ifTrue(function ($v) {
-                            return empty($v) or isset($v['path_patterns']);
+                            return 0 === count($v) || isset($v['path_patterns']) || isset($v['host_patterns']);
                         })
                         ->then(function ($v) {
                             return ['default' => $v];
@@ -68,7 +68,13 @@ final class Configuration implements ConfigurationInterface
                         ->addDefaultsIfNotSet()
                         ->children()
                             ->arrayNode('path_patterns')
+                                ->defaultValue([])
                                 ->example(['^/api', '^/api(?!/admin)'])
+                                ->prototype('scalar')->end()
+                            ->end()
+                            ->arrayNode('host_patterns')
+                                ->defaultValue([])
+                                ->example(['^api\.'])
                                 ->prototype('scalar')->end()
                             ->end()
                         ->end()
--- a/DependencyInjection/NelmioApiDocExtension.php
+++ b/DependencyInjection/NelmioApiDocExtension.php
@@ -66,7 +66,7 @@ final class NelmioApiDocExtension extends Extension implements PrependExtensionI
                     new TaggedIteratorArgument('nelmio_api_doc.model_describer'),
                 ]);
 
-            if (0 === count($areaConfig['path_patterns'])) {
+            if (0 === count($areaConfig['path_patterns']) && 0 === count($areaConfig['host_patterns'])) {
                 $container->setDefinition(sprintf('nelmio_api_doc.routes.%s', $area), $routesDefinition)
                     ->setPublic(false);
             } else {
@@ -74,7 +74,7 @@ final class NelmioApiDocExtension extends Extension implements PrependExtensionI
                     ->setPublic(false)
                     ->setFactory([
                         (new Definition(FilteredRouteCollectionBuilder::class))
-                            ->addArgument($areaConfig['path_patterns']),
+                            ->addArgument($areaConfig),
                         'filter',
                     ])
                     ->addArgument($routesDefinition);
--- a/Routing/FilteredRouteCollectionBuilder.php
+++ b/Routing/FilteredRouteCollectionBuilder.php
@@ -11,23 +11,41 @@
 
 namespace Nelmio\ApiDocBundle\Routing;
 
+use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\Routing\Route;
 use Symfony\Component\Routing\RouteCollection;
 
 final class FilteredRouteCollectionBuilder
 {
-    private $pathPatterns;
+    private $options;
 
-    public function __construct(array $pathPatterns = [])
+    public function __construct(array $options = [])
     {
-        $this->pathPatterns = $pathPatterns;
+        $resolver = new OptionsResolver();
+        $resolver
+            ->setDefaults([
+                'path_patterns' => [],
+                'host_patterns' => [],
+            ])
+            ->setAllowedTypes('path_patterns', 'string[]')
+            ->setAllowedTypes('host_patterns', 'string[]')
+        ;
+
+        if (array_key_exists(0, $options)) {
+            @trigger_error(sprintf('Passing an indexed array with a collection of path patterns as argument 1 for `%s()` is deprecated since 3.2.0, expected structure is an array containing parameterized options.', __METHOD__), E_USER_DEPRECATED);
+
+            $normalizedOptions = ['path_patterns' => $options];
+            $options = $normalizedOptions;
+        }
+
+        $this->options = $resolver->resolve($options);
     }
 
     public function filter(RouteCollection $routes): RouteCollection
     {
         $filteredRoutes = new RouteCollection();
         foreach ($routes->all() as $name => $route) {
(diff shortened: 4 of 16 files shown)


phansys:
I would appreciate any help on how to set default values separately for `path_patterns` and `host_patterns` at `Configuration::getConfigTreeBuilder()`. Thank you in advance.

Routing/FilteredRouteCollectionBuilder.php
```diff
@@ -37,12 +40,32 @@ public function filter(RouteCollection $routes): RouteCollection
...
+        }
+
+        if (!$matchingPath) {
+            return false;
+        }
+
+        $matchingHost = empty($this->hostPatterns);
+        foreach ($this->hostPatterns as $hostPattern) {
+            if (preg_match('{'.$hostPattern.'}', $route->getHost())) {
+                $matchingHost = true;
```
GuilhemN: We can return true here directly

phansys: That's true, but I decided to respect the same logic than checking paths, in order to allow this method to accept other filters in future too.

Routing/FilteredRouteCollectionBuilder.php
```diff
@@ -18,9 +18,12 @@
 {
     private $pathPatterns;
 
-    public function __construct(array $pathPatterns = [])
+    private $hostPatterns;
+
+    public function __construct(array $pathPatterns = [], array $hostPatterns = [])
```
dbu: its kindof the last moment for this, but i think we should use an options array with option resolver here, to allow for future extensions. then you can pass something like

```
[
    'path_patterns' => [...],
    'host_patterns' => [...],
]
```

the options make it easier to omit parameters that are not needed.

then again this is all configured from symfony config, so not as important as it would be otherwise.

Routing/FilteredRouteCollectionBuilder.php
dbu: i think that point makes sense. but maybe this could be rewritten into one method per type (host, path) and then the main match would be `return $this->matchPath() && $this->matchHost();` . the specific match methods can then use early returns

Routing/FilteredRouteCollectionBuilder.php
phansys: I agree that could be nice to have just one configuration argument, but I think that will break the BC. If there is no issue with that concern, I can work on that way.

Routing/FilteredRouteCollectionBuilder.php
phansys: Splitted into 2 separated methods.

Routing/FilteredRouteCollectionBuilder.php
dbu: for bc, you could check if the first key is 0 and if it is, do $options = ['path' => $options], then its BC

phansys: Done.

CHANGELOG.md
```diff
@@ -7,6 +7,30 @@ CHANGELOG
...
 
+Config
+* `nelmio_api_doc.areas` added support to filter by host patterns.
+
+  ```yml
+  nelmio_api_doc:
+      routes: [ host_patterns: [ ^api\. ] ]
+  ```
+
+FilteredRouteCollectionBuilder
```
dbu: i would opt to not mention this here, as the builder is not expected to be called directly. so the BC is for the edge case of somebody doing something custom. those people will see the deprecation note. having it here is too verbose imho.

DependencyInjection/Configuration.php
```diff
@@ -48,10 +48,10 @@ public function getConfigTreeBuilder()
                 ->end()
                 ->arrayNode('areas')
                     ->info('Filter the routes that are documented')
-                    ->defaultValue(['default' => ['path_patterns' => []]])
+                    ->defaultValue(['default' => ['path_patterns' => [], 'host_patterns' => []]])
                     ->beforeNormalization()
                         ->ifTrue(function ($v) {
-                            return empty($v) or isset($v['path_patterns']);
+                            return empty($v) || isset($v['path_patterns']) || isset($v['host_patterns']);
```
dbu: don't the keys always exist? could we instead `count`?

DependencyInjection/NelmioApiDocExtension.php
```diff
@@ -66,15 +66,15 @@ public function load(array $configs, ContainerBuilder $container)
                     new TaggedIteratorArgument('nelmio_api_doc.model_describer'),
                 ]);
 
-            if (0 === count($areaConfig['path_patterns'])) {
+            if (empty($areaConfig['path_patterns']) && empty($areaConfig['host_patterns'])) {
```
dbu: i would keep the count. empty is a much more vague statement. performance is not cricital, as this is only executed during container building.

Routing/FilteredRouteCollectionBuilder.php
```diff
@@ -11,38 +11,67 @@
...
+            ->setDefaults([
+                'path_patterns' => [],
+                'host_patterns' => [],
+            ])
+            ->setAllowedTypes('path_patterns', 'string[]')
+            ->setAllowedTypes('host_patterns', 'string[]')
+        ;
+
+        if (isset($options[0])) {
+            @trigger_error(sprintf('Passing an indexed array with a collection of path patterns as argument 1 for `%s()` is deprecated since 3.2.0, expected structure is an array containing parameterized options.', __METHOD__), E_USER_DEPRECATED);
```
dbu: excellent! its perfect when the notice explains what is wrong and what should be done.

Routing/FilteredRouteCollectionBuilder.php
```diff
@@ -11,38 +11,67 @@
...
     {
-        foreach ($this->pathPatterns as $pathPattern) {
+        foreach ($this->options['path_patterns'] as $pathPattern) {
             if (preg_match('{'.$pathPattern.'}', $route->getPath())) {
                 return true;
             }
         }
 
-        return false;
+        return empty($this->options['path_patterns']);
```
dbu: i'd prefer count here as well, and below for matchHost

Tests/Routing/FilteredRouteCollectionBuilderTest.php
```diff
@@ -22,22 +22,76 @@
 class FilteredRouteCollectionBuilderTest extends TestCase
```
dbu: can you please add a test with a host_pattern that makes sure host is also taken into account? the set of routes should be one that only matches path, one only host, and one both and the check should make sure that only the one matching both is accepted.

phansys: Done.

DependencyInjection/Configuration.php
phansys: If the keys aren't configured, they don't exist before normalization.

dbu: ah, right. sorry for the confusion!

Tests/Routing/FilteredRouteCollectionBuilderTest.php
```diff
@@ -22,22 +22,137 @@
...
+
+        $routes = new RouteCollection();
+        foreach ($this->getRoutes() as $name => $route) {
+            $routes->add($name, $route);
+        }
+
+        $routeBuilder = new FilteredRouteCollectionBuilder($options);
+        $filteredRoutes = $routeBuilder->filter($routes);
+
+        $this->assertCount(5, $filteredRoutes);
```
dbu: why is this 5? looking at getRoutes, i would have thought we end up with 4 hits: r1, r2, r3, r9. am i missing something?

phansys: You're right @dbu, thanks for pointing. It seems that 'host_patterns' => [''] is allowing "r7" too (`preg_match('{}', 'www.example.com')`). I don't know if we should forbid an empty string here, or treat it in a different way, something like
```php
if (('' === $hostPattern && $hostPattern === $route->getHost()) || preg_match('{'.$hostPattern.'}', $route->getHost())) {
    return true;
}
```

dbu: ah, right, its regular expressions. i think configuring `''` as pattern is simply a mistake. if you don't specify any expression, all paths|hosts match. if you explicitly want an empty host, the expression should be `^$`.

phansys: I've updated the test in order to use `^$`. Thank you.

dbu:
not sure whats up with the lowest version build, that seems new to me and did not fail in master. but looks like it affects a lot of things. is something filtering out everything because of a host_pattern in the functional tests?

phansys:
I've added the "HTTP_HOST" setting at [client](https://github.com/nelmio/NelmioApiDocBundle/pull/1257/commits/db9ce7bddaff6dba14eccae58f135858edad547c#diff-9b347a85c2cffeea1be718493ee039cfR27), but the tests are still [failing](https://travis-ci.org/nelmio/NelmioApiDocBundle/jobs/353969570).

phansys:
Tests fixed by setting `Route(host="...")` at test controllers.

GuilhemN:
@dbu https://github.com/nelmio/NelmioApiDocBundle/pull/1260 fixes Api-Platform support, they changed the class of one of their services.
Thanks for your work @phansys! Will review ASAP :)

dbu:
the fix of the api platform BC break is merged. can you please rebase this branch on master?

phansys:
Done.

PR #1260, opened 2018-03-17
Fix Api-Platform support

GuilhemN:


Describer/ApiPlatformDescriber.php
```diff
@@ -12,13 +12,18 @@
...
 use ApiPlatform\Core\Swagger\Serializer\DocumentationNormalizer;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 
 final class ApiPlatformDescriber extends ExternalDocDescriber
 {
-    public function __construct(Documentation $documentation, DocumentationNormalizer $normalizer, UrlGeneratorInterface $urlGenerator)
+    public function __construct(Documentation $documentation, $normalizer, UrlGeneratorInterface $urlGenerator)
     {
+        if (!$normalizer instanceof ApiGatewayNormalizer && !$normalizer instanceof DocumentationNormalizer) {
+            throw new \LogicException(sprintf('Argument $normalizer of %s must be an instance of %s or %s. %s provided.', __METHOD__, ApiGatewayNormalizer::class, DocumentationNormalizer::class, get_class($normalizer)));
```
phansys: Shouldn't this be `\InvalidArgumentException` instead?

dbu: yeah, probably. did not notice it when i merged. nobody should be catching this exception so its not super important, but i think InvalidArgumentException is more common for this.

Issue #1254, opened 2018-03-12
Symfony form in GET-request


Nadrshin:
Is it possible to use symfony form in GET-reqeust?

dbu:
would that mean a GET request with a body? this is very unusual, and some proxies might discard the body. afaik elasticsearch used to do it, but they switched to POST meanwhile because GET with body was creating problems with tools not expecting it.

Nadrshin:
Without body.

Let's say I have a form FilterType with two fields ("name" and "surname") and block prefix "filter". In NelmioApiDocBundle v2 I used this form in "input":

`input = "ApiBundle\Form\Type\FilterType"`

In the sandbox I made the following request:

`/api/v1/users?filter[name]=Michael&filter[surname]=Phelps`

dbu:
ah, that sounds useful. maybe @GuilhemN can help here?

GuilhemN:
That's not supported in 3.x unfortunately, mostly due to the way parameters are implemented in swagger. That could be added but the biggest challenge is to find a nice syntax.

Maybe we need to rethink `@Model` to cover all cases more nicely. What do you think about extending all zircote/swagger-php annotations in the bundle and add a `model` field to the annotations needing it instead of having `@Model` ? That would allow us to provide features more integrated with zircote/swagger-php annotations...

dbu:
could it be possible to add a `model` field to zircote/swagger-php instead? that might be more straightforward and less confusing. but is only possible if the model field would make sense for zircote itself as well - but from how i understand things, it would make sense.

GuilhemN:
Thinking about it again we could also use the ``ref`` field:
```php
/**
 * @Response(response=200,
 *      @Schema(ref=@Model(type=""))
 * )
 */
```
when it is not clear what ``@Model`` should be used for. That makes things much clearer imo.

> could it be possible to add a model field to zircote/swagger-php instead? that might be more straightforward and less confusing. but is only possible if the model field would make sense for zircote itself as well - but from how i understand things, it would make sense.

Well they use ``ref`` a lot, I don't think this would be accepted. Plus we would have no way to pass options as well (serialization groups for instance).

jmalinens:
We have the same problem. We use Symfony Forms in GET request for filtering content.
That would be nice feature!

torinaki:
I have faced with the same issue during describing [JSON:API](http://jsonapi.org/) endpoints. 
[...]

PR #1261, opened 2018-03-17
Support @Model through `ref`

GuilhemN:
Implements https://github.com/nelmio/NelmioApiDocBundle/issues/1254#issuecomment-373918345.

>Thinking about it again we could also use the ``ref`` field:
> ```php
> /**
>  * @Response(response=200,
>  *      @Schema(ref=@Model(type=""))
>  * )
>  */
> ```
> when it is not clear what ``@Model`` should be used for. That makes things much clearer imo.
> 
> > could it be possible to add a model field to zircote/swagger-php instead? that might be more straightforward and less confusing. but is only possible if the model field would make sense for zircote itself as well - but from how i understand things, it would make sense.
> 
> Well they use ``ref`` a lot, I don't think this would be accepted. Plus we would have no way to pass options as well (serialization groups for instance).

### DIFF
--- a/Resources/doc/faq.rst
+++ b/Resources/doc/faq.rst
@@ -1,56 +1,6 @@
 Frequently Asked Questions (FAQ)
 ================================
 
-* Q: I use ``@Model`` to document an operation and the bundle understands I want an array of models while I only want one.
-
-  A: You most likely nested ``@Model`` in a ``@Schema`` annotation. The ``@Model`` annotation acts like a ``@Schema`` annotation, so
-     when nested, the bundle considers that you're documenting an array of models.
-
-     For instance, the following example::
-
-         /**
-          * @SWG\Response(
-          *   response="200",
-          *   description="Success",
-          *   @SWG\Schema(@Model(type=User::class))
-          * )
-          */
-         public function getUserAction()
-         {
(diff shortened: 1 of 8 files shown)


Resources/doc/faq.rst
```diff
@@ -1,56 +1,6 @@
 Frequently Asked Questions (FAQ)
 ================================
 
-* Q: I use ``@Model`` to document an operation and the bundle understands I want an array of models while I only want one.
```
dbu: would it not make sense to keep the FAQ entry and explain the ref= vs no ref= syntax?

GuilhemN: In the last commit I deprecated nesting `@Model` with `@Schema` thus removing the problem / the need of this Q/A imo. The no ref= syntax is now just a shortcut that's described in a tip box. 
What do you think?

GuilhemN:
> i think #1263 is the same as this right? but yes, looks like a good improvement that is more explicit and avoids confusion.

Well in fact #1263 is based on this PR (I prefered doing my changes in two steps to simplify reviewing)

GuilhemN:
Thank you for your reviews @dbu 😃

PR #1263, opened 2018-03-18
Support ``@Model`` in in-object annotations

GuilhemN:
That's quite weird to not be able to use ``@Model`` in objects, this PR fixes this limitation.

### DIFF
--- a/ModelDescriber/Annotations/AnnotationsReader.php
+++ b/ModelDescriber/Annotations/AnnotationsReader.php
@@ -13,6 +13,7 @@ namespace Nelmio\ApiDocBundle\ModelDescriber\Annotations;
 
 use Doctrine\Common\Annotations\Reader;
 use EXSyst\Component\Swagger\Schema;
+use Nelmio\ApiDocBundle\Model\ModelRegistry;
 
 /**
  * @internal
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,27 @@ CHANGELOG
...
 * Add a documentation form extension. Use the ``documentation`` option to define how a form field is documented.
 * Allow references to config definitions in controllers.
+* Using `@Model` implicitely in `@SWG\Schema`, `@SWG\Items` and `@SWG\Property` is deprecated. Use `ref=@Model()` instead.
+
+  Before:
+  ```php
+  /**
+   * This was considered as an array of models.
+   *
+   * @SWG\Property(@Model(model=FooClass::class))
```
kunicmarko20: This does not work, model does not exist, maybe type? but then I get ```User Notice: Unexpected @Nelmio\ApiDocBundle\Annotation\Model(), expected to be inside @SWG\Parameter, @SWG\Response```

GuilhemN: > This does not work, model does not exist, maybe type?

That's a typo indeed 😄 

> but then I get User Notice: Unexpected @Nelmio\ApiDocBundle\Annotation\Model(), expected to be inside @SWG\Parameter, @SWG\Response

In fact using this annotation in an object like described above will be allowed by this PR, what's deprecated is when ``@Model`` was used in ``@SWG\Parameter`` itself nested with ``@SWG\Response``, but that's easier to describer without ``@SWG\Response`` imo:
```php
/**
 * @SWG\Response(response=200, 
 *     @SWG\Schema(type="object", @SWG\Property(@Model(...)))
 * )
 */

kunicmarko20: I thought because it says `before` that it already works 😄 thank you for the explanation.

GuilhemN:
@dbu shall we release 3.2 now ? :)

dbu:
sounds like a good moment to do it, yes. go for it!

PR #1264, opened 2018-03-19
Updated SwaggerDocblockConvertCommand command in UPGRADE-3.0.md

adeelnawaz:
Changes:  
- Allowed null input in `escapeQuotes(string $str)` - description of an API can be null
- Added command option `views` to input comma separated list of views - the API can have view(s) other than `default`

---

HowToTest:  
- The command should work for APIs with no description
  (Believe me! there are people who create APIs without description; but I can remove this change if you don't like it)
- The command should be able to take a comma separated list of views and convert docblocks for APIs of all those views


GuilhemN:
Thank you @adeelnawaz!

PR #1273, opened 2018-03-22
command missing precission

decima:
I had to do that in order to not loose code.


gesof:
So that's why some code got removed

PR #1285, opened 2018-04-07
Fix #1283: Invalid swagger.json generation - definition for entities …

### LINKED ISSUE #1283: Invalid swagger.json generation - definition for entities with both Assert\NotBlank|Assert\NotNull and Assert\Length
Having an entity with both `@Assert\NotBlank` and `@Assert\Length` generates `swagger.json` from array like `[0 => "name", 2 => "accountId"]`, so we get:
```json
"definitions":{"User":{"required":{"0":"name","2":"accountId"}
```
This is due to the way `array_unique()` function works - it preserves keys (also the numeric ones).

FIX
https://github.com/nelmio/NelmioApiDocBundle/blob/ff590b013c59a37cf624c38c2ae4e1799d518209/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php#L100 could be replaced with (additional `array_values()` function):
```php
$this->schema->setRequired(array_values(array_unique($existingRequiredFields)));
```

jlekowski:
…with Assert\NotBlank and Assert\Length

Having an entity with both `@Assert\NotBlank` and `@Assert\Length` may result in generating `swagger.json` from array like `[0 => "name", 2 => "accountId"]`, so we get:
```json
"definitions":{"User":{"required":{"0":"name","2":"accountId"}
```
This is due to the way `array_unique()` function works - it preserves keys (also the numeric ones). 

Example:
```php
<?php

$array = array_unique([1,1]);
$array[] = 2;
var_dump($array);
/*
array(2) {
  [0]=>
  int(1)
  [2]=>
  int(2)
}
*/
```

### DIFF
--- a/Tests/ModelDescriber/Annotations/SymfonyConstraintAnnotationReaderTest.php
+++ b/Tests/ModelDescriber/Annotations/SymfonyConstraintAnnotationReaderTest.php
@@ -1,0 +1,48 @@
+<?php
+
+/*
+ * This file is part of the NelmioApiDocBundle package.
+ *
+ * (c) Nelmio
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Nelmio\ApiDocBundle\Tests\ModelDescriber\Annotations;
+
+use Doctrine\Common\Annotations\AnnotationReader;
+use EXSyst\Component\Swagger\Schema;
+use Nelmio\ApiDocBundle\ModelDescriber\Annotations\SymfonyConstraintAnnotationReader;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Validator\Constraints as Assert;
+
+class SymfonyConstraintAnnotationReaderTest extends TestCase
+{
+    public function testUpdatePropertyFix1283()
+    {
+        $entity = new class() {
+            /**
+             * @Assert\NotBlank()
+             * @Assert\Length(min = 1)
+             */
+            private $property1;
+            /**
+             * @Assert\NotBlank()
+             */
+            private $property2;
+        };
+        $reflectionProperties = (new \ReflectionClass($entity))->getProperties();
+        $property = new Schema();
+        $schema = new Schema();
+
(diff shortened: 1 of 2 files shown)


Tests/ModelDescriber/Annotations/SymfonyConstraintAnnotationReaderTest.php
```diff
@@ -0,0 +1,71 @@
...
+    private $reader;
+
+    private $symfonyConstraintAnnotationReader;
+
+    public function testUpdatePropertyFix1283()
+    {
+        $reflectionProperty1 = $this->createMock(\ReflectionProperty::class);
+        $reflectionProperty2 = $this->createMock(\ReflectionProperty::class);
+        $property = $this->createMock(Schema::class);
+        $schema = $this->createMock(Schema::class);
```
GuilhemN: There is no need to mock Schema, you can just check the final value of `required` :)

Tests/ModelDescriber/Annotations/SymfonyConstraintAnnotationReaderTest.php
```diff
@@ -0,0 +1,71 @@
...
+class SymfonyConstraintAnnotationReaderTest extends TestCase
+{
+    private $reader;
+
+    private $symfonyConstraintAnnotationReader;
+
+    public function testUpdatePropertyFix1283()
+    {
+        $reflectionProperty1 = $this->createMock(\ReflectionProperty::class);
+        $reflectionProperty2 = $this->createMock(\ReflectionProperty::class);
```
GuilhemN: I'd rather avoid mocking as much as possible, why not create a dummy object with Rhode properties and point to them ?

jlekowski:
@GuilhemN, I don't find a test very `unit` if you provide actual dependencies (and their actual dependencies, and actual dependencies of these dependencies), and run the functionality around them.
But this is your project, and if you're happy with this way of testing it, I'm fine with it.

GuilhemN:
Well from my experience mocking is more likely to hide bugs and just complexify tests. Furthermore for data objects.

Thank you!

dbu:
RE unit test: imho its better to not mock value objects. and in this case, we do want to know if things actually work together, so this seems the right thing. thanks!

Issue #1292, opened 2018-04-12
@Model Naming


thhan:
I'm trying to name a Model specificate.

An exmaple:

```php
/**
* @SWG\Response(
*     description="Successfully created",
*     response="201",
*     @Model(type=Folder::class, groups={"folder_detail"}, definition="Folder_Detail")
* )
*/
```

So in ApiDoc the response Model is named "Folder_Detail" and not "Folder3". Is something like this included in the project?

valiknet18:
I think it's will be a good feature.
I have many entities with same name (ex. User, User2, User3), i want to have ability to rename this models to Login, Registration, Profile.

PR #1293, opened 2018-04-12
Add name in @Model

thhan:
It add name in Model to documentates results better ApiDoc. #1292

### DIFF
--- a/Annotation/Model.php
+++ b/Annotation/Model.php
@@ -40,4 +40,9 @@ final class Model extends AbstractAnnotation
      * @var string[]
      */
     public $groups;
+
+    /**
+     * @var string
+     */
+    public $definition;
 }
--- a/Annotation/Model.php
+++ b/Annotation/Model.php
@@ -44,5 +44,5 @@ final class Model extends AbstractAnnotation
     /**
      * @var string
      */
-    public $definition;
+    public $name;
 }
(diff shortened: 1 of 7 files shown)


GuilhemN:
> You have to specify for each and each serialization group. I think this not handling easy anyway.

Well this affects the model directly, not just its name so it's different imo.

Wouldn't you prefer just having something like?
```yml
nelmio_api_doc:
    models:
        names:
            Foo: App\Entity\Foo
            Foo_light: {type: App\Entity\Foo, groups: [light]}

**Comment by GuilhemN** [comment]
:+1: for using the annotations though i'd rather use `@SWG\Definition::$definition`:
```php
/**
 * @SWG\Definition(definition={"FooDetail": null, "FooDetail": ["foo_detail"]})
 */
```
What do you think ?

goetas:
I stumbled in the same problem. I agre that the current implementation is not working well as it will require to define names for everything and will not work well when there are many available groups.

goetas:
@thhan 
> I suggest to add the definition in the entity annotation.

since currently no other metadata format are supported except of annotations  that will not work for us as the entities come from a third party package that we do not have control over it. having it defined in the configuration will work much better for us as example

GuilhemN:
Just merged https://github.com/nelmio/NelmioApiDocBundle/pull/1312. Thanks !

PR #1294, opened 2018-04-12
Ignore unsupported HTTP methods

### LINKED ISSUE #1290: Link & Unlink verbs run a ContextErrorException
Hi, 

I'm running a sf 3.4 application. 
Since I migrate nelmio to 3.x if I try to doc a method that use the verbs `link` or `unlink`  (`@FOS\RestBundle\Controller\Annotations\Unlink`) the web page conclude in a "Notice : undefined index: [link|unlink]".
Is that normal ? Should I stop using those verbs if I want to doc my route ?

notFloran:
If an action use a LINK, UNLINK or COPY methods the bundle fails with: `Notice: Undefined index: link
 in vendor/nelmio/api-doc-bundle/Describer/SwaggerPhpDescriber.php (line 174)`

OpenAPI doesn't supports the LINK and UNLINK methods so the bundle must ignore this methods.

Fix #1290

### DIFF
--- a/Describer/SwaggerPhpDescriber.php
+++ b/Describer/SwaggerPhpDescriber.php
@@ -206,6 +206,7 @@ final class SwaggerPhpDescriber implements ModelRegistryAwareInterface
                 $path = $this->normalizePath($route->getPath());
                 $httpMethods = $route->getMethods() ?: Swagger::$METHODS;
                 $httpMethods = array_map('strtolower', $httpMethods);
+                $httpMethods = array_intersect($httpMethods, Swagger::$METHODS);
 
                 yield $method => [$path, $httpMethods];
             }
--- a/Describer/SwaggerPhpDescriber.php
+++ b/Describer/SwaggerPhpDescriber.php
@@ -209,10 +209,10 @@ final class SwaggerPhpDescriber implements ModelRegistryAwareInterface
                 $path = $this->normalizePath($route->getPath());
                 $httpMethods = $route->getMethods() ?: Swagger::$METHODS;
                 $httpMethods = array_map('strtolower', $httpMethods);
-                $validHttpMethods = array_intersect($httpMethods, Swagger::$METHODS);
+                $supportedHttpMethods = array_intersect($httpMethods, Swagger::$METHODS);
 
-                if (empty($validHttpMethods)) {
-                    $this->logger->warning('No valid HTTP method for path', [
+                if (empty($supportedHttpMethods)) {
+                    $this->logger->warning('None of the HTTP methods specified for path {path} are supported by swagger-ui, skipping this path', [
                         'path' => $path,
                         'methods' => $httpMethods,
                     ]);
@@ -220,7 +220,7 @@ final class SwaggerPhpDescriber implements ModelRegistryAwareInterface
                     continue;
                 }
 
-                yield $method => [$path, $validHttpMethods];
+                yield $method => [$path, $supportedHttpMethods];
             }
         }
     }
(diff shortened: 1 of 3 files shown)


dbu:
i just noticed a similar problem with an end point that supports COPY. are you sure that the javascript part can not handle this? if i understand correctly, this PR will simply drop those methods without reporting that there is a problem. is there anything better we can do?

dbu:
i just noticed a similar problem with an end point that supports COPY. are you sure that the javascript part can not handle this? if i understand correctly, this PR will simply drop those methods without reporting that there is a problem. is there anything better we can do?

notFloran:
> i just noticed a similar problem with an end point that supports COPY. are you sure that the javascript part can not handle this? 

Swagger/OpenAPI only supports : get, put, post, delete, options, head, patch. So this end points  will never display in the javascript parts.

See: 

* https://github.com/OAI/OpenAPI-Specification/issues/480
* https://github.com/zircote/swagger-php/issues/321
* https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#pathItemObject

> if i understand correctly, this PR will simply drop those methods without reporting that there is a problem. is there anything better we can do?

Yes, methods that are not in `['get', 'post', 'put', 'patch', 'delete', 'options', 'head']` are ignored to allow the page to be displayed.

I don't see how to improve that, maybe a log ?

Describer/SwaggerPhpDescriber.php
```diff
@@ -206,6 +206,7 @@ private function getMethodsToParse(): \Generator
                 $path = $this->normalizePath($route->getPath());
                 $httpMethods = $route->getMethods() ?: Swagger::$METHODS;
                 $httpMethods = array_map('strtolower', $httpMethods);
+                $httpMethods = array_intersect($httpMethods, Swagger::$METHODS);
```
dbu: what happens if the methods become an empty array here, because only unsupported methods are in the list?

also, is it invalid for open api spec to specify this methods, or is it "just" a bug of the swagger-ui library? if its a problem of that library, we should see if we can do something later in the process, or have a configuration option, so that people that care about the specification of their api can get a full spec that does not drop methods.

notFloran: > what happens if the methods become an empty array here, because only unsupported methods are in the list?

Nothing, in the rest of the code we go through the methods to consolidate the annotations and add them to the analysis so if a path does not have methods it is ignored.

I have changed the code, if there are no valid methods we go to the next route.

> also, is it invalid for open api spec to specify this methods, or is it "just" a bug of the swagger-ui library? if its a problem of that library, we should see if we can do something later in the process, or have a configuration option, so that people that care about the specification of their api can get a full spec that does not drop methods.

It's invalid for OpenAPI spec, the spec only supports : get, put, post, delete, options, head, patch.

See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#pathItemObject for the spec and this issue https://github.com/OAI/OpenAPI-Specification/issues/480

Describer/SwaggerPhpDescriber.php
```diff
@@ -206,8 +209,18 @@ private function getMethodsToParse(): \Generator
                 $path = $this->normalizePath($route->getPath());
                 $httpMethods = $route->getMethods() ?: Swagger::$METHODS;
                 $httpMethods = array_map('strtolower', $httpMethods);
+                $validHttpMethods = array_intersect($httpMethods, Swagger::$METHODS);
 
-                yield $method => [$path, $httpMethods];
+                if (empty($validHttpMethods)) {
+                    $this->logger->warning('No valid HTTP method for path', [
```
GuilhemN: What about replacing `no valid` by `no supported`? It is less alarming imo :)

dbu: to help the user some more, we could even say: "None of the HTTP methods specified for path {path} are supported by swagger-ui, skipping this path"

GuilhemN:
Thank you @notFloran :)

PR #1295, opened 2018-04-13
@SWG\Response shoud be in @SWG\Get or @SWG\Post

busaev:
I modify documentation following this resource http://bfanger.nl/swagger-explained/#schemaObject
Example in current version is don't work.


GuilhemN:
The `@SWG\Get` annotation is not necessary using this bundle, it is implicit and the bundle automatically adds it based on your symfony routes.

PR #1296, opened 2018-04-16
Add the swagger spec info name in <title>

magnetik:
It can be confusing to have only "NelmioApiDocBundle" in the title bar when browing the doc.

I've added the title of the documentation so it's clearer.


GuilhemN:
Thank you @magnetik!

PR #1297, opened 2018-04-16
Load class NelmioApiDocBundle annotations

### LINKED ISSUE #993: add ability to define common configuration via class level annotations

magnetik:
This should fix #993, at least partially. 

My need was to load `@Security` class wize.


magnetik:
It might be "too stupid" when declaring the annotation on both the class and method, but it felt overkill to handle this.

GuilhemN:
Thank you @magnetik :)

PR #1301, opened 2018-04-20
Improve support for JMS dictionaries (associative arrays)

goetas:
Doc: https://swagger.io/docs/specification/data-models/dictionaries/

### DIFF
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -98,11 +98,9 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
                     $property->setType('object');
 
                     $typeDef = $this->findPropertyType($type, $groups);
-                    $ref = new \ReflectionObject($property);
-                    $p = $ref->getProperty('additionalProperties');
-                    $p->setAccessible(true);
+
                     // in the case of a virtual property, set it as free object type
-                    $p->setValue($property, $typeDef ?: []);
+                    $property->merge(['additionalProperties' => $typeDef ?: []]);
 
                     continue;
                 } else {
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -92,33 +92,35 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
                 continue;
             }
 
-            if ($type = $this->getNestedTypeInArray($item)) {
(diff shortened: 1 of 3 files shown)


ModelDescriber/JMSModelDescriber.php
```diff
@@ -92,33 +92,35 @@ public function describe(Model $model, Schema $schema)
...
-                $property->setType('array');
-                $property = $property->getItems();
+            if ($nestedType = $this->getNestedTypeInArray($item)) {
+                list($type, $isHash) = $nestedType;
+                if ($isHash) {
+                    $property->setType('object');
+
+                    $typeDef = $this->findPropertyType($type, $groups);
+                    $ref = new \ReflectionObject($property);
+                    $p = $ref->getProperty('additionalProperties');
```
goetas: Not really happy about this but the current swagger-php api does not allow to set `additionalProperties` except via constructor args

ModelDescriber/JMSModelDescriber.php
```diff
@@ -92,33 +92,35 @@ public function describe(Model $model, Schema $schema)
                 continue;
             }
 
-            if ($type = $this->getNestedTypeInArray($item)) {
-                $property->setType('array');
-                $property = $property->getItems();
+            if ($nestedType = $this->getNestedTypeInArray($item)) {
+                list($type, $isHash) = $nestedType;
```
lsmith77: I guess unless we introduce a new class for the type, we are stuck having to build an array as the return value :-/

goetas: true.. but not sure if it makes sense to introduce a new class a private method

ModelDescriber/JMSModelDescriber.php
GuilhemN: Actually you can call `$schema->merge(['additionnalProperties' => ...])`. (The absence of a setter is most likely just a mistake)

goetas: ah! nice!

goetas: fixed

ModelDescriber/JMSModelDescriber.php
GuilhemN: I think it's fine to use an array in this case

lsmith77: yeah. if we dont have a general case where it would help and given that its a private method

GuilhemN:
Thank you @goetas :)

Issue #1302, opened 2018-04-22
How i can add "consumes"?


Gemorroj:
```php
    /**
     * @Route("/api/private/categories/add", methods={"POST"}, defaults={"_format": "json"})
     * @Operation(
     *     consumes={"application/x-www-form-urlencoded"},
     * )
     * @SWG\Response(
     *     response=200,
     *     description="Добавление категории",
     *     @Model(type=Category::class, groups={"category"}))
     * )
     * @SWG\Response(
     *     response=400,
     *     description="Ошибка валидации"
     * )
     * @SWG\Parameter(
     *     name="categoryName",
     *     in="formData",
     *     type="string",
     *     description="Имя категории",
     *     required=true
     * )
     * @Security(name="Bearer")
     */
```

i have error: `User Notice: Unable to merge @SWG\Post() in \PrivateController->addCategoryAction()`
I do not use `FosRestBundle`

chregu:
Exactly the same question/issue here.
Also with eg `@Operation(operationId="myownId")`

GuilhemN:
This is fixed by #1321 :)

PR #1308, opened 2018-04-26
check if method/path already exists and merge if yes

chregu:
Fix for #1302 
The problem is, that if an existing definition already exists, swagger php refuses to "merge" that. This patch check for existing definition, and if it finds one, merges it.
That solves the issue with having @Route and @Operation in one definition

No idea, if this has some side effects. It works for our case, at least

### DIFF
--- a/Describer/SwaggerPhpDescriber.php
+++ b/Describer/SwaggerPhpDescriber.php
@@ -67,7 +67,7 @@ final class SwaggerPhpDescriber implements ModelRegistryAwareInterface
     private function getAnnotations(Swagger $api): Analysis
     {
         $analysis = new Analysis();
-        $analysis->addAnnotation(new class($api) extends SWG\Swagger {
+        $this->addAnnotation($analysis, new class($api) extends SWG\Swagger {
             private $api;
 
             public function __construct(Swagger $api)
@@ -140,7 +140,7 @@ final class SwaggerPhpDescriber implements ModelRegistryAwareInterface
                         $operation->path = $path;
                         $operation->mergeProperties($annotation);
 
-                        $analysis->addAnnotation($operation, null);
+                        $this->addAnnotation($analysis, $operation);
                     }
 
                     continue;
@@ -152,7 +152,7 @@ final class SwaggerPhpDescriber implements ModelRegistryAwareInterface
                         $annotation->path = $path;
                     }
 
-                    $analysis->addAnnotation($annotation, null);
+                    $this->addAnnotation($analysis, $annotation);
 
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


magnetik:
You should probably add a test case for a case that was not working before

GuilhemN:
I always thought the solution to this was just to put everything under `@Operation` but that's not a nice solution in fact.

But I don't like merging several `@Operation` together much, maybe we should first extract `@Operation` annotations and then merge other annotations in them, instead of merging them together, what do you think ?

chregu:
I don't have a strong opinion on this. As long as I can have one `@Operation` and one `@Route` (or whatever other annotation), I'd be more than happy.

Just putting everything into `@Operation` wouldn't help, if you use `@Route` for other things like Symfony Routing. That would still fail with the current solution, right?

dbu:
@GuilhemN what do you think of chregu's last question?

GuilhemN:
> Just putting everything into `@Operation` wouldn't help, if you use `@Route` for other things like Symfony Routing. That would still fail with the current solution, right?

I meant putting only annotations related to swagger under `@Operation` 😃 No need to move `@Route` so it should be work fine.

And I've reconsidered my position against merging operations, I'm fine with it after all, we already do it in other places in fact. The only thing that bothers me here is that all annotations are checked at each call to ``addAnnotation``, I'd rather do something like:
```diff
diff --git a/Describer/SwaggerPhpDescriber.php b/Describer/SwaggerPhpDescriber.php
index 64b44a7..13d2160 100644
--- a/Describer/SwaggerPhpDescriber.php
+++ b/Describer/SwaggerPhpDescriber.php
@@ -115,6 +115,7 @@ final class SwaggerPhpDescriber implements ModelRegistryAwareInterface
             $nestedContext = clone $context;
             $nestedContext->nested = true;
             $implicitAnnotations = [];
+            $operations = [];
             $tags = [];
             $security = [];
             foreach ($annotations as $annotation) {
@@ -128,6 +129,7 @@ final class SwaggerPhpDescriber implements ModelRegistryAwareInterface
                         $operation->path = $path;
                         $operation->mergeProperties($annotation);

+                        $operations[$httpMethod] = $operation;
                         $analysis->addAnnotation($operation, null);
                     }

@@ -140,6 +142,7 @@ final class SwaggerPhpDescriber implements ModelRegistryAwareInterface
                         $annotation->path = $path;
                     }

+                    $operations[$operation->method] = $operation;
                     $analysis->addAnnotation($annotation, null);

                     continue;
@@ -186,7 +189,11 @@ final class SwaggerPhpDescriber implements ModelRegistryAwareInterface
                 }

                 $operation = new $annotationClass($constructorArg);
-                $analysis->addAnnotation($operation, null);
+                if (isset($operations[$httpMethod])) {
+                    $operations[$httpMethod]->mergeProperties($operation);
+                } else {
+                    $analysis->addAnnotation($operation, null);
+                }
             }
         }
```

I'll try to submit a PR with this change.

GuilhemN:
Please have a look at https://github.com/nelmio/NelmioApiDocBundle/pull/1321

GuilhemN:
Closing as I just merged #1321. Thank you!

PR #1224, opened 2018-02-09
[Forms] Getting child type better for collections. Format/Example extensions. 

aequasi:
Solves #1184

Issue #1184, opened 2018-01-12
Modeling Form Collections should convert entry_type


aequasi:
https://symfony.com/doc/current/reference/forms/types/collection.html#entry-type

When the spec gets generated, it ends up looking like:

    "categories": "[{Symfony\\Component\\Form\\Extension\\Core\\Type\\TextType}]",

when it should be:

    "categories": "[string]",

Even doing `['entry_type' => "string"]` makes the spec look like:

    "categories": "[{string}]",

Also, is it possible to set examples in a form type?

GuilhemN:
The default example we set for form collections most likely is wrong indeed (see https://github.com/nelmio/NelmioApiDocBundle/blob/master/ModelDescriber/FormModelDescriber.php#L122), please submit a PR if you can think of an improvement.

> Also, is it possible to set examples in a form type?

Not for now but it could be, we just need to port https://github.com/nelmio/NelmioApiDocBundle/blob/2.x/Form/Extension/DescriptionFormTypeExtension.php.

PR #1309, opened 2018-04-27
Handle form collection types

goetas:
Fixes https://github.com/nelmio/NelmioApiDocBundle/issues/1184
Closes https://github.com/nelmio/NelmioApiDocBundle/pull/1224 (alternative implementation to it)

IMO this solution is better than #1224 because it does less "hacks" to infer the type, it just calls recursively itself solving the inner collection type. 

Since this PR just moves the code in a private method, suggest to have a look at a [split mode with no white spaces diff mode](https://github.com/nelmio/NelmioApiDocBundle/pull/1309/files?w=1) that should be much easier to review.


goetas:
happy to see you like it!

GuilhemN:
Thank you :)

PR #1311, opened 2018-05-02
Respect jms/serializer nested group exclusion

goetas:
Currently the logic used to infer groups from the JMS serializer is not exactly correct.
When using the [deeper-branches-group override](https://github.com/schmittjoh/serializer/blob/1.x/doc/cookbook/exclusion_strategies.rst#overriding-groups-of-deeper-branches-of-the-graph) feature, when an override is not found, the fallback is to the default group, while the current nelmio implementation is to fallback to the currently used groups.

This is visible on JMS [here](https://github.com/schmittjoh/serializer/blob/5824ce8fa1c2e7db57d47072a6ce9d2feec229da/src/JMS/Serializer/Exclusion/GroupsExclusionStrategy.php#L80) while the curren implementation does not override the group with the default one (see [here](https://github.com/nelmio/NelmioApiDocBundle/blob/2c72aa4efff07862514aaf2cf078fabc58c465d6/ModelDescriber/JMSModelDescriber.php#L74)).

The implementation might be not perfect but it works, feedback is welcome.

### DIFF
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -35,6 +35,11 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
     private $doctrineReader;
     private $previousGroups = [];
 
+    /**
+     * @var array
+     */
+    private $propertyTypeUseGroupsCache = [];
+
     public function __construct(
         MetadataFactoryInterface $factory,
         PropertyNamingStrategyInterface $namingStrategy,
@@ -183,17 +188,26 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
      */
     private function propertyTypeUsesGroups(array $type)
     {
+        if (!array_key_exists($type['name'], $this->propertyTypeUseGroupsCache)) {
+            return $this->propertyTypeUseGroupsCache[$type['name']];
+        }
+
         try {
             $metadata = $this->factory->getMetadataForClass($type['name']);
 
             foreach ($metadata->propertyMetadata as $item) {
-                if ($item->groups && $item->groups != [GroupsExclusionStrategy::DEFAULT_GROUP]) {
+                if (null !== $item->groups && $item->groups != [GroupsExclusionStrategy::DEFAULT_GROUP]) {
+                    $this->propertyTypeUseGroupsCache[$type['name']] = true;
+
                     return true;
                 }
             }
+            $this->propertyTypeUseGroupsCache[$type['name']] = false;
 
             return false;
         } catch (\ReflectionException $e) {
+            $this->propertyTypeUseGroupsCache[$type['name']] = null;
+
             return null;
         }
     }
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -33,6 +33,7 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
     private $factory;
     private $namingStrategy;
     private $doctrineReader;
+    private $previousGroups = [];
 
     public function __construct(
         MetadataFactoryInterface $factory,
@@ -70,8 +71,14 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
 
             $name = $this->namingStrategy->translateName($item);
             $groups = $model->getGroups();
+
+            $previousGroups = null;
             if (isset($groups[$name]) && is_array($groups[$name])) {
+                $previousGroups = $groups;
                 $groups = $model->getGroups()[$name];
+            } elseif (!isset($groups[$name]) && !empty($this->previousGroups[spl_object_hash($model)])) {
+                // $groups = $this->previousGroups[spl_object_hash($model)]; use this for jms/serializer 2.0
+                $groups = false === $this->propertyTypeUsesGroups($item->type) ? null : [GroupsExclusionStrategy::DEFAULT_GROUP];
             } elseif (is_array($groups)) {
                 $groups = array_filter($groups, 'is_scalar');
             }
@@ -97,7 +104,7 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
                 continue;
             }
 
-            $this->describeItem($item->type, $property, $groups);
+            $this->describeItem($item->type, $property, $groups, $previousGroups);
         }
     }
 
@@ -121,10 +128,11 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
     /**
      * @param string     $type
      * @param array|null $groups
+     * @param array|null $previousGroups
      *
      * @return array|null
      */
-    private function findPropertyType(string $type, array $groups = null)
+    private function findPropertyType(string $type, array $groups = null, array $previousGroups = null)
     {
         $typeDef = [];
         if (in_array($type, ['boolean', 'string', 'array'])) {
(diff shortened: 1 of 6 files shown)


ModelDescriber/JMSModelDescriber.php
```diff
@@ -71,8 +72,13 @@ public function describe(Model $model, Schema $schema)
 
             $name = $this->namingStrategy->translateName($item);
             $groups = $model->getGroups();
+
+            $usingNestedGroups = false;
             if (isset($groups[$name]) && is_array($groups[$name])) {
                 $groups = $model->getGroups()[$name];
+                $usingNestedGroups = true;
+            } elseif (!isset($groups[$name]) && !empty($this->usingNestedGroups[spl_object_hash($model)])) {
+                $groups = [GroupsExclusionStrategy::DEFAULT_GROUP];
```
goetas: here might also work having just `[]`

goetas: solved with https://github.com/nelmio/NelmioApiDocBundle/pull/1311/commits/b807ef4d8dffb427e8f32f444c2809f29c31208c

ModelDescriber/JMSModelDescriber.php
```diff
@@ -71,8 +72,13 @@ public function describe(Model $model, Schema $schema)
 
             $name = $this->namingStrategy->translateName($item);
             $groups = $model->getGroups();
+
+            $usingNestedGroups = false;
             if (isset($groups[$name]) && is_array($groups[$name])) {
                 $groups = $model->getGroups()[$name];
+                $usingNestedGroups = true;
+            } elseif (!isset($groups[$name]) && !empty($this->usingNestedGroups[spl_object_hash($model)])) {
+                $groups = false === $this->propertyTypeUsesGroups($item->type) ? null : [GroupsExclusionStrategy::DEFAULT_GROUP];
```
goetas: `propertyTypeUsesGroups` is used to detect if the new group will have effect on the type. 
This avoids to create extra types when the group is going to be "Default" and all the props of the object do not use groups or are in the default group.

GuilhemN: Imo it is not worth the complexity, it might even be undesirable for some. I would stay simple and always use the default group :)

ModelDescriber/JMSModelDescriber.php
```diff
@@ -71,8 +72,13 @@ public function describe(Model $model, Schema $schema)
 
             $name = $this->namingStrategy->translateName($item);
             $groups = $model->getGroups();
+
+            $usingNestedGroups = false;
             if (isset($groups[$name]) && is_array($groups[$name])) {
                 $groups = $model->getGroups()[$name];
+                $usingNestedGroups = true;
+            } elseif (!isset($groups[$name]) && !empty($this->usingNestedGroups[spl_object_hash($model)])) {
```
GuilhemN: Why is `$this->usingNestedGroups` needed? Are groups inherited when there is no groups override?

goetas: When the first `if` branch is not matched, we need to understand if:
- $groups is null: no group exclusion policy was used 
- $groups is 'Default': this model was found trough a **previous** model that had groups defined but not for a property holding the current model

GuilhemN:
I think this introduces a new bug:
If you use a model with the same groups, one with nested groups, the other without, one will win over the other (and you don't know which).

goetas:
Good catch, will check

goetas:
For the serializer 1.x still have to fin a solution :-/ but for the version 2.x https://github.com/schmittjoh/serializer/pull/946 should make it simpler

goetas:
The code currently pushed here follows the upcoming 2.0 release strategy. 
Will put it on hold for few days and re-try to find a working solution for the serializer 1.x deep-group strategy

ModelDescriber/JMSModelDescriber.php
```diff
@@ -164,9 +171,12 @@ private function findPropertyType(string $type, array $groups = null)
...
             }
 
-            $typeDef['$ref'] = $this->modelRegistry->register(
-                new Model(new Type(Type::BUILTIN_TYPE_OBJECT, false, $type), $groups)
-            );
+            $model = new Model(new Type(Type::BUILTIN_TYPE_OBJECT, false, $type), $groups);
+            $typeDef['$ref'] = $this->modelRegistry->register($model);
+
+            if ($previousGroups) {
+                $this->previousGroups[spl_object_hash($model)] = $previousGroups;
```
GuilhemN: This suffers from the issue pointed in https://github.com/nelmio/NelmioApiDocBundle/pull/1311#issuecomment-387314391, right?

Why not using the same strategy as symfony in JMS\Serializer and keep the direct parent groups if none are specifically defined?

goetas: This implementation is slightly different than what was before (and it keeps the groups defined in the parent). [here](https://github.com/schmittjoh/serializer/pull/946) i have also changed the strategy directly in the serializer. Should not suffer from https://github.com/nelmio/NelmioApiDocBundle/pull/1311#issuecomment-387314391 but will verify...

goetas:
Rebased on the latest master

goetas:
@GuilhemN I have rebased this. I think that when combined https://github.com/nelmio/NelmioApiDocBundle/pull/1366 and https://github.com/nelmio/NelmioApiDocBundle/pull/1367 will give a pretty solid management of serialization groups.

I've verified that the issue discussed in https://github.com/nelmio/NelmioApiDocBundle/pull/1311#issuecomment-387314391 is still there.
IMO is still an edge case, but with this 3 PRs the amount of duplicated objects will be much less than what already is.

goetas:
personally I use this feature very often to manage complex api where is necessary to manage the visibility of fields when the same entity is represented multiple times.

example: 
the user object displayed in: 
- user profile (full info)
- user profile contained in a chat, the chat has all the info but the user entity contains less data
- user profile contained in a chat message, the chat message has all the info but the user entity only the basics

This is handled via:
- `@Model(type=User::class, groups={"Default"})` (fill view)
- `@Model(type=User\Chat::class, groups={"Default", "moderator":{"mini"}})` (mini view)
- `@Model(type=User\Chat\Message::class, groups={"Default", "sentBy":{"mini"}})` (micro view)

- `@Model(type=User\Chat::class, groups={"Default", "moderator":{"mini"}, "admin":{"Default"}})` (mini view for the moderator but full view for the admin)

In this way is possible to define simple nested grouping without having to create multiple views of all the parent objects.

This is currently a feature that jms/serializer offers and nelmio does not support properly. 
Users that do not use nested groups should not be affected at all by this change. 
On the other hand, **for users that use the nested groups, the currently generated documentation by nelmio-api-doc is completely wrong**. This PR in my opinion is more of a bug-fix rather than a new-feature. 

> Is this feature worth the support?

In my opinion is worth to support because the current implementation is wrong, the new implementation is still not perfect but more accurate that the actual.

goetas:
@GuilhemN rebased and implemented the latest suggestions

GuilhemN:
Merged in https://github.com/nelmio/NelmioApiDocBundle/commit/63cee64e6b02ef6f3685ee4a568a81444936d2fe, thanks!

goetas:
https://github.com/nelmio/NelmioApiDocBundle/pull/1384 broke the implementation of this since now groups are registered in a different context.... will have to work on this...

goetas:
This was relying on the fact that the object are visited following the order in the "controller" `@SWG` annotations... but now the models are described in two different places [here](https://github.com/nelmio/NelmioApiDocBundle/blob/c805beb3f5f02004c31ef5039483567d80430b38/ApiDocGenerator.php#L70) and [here](https://github.com/nelmio/NelmioApiDocBundle/blob/c805beb3f5f02004c31ef5039483567d80430b38/Model/ModelRegistry.php#L46) (the last one has a completely different context... making probably impossible to achieve what was proposed in this PR)

PR #1312, opened 2018-05-03
Implement alternative naming system via configuration

goetas:
This is an alternative implementation of https://github.com/nelmio/NelmioApiDocBundle/pull/1293 to solve https://github.com/nelmio/NelmioApiDocBundle/issues/1292. 
Implements a suggested approach by @GuilhemN expressed [here](https://github.com/nelmio/NelmioApiDocBundle/pull/1293#issuecomment-382792019)

```yml
nelmio_api_doc:
    models:
        names:
            - { alias: MainUser,        type: App\Entity\User}
            - { alias: MainUser_light,  type: App\Entity\User, groups: [light] }
            - { alias: MainUser_secret, type: App\Entity\User, areas: [private] }
            - { alias: MainUser,        type: App\Entity\User, groups: [standard], areas: [private] }
```

In this case the class ``App\Entity\User`` will be named:

- ``MainUser`` in the default area and default group
- ``MainUser_light`` in the default area when the group is equal to ``light``
- ``MainUser_secret`` in the ``private`` area
- ``MainUser_light`` in the ``private`` area when the group is equal to ``light``
@GuilhemN if you think this can be a good alternative to be merged I can add also tests


goetas:
great to hear! Will do in the next days!

goetas:
Tests added. 

Is StyleCI gone?

DependencyInjection/Configuration.php
```diff
@@ -85,6 +85,23 @@ public function getConfigTreeBuilder()
...
+                            ->useAttributeAsKey('alias')
+                            ->prototype('array')
+                                ->children()
+                                    ->scalarNode('type')->end()
+                                    ->arrayNode('groups')
+                                        ->defaultValue([])
+                                        ->prototype('scalar')->end()
+                                    ->end()
+                                    ->arrayNode('areas')
+                                        ->defaultValue(['default'])
```
goetas: I've set  `default` as default area. Hope is correct

Model/ModelRegistry.php
```diff
@@ -105,6 +110,23 @@ private function generateModelName(Model $model): string
...
+    /**
+     * @param Model $model
+     *
+     * @return string|null
+     */
+    private function getAlternativeName(Model $model)
+    {
+        $type = $model->getType();
+        foreach ($this->alternativeNames as $alternativeName => $criteria) {
+            if ($type->getClassName() === $criteria['type'] && $criteria['groups'] == $model->getGroups()) {
```
goetas: here `==` comparison is intentional (and tested) to make `$model->getGroups()=null` being equal to `$criteria['groups']=[]`

Model/ModelRegistry.php
```diff
@@ -105,6 +110,27 @@ private function generateModelName(Model $model): string
...
+     * @return string|null
+     */
+    private function getAlternativeName(Model $model)
+    {
+        $type = $model->getType();
+        foreach ($this->alternativeNames as $alternativeName => $criteria) {
+            if (
+                Type::BUILTIN_TYPE_OBJECT === $type->getBuiltinType() &&
+                $type->getClassName() === $criteria['type'] &&
+                $criteria['groups'] == $model->getGroups()
```
goetas: here `==` comparison is intentional (and tested) to make `$model->getGroups()=null` being equal to `$criteria['groups']=[]`

Resources/doc/alternative_names.rst
```diff
@@ -0,0 +1,26 @@
...
+                MainUser: App\Entity\User
+                MainUser_light: { type: App\Entity\User, groups: [light]}
+                MainUser_secret: { type: App\Entity\User, groups: [light], areas: [private]}
+
+
+In this case the class ``App\Entity\User`` will be named:
+
+- ``MainUser`` in the default area and default group
+- ``MainUser_light`` in the default area when the group is equal to ``light``
+- ``MainUser_secret`` in the ``private`` area when the group is equal to ``light``
```
GuilhemN: Are you sure about this one? Won't `MainUser_light` be used since it comes first ?

goetas:
@GuilhemN I have slightly changed the implementation to allow to create the same alias name in two different area

Resources/doc/alternative_names.rst
goetas: Not sure to follow this. `MainUser_light` is used only for the default area, the `private` area does not inherit any alias from the default

goetas:
@GuilhemN ping :)

DependencyInjection/NelmioApiDocExtension.php
```diff
@@ -59,8 +59,13 @@ public function load(array $configs, ContainerBuilder $container)
 
         $container->setParameter('nelmio_api_doc.areas', array_keys($config['areas']));
         foreach ($config['areas'] as $area => $areaConfig) {
+            $nameAliases = $this->findNameAliases($config['models']['names'], $area);
+
+            print_r($nameAliases);
```
GuilhemN: Debug? :)

DependencyInjection/Configuration.php
GuilhemN: Isn't it more logical to add them to all areas by default? (Empty field)

DependencyInjection/NelmioApiDocExtension.php
goetas: my bad! will fix it!

goetas:
it looks github does not want me to comment!
implement alternative naming system via configuration by goetas pull request 1312 nelmio nelmioapidocbundle

> Isn't it more logical to add them to all areas by default? (Empty field)

Not really sure about it. If we do as you suggest we need to handle priorities and exclusions... as example overriding (or even harder, unsetting) an alias set in the default area.

GuilhemN:
> Not really sure about it. If we do as you suggest we need to handle priorities and exclusions... as example overriding (or even harder, unsetting) an alias set in the default area.

Why would we need priorities/exclusions? Users would just need to be more explicit if they want to define something only in `default` but I don't think it's the most common case so imo it's better to add rules to all areas by default.

goetas:
@GuilhemN you were right in https://github.com/nelmio/NelmioApiDocBundle/pull/1312#discussion_r192116642

I have rebased, updated tests and updated the documentation. Aliases that do not specify an area are applied to all the areas.

GuilhemN:
Last minor comment, that's great :+1:

goetas:
Done!

DependencyInjection/NelmioApiDocExtension.php
```diff
@@ -152,4 +155,21 @@ public function load(array $configs, ContainerBuilder $container)
...
+
+    private function findNameAliases(array $names, string $area): array
+    {
+        $nameAliases = array_filter($names, function (array $aliasInfo) use ($area) {
+            return empty($aliasInfo['areas']) || in_array($area, $aliasInfo['areas'], true);
+        });
+
+        $aliases = [];
+        foreach ($nameAliases as $nameAlias) {
+            $aliases[$nameAlias['alias']] = [
```
GuilhemN: This feels weird, last definition of an alias will win but first definition of a type will win. What about making the last win too in the ModelRegistry?

Model/ModelRegistry.php
GuilhemN: Just for the record, it may be more logical to differenciate these two values (like symfony does if I remember right).

goetas:
I would keep the `==` comparison as it is compatible with jms serializer

GuilhemN:
@goetas ok, let's keep it :)
Thank you for working on this new feature!

goetas:
Thanks to you for reviewing it!

carlosgoro:
Hi.

I'm reviewing the documentation in Symfony ( https://symfony.com/doc/master/bundles/NelmioApiDocBundle/alternative_names.html ), and I can't find the correct way to use this feature.

How can I reference the alias in the @Model annotation instead of @Model(type=EntityName::class, groups={"groupName"}) ?

Thanks in advance.

GuilhemN:
I think we should always register models with alternative names and register them first to ensure their name is not suffixed. This would allow to reference them directly and would avoid duplication group names, wdyt?

goetas:
ah, to avoid collisions? an alternative name that in reality already exists in the codebase ?

GuilhemN:
That's it and also to allow to use references without ever using `@Model`.

goetas:
could be an  idea.

GuilhemN:
Will try to do it during the summer, or would you like to submit a PR @carlosgoro?

bobvandevijver:
If I understand correctly, the lightbulb part in the [documentation](https://symfony.com/doc/current/bundles/NelmioApiDocBundle/alternative_names.html) doesn't work yet?

I would really like to see aliases which I can simply re-use 😄

PR #1316, opened 2018-05-07
add more entries to the faq

dbu:
* shared definitions
* optional path parameters

### DIFF
--- a/Resources/doc/faq.rst
+++ b/Resources/doc/faq.rst
@@ -1,10 +1,128 @@
 Frequently Asked Questions (FAQ)
 ================================
 
-* Q: How do I fix 404 or 406 HTTP status on NelmioApiDocBundle assets files (css, js, images)?
+Sharing parameter configuration
+-------------------------------
 
-  A: Just execute this command to solve it:
+Q: I use the same value in multiple end points. How can I avoid duplicating the descriptions?
 
-  .. code-block:: bash
+A: You can configure ``definitions`` in the nelmio_api_doc configuration and then reference them:
 
-      $ bin/console assets:install --symlink
+.. code-block:: yaml
+
+    # config/nelmio_api_doc.yml
+    nelmio_api_doc:
+        documentation:
+            definitions:
+                NelmioImageList:
+                    description: "Response for some queries"
+                    type: object
+                    properties:
+                        total:
+                            type: integer
+                            example: 42
+                        items:
+                            type: array
+                            items:
+                                $ref: "#/definitions/ImageMetadata"
+
+.. code-block:: php
+
+    // src/App/Controller/NelmioController.php
+
+    /**
+     * @SWG\Response(
+     *     response=200,
+     *     description="List of image definitions",
+     *     @SWG\Schema(
+     *       type="object",
+     *       title="ListOperationsResponse",
+     *       additionalProperties={"$ref": "#/definitions/NelmioImageList"}
+     *     )
+     */
+
+Optional Path Parameters
+------------------------
+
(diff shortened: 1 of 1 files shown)


Resources/doc/faq.rst
```diff
@@ -1,10 +1,128 @@
...
+     *     response=200,
+     *     description="Json object with all user meta data or a json string with the value of the requested field"
+     * )
+     */
+    public function getAction(string $user, string $metaName = null)
+    {
+        ...
+    }
+
+A: Optional path parameters are not supported by the swagger specification. The solution to this is to define two
```
phansys: I think we should use "OpenAPI" instead of "swagger" here and later, at line 115.

Resources/doc/faq.rst
```diff
@@ -1,10 +1,128 @@
...
+    }
+
+The first action is redundant for Symfony, but adds all the relevant documentation for swagger.
+
+Asset files not loaded
+----------------------
+
+Q: How do I fix 404 or 406 HTTP status on NelmioApiDocBundle assets files (css, js, images)?
+
+A: The assets normally are installed by composer in the ``installAssets`` step.
```
phansys: Since `installAssets` method is provided by `SensioDistributionBundle` and can be run by composer only if configured in `scripts` section, I think we could reword this sentence to something like the following:
"The assets normally are installed by composer if any of their command events (usually "post-install-cmd" or "post-install-cmd") triggers the `ScriptHandler::installAssets` script. If you have not set up that script, you can manually execute this command ..."

Resources/doc/faq.rst
dbu: absolutely right. old habits of mine :-/

Resources/doc/faq.rst
dbu: i only moved this around ;-)

but i agree with you, i find your suggestion more understandable. updating.

Resources/doc/faq.rst
```diff
@@ -1,10 +1,129 @@
...
+     *     description="Json object with all user meta data or a json string with the value of the requested field"
+     * )
+     */
+    public function getAction(string $user, string $metaName = null)
+    {
+        ...
+    }
+
+A: Optional path parameters are not supported by the OpenAPI specification. The solution to this is to define two
+   separate actions in your controller. For example::
```
GuilhemN: Shouldn't we do that automatically then ?

dbu: i think its difficult to do it automatically. its not very clean what we had in our example, as the response was different depending on whether the optional parameter is set or not. splitting it allows for a precise description. doing all that with annotations would be cumbersome, as right now the annotations are on the whole PHP method which represents a single api end point. if one PHP method represents more than one end point, we need to tell for each annotation if its about both or only one end point.

but if you have a cool idea for it, it could be nice. for uses cases where the response format is the same (e.g. paging with default page 0) its indeed a lot of redundancy. then again with paging, there is no loss if you have to specify page 0 explicitly.

dbu:
shall i merge this myself? i am used to the person doing the review also hitting merge, rather than the author. to the point where i feel awkward merging my own code. but if you think its good that i hit the button myself after getting the approval reviews, i can do that ;-)

Resources/doc/faq.rst
GuilhemN: > if one PHP method represents more than one end point, we need to tell for each annotation if its about both or only one end point.

Good point, let's keep your answer as is 👍

GuilhemN:
Sorry I was waiting for your answer and didn't come back to this issue until now 😄 
Thank you!

PR #1319, opened 2018-05-09
Fix Model annotation example in docs

### LINKED ISSUE #1318: Using ref=@Model causes "Unexpected parameter"
Hi, I have this annotation:

```php
    /**
     * @Route("/", name="get_footers")
     * @Method({"GET"})
     * @SWG\Response(
     *      response="200",
     *      @SWG\Schema(
     *          type="object",
     *          @SWG\Property(property="data", ref=@Model(Footers::class))
     *      )
     * )
     */
```

This is almost the same as [example in docs](https://symfony.com/doc/master/bundles/NelmioApiDocBundle/index.html#use-models) (the second one in the green box).

However I'm getting:
```
Exception thrown when handling an exception (Symfony\Component\Config\Exception\FileLoaderLoadException: User Notice: Unexpected parameter in @Nelmio\ApiDocBundle\Annotation\Model() in /.../Controller/ (which is being imported from "/.../config/routes/annotations.yaml"). Make sure annotations are installed and enabled.)
```

But the `Model` annotation is imported and all other SWG annotations are working. Also the `Model` annotation works if used on the root level of `Response` (ie. not inside Property).

Is this a bug in docs, in the library or am I doing something wrong?

Thanks :)

OndraM:
Fixes #1318.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -218,7 +218,7 @@ This annotation has two options:
           *
           * @SWG\Response(
           *     @SWG\Schema(type="object",
-          *         @SWG\Property(property="foo", ref=@Model(FooClass::class))
+          *         @SWG\Property(property="foo", ref=@Model(type=FooClass::class))
           *     )
           * )
           */


GuilhemN:
Good catch, thank you @OndraM!

PR #1304, opened 2018-04-23
Allow to use custom form type with parent

magnetik:
This fixes #1303 

I'm not quite sure about the condition, but it seems to work =)

PR #1320, opened 2018-05-09
Support for extended builtin form types

maryo:
The same problem as in https://github.com/nelmio/NelmioApiDocBundle/pull/1304
I just tried to do it my way and I took into account the PR comments
The tests are the same as in the mentioned PR, just renamed

### DIFF
--- a/ModelDescriber/FormModelDescriber.php
+++ b/ModelDescriber/FormModelDescriber.php
@@ -16,10 +16,12 @@ use Nelmio\ApiDocBundle\Describer\ModelRegistryAwareInterface;
 use Nelmio\ApiDocBundle\Describer\ModelRegistryAwareTrait;
 use Nelmio\ApiDocBundle\Model\Model;
 use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\FormType;
 use Symfony\Component\Form\FormConfigBuilderInterface;
 use Symfony\Component\Form\FormFactoryInterface;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\Form\FormTypeInterface;
+use Symfony\Component\Form\ResolvedFormTypeInterface;
 use Symfony\Component\PropertyInfo\Type;
 
 /**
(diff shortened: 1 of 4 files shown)


ModelDescriber/FormModelDescriber.php
```diff
@@ -88,106 +90,108 @@ private function parseForm(Schema $schema, FormInterface $form)
...
-                    } else {
-                        $property->setType($type)->setEnum($enums);
-                    }
-                }
-
-                return true;
             }
 
-            if ('checkbox' === $blockPrefix) {
-                $property->setType('boolean');
```
maryo: Was the missing return here intentional or not?

Tests/Functional/Form/ExtendedBuiltinType.php
```diff
@@ -0,0 +1,36 @@
...
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver
+            ->setDefaults([
+                'choices' => [
+                    'foo' => 'foo',
+                    'bar' => 'bar',
+                ],
+            ])
+            ->setRequired('required_option');
```
maryo: I've added a required option. It was actually the primary reason why I created this PR. I have my own extended EntityType in my project which was not considered as a builtin type thus it was added into the modelRegistry and later it was failing because of missing `class` option.

ModelDescriber/FormModelDescriber.php
GuilhemN: I don't think it was, good catch 👍

GuilhemN:
That's great, thank you!

ModelDescriber/FormModelDescriber.php
```diff
@@ -209,8 +215,21 @@ private function isNumbersArray(array $array): bool
...
+     *
+     * @return ResolvedFormTypeInterface|null
+     */
+    private function getBuiltinFormType(ResolvedFormTypeInterface $type)
     {
-        return 0 === strpos($type, 'Symfony\Component\Form\Extension\Core\Type');
+        do {
+            $class = get_class($type->getInnerType());
+
+            if (FormType::class !== $class && 0 === strpos($class, 'Symfony\Component\Form\Extension\Core\Type\\')) {
```
phansys: I know this check is here before this PR, but how is supposed the condition `('entity' === $blockPrefix)` inside `findFormType()` to be fulfilled if "entity" block prefix is declared inside `Symfony\Bridge\Doctrine\Form\Type\` namespace?

maryo: You're right. It won't be fulfilled. I will try to fix it.

phansys: Note that I'm not experiencing any issue since I haven't checked out this commit yet, just wondering based on what I see here.

maryo: I see. Very good catch .-)

PR #1323, opened 2018-05-10
FormModelDescriber entity form type regression

maryo:
I fixed a regression bug I introduced with my last PR which @phansys noticed.
Check https://github.com/nelmio/NelmioApiDocBundle/pull/1320/files#r187460288 for explanation

### DIFF
--- a/ModelDescriber/FormModelDescriber.php
+++ b/ModelDescriber/FormModelDescriber.php
@@ -225,7 +225,15 @@ final class FormModelDescriber implements ModelDescriberInterface, ModelRegistry
         do {
             $class = get_class($type->getInnerType());
 
-            if (FormType::class !== $class && 0 === strpos($class, 'Symfony\Component\Form\Extension\Core\Type\\')) {
+            if (FormType::class === $class) {
+                return null;
+            }
+
+            if ('entity' === $type->getBlockPrefix()) {
+                return $type;
+            }
+
+            if (0 === strpos($class, 'Symfony\Component\Form\Extension\Core\Type\\')) {
                 return $type;
             }
         } while ($type = $type->getParent());
--- a/Tests/Functional/Form/EntityType.php
+++ b/Tests/Functional/Form/EntityType.php
@@ -1,0 +1,31 @@
+<?php
+
+/*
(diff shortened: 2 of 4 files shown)


Tests/Functional/Form/EntityType.php
```diff
@@ -0,0 +1,31 @@
...
+        $resolver
+            ->setDefault('multiple', false)
+            ->setRequired('class');
+    }
+
+    public function getParent(): string
+    {
+        return ChoiceType::class;
+    }
+}
```
maryo: I wanted to cover it with a basic test and don't think mocking Doctrine and EntityType from Doctrine Bridge would worth it so I created a dummy one.

ModelDescriber/FormModelDescriber.php
```diff
@@ -218,18 +218,18 @@ private function isNumbersArray(array $array): bool
...
-     * @return ResolvedFormTypeInterface|null
+     * @return bool
      */
-    private function getBuiltinFormType(ResolvedFormTypeInterface $type)
+    private function isBuiltinType(ResolvedFormTypeInterface $type): bool
     {
         do {
             $class = get_class($type->getInnerType());
 
             if (FormType::class !== $class && 0 === strpos($class, 'Symfony\Component\Form\Extension\Core\Type\\')) {
```
GuilhemN: What about checking for `EntityType` namespace here as well instead?
Your change looked more logical to me as we didn't have to iterate over all the hierarchy twice.

maryo: It probably was. So i reverted the  change and I check the 'entity' prefix there now because otherwise mocking Doctrine in the test would be needed.

But on the other hand the condition is now more complex (but still simple enough) and is dependent on the logic inside `findFormType`. If someone in the future will need to add another condition in the `findFormType` from another namespace then the `getBuiltinFormType` method will need to change also.

Maybe it would be good to implement it using a map like this.

`
'text' => 'string',
'number => 'number',
// ...
'datetime' => [$this, 'desribeDateTimeType'],
'choice' => [$this, 'describeChoiceType'],
`

GuilhemN:
Thank you @maryo!

PR #1331, opened 2018-05-28
Run CI pipeline with PHP 7.2

markuspoerschke:
This will resolve #1198.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: php
 php:
   - 7.0
   - 7.1
+  - 7.2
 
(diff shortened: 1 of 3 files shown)


dbu:
the `each` deprecation might be in phpunit, but can you check the bundle code if there is anything about `each` in it?

markuspoerschke:
The `each` deprecation was caused by PHPUnit itself. Updating to PHPUnit 6.5 solved the problem.

GuilhemN:
Thank you for taking care of this!

PR #1332, opened 2018-05-28
Prevent Swagger UI from using resources via CDNs

markuspoerschke:
The API documentation of a service should not "leak" information to a third party. Using Google Fonts is not nessecary in this particular situation.

Also there is a new data protection law in European Union (GDRP). To avoid data protection issues the usage of google fonts was removed. The system fonts provided by the operating systems are used instead.

### DIFF
--- a/Resources/public/style.css
+++ b/Resources/public/style.css
@@ -171,7 +171,7 @@ header #logo img {
 
 #formats {
   text-align:right;
-  font-family: Open Sans,sans-serif;
+  font-family: sans-serif;
   width: 100%;
   max-width: 1460px;
   padding: 0px 60px;
@@ -287,5 +287,5 @@ text-align:center;
 .swagger-ui .info .title small pre,
 .swagger-ui .scopes h2,
 .swagger-ui .errors-wrapper hgroup h4 {
-  font-family: Open Sans,sans-serif!important;
+  font-family: sans-serif!important;
 }
--- a/Resources/views/SwaggerUi/index.html.twig
+++ b/Resources/views/SwaggerUi/index.html.twig
@@ -11,7 +11,6 @@ file that was distributed with this source code. #}
     <meta charset="UTF-8">
     <title>{{ swagger_data.spec.info.title }} - NelmioApiDocBundle</title>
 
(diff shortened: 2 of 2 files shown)


markuspoerschke:
@dbu Actually this will only change the design. Nothing will break if the font is not available. If you want to have a template that is closest to swagger ui and therefore to wait until Google Fonts was removed there, then I recommend to close this PR becasue adding a configuration option for this seems to be an overkill to me.

dbu:
the final call is to @GuilhemN anyways ;-)

on the topic of customizing the template, one option could be to just define many blocks that can be overwritten, so that users can extend the template and redefine the blocks they want. one block could be "external_fonts" and then one can extend the template and overwrite just that block to be empty.

the only configuration needed for this should be one to tell what twig template to use. afaik atm its hardcoded - and with the symfony template overwrite, you can't extend the other template as symfony is "replacing" one file with the other.

markuspoerschke:
@dbu That can be solved by creating a "base" template and some inheritance. Give me some time to open a new PR for that. :)

GuilhemN:
About using Google fonts I have no strong opinion on the solution to adopt so I'd prefer to wait and see what people respond on https://github.com/swagger-api/swagger-ui/pull/4598 to have an idea of pros and cons and maybe different solutions.

About having a base template it has been requested several times already so would be a great addition :)

markuspoerschke:
I close this PR and I will come up with another solution that will not remove the given fonts.

markuspoerschke:
swagger-api/swagger-ui#4598 was accepted, therefore I think this one could be merged as well.

(Sorry for the closing, opening the issue, I did not check the other issue before)

dbu:
makes sense to me to merge, then.

maybe you could add a doc section how to re-add the font when wanted? and this needs an entry in the changelog, along with the reason (GDPR) why this is done, and a mention of the doc that explains how to re-add if desired.

dbu:
with the changes in #1357 , all affected parts, notably the documentation, change. i suggest to just do that inside #1357 even if its not semantically related.

PR #1334, opened 2018-05-31
symfony/property-info 4.1.0 looks broken

goetas:


dbu:
indeed. now the question whether this is a regression of the symfony component, or whether we used some kind of unintended behaviour of the component. did you check the changes to the component between 4.0 and 4.1? are they large?

goetas:
Not really...

goetas:
see here the reason of the bug https://github.com/symfony/symfony/issues/27449#issuecomment-393851189

PR #1343, opened 2018-06-15
Create .gitattributes

Jean85:
This file reduces the stuff that gets exported when packaging this lib in the vendor folder through Composer, i.e. the test lib does not belong there.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+.*                  export-ignore
+phpunit*            export-ignore
+Tests               export-ignore


GuilhemN:
I think we should keep exporting them, as does symfony (see https://github.com/symfony/symfony/issues/6605), they are part of the documentation and not that heavy in our case.

GuilhemN:
Closing for the same reasons symfony did.
Thanks for the proposal still!

Issue #1330, opened 2018-05-24
[RFC] Allow to define parameters of swagger ui


magnetik:
Hey,

I'm interested in customizing some [parameters of swagger ui](https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md). I mainly want to hide the models by default (`defaultModelExpandDepth`).

Can we add some config that will be passed to `Resources/public/init-swagger-ui.js` to configure this ?

I think I can provide a PR :)

dbu:
in the past, i overwrote the twig template for this. but it sounds nice - alternatively we could add more twig blocks in the template so that you can extend the template and e.g. overwrite the swagger-ui initialization block with your configuration.

KiSchulte:
Especially the Model Section is annoying.

chrisguitarguy:
Related: 
- https://github.com/nelmio/NelmioApiDocBundle/issues/1238
- https://github.com/nelmio/NelmioApiDocBundle/issues/1097

DjordyKoert:
This is possible with the `html_config` config option 😄 

https://symfony.com/bundles/NelmioApiDocBundle/current/configuration_reference.html#html-config

PR #1358, opened 2018-07-11
[WIP] Allow to override template for swagger-ui

magnetik:
Following @dbu comment on #1330 

Not sure if it's the right path. Is it ?

To add a block to define parameters, I might be forced to inline `init-swagger-ui.js`. I'm not sure that it make sense to have it in a different file anyway.

### DIFF
--- a/Controller/SwaggerUiController.php
+++ b/Controller/SwaggerUiController.php
@@ -20,6 +20,8 @@ use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 
 final class SwaggerUiController
 {
+    private $view;
+
     private $generatorLocator;
 
     private $twig;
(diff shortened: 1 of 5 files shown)


Controller/SwaggerUiController.php
```diff
@@ -20,14 +20,16 @@
...
+
     private $generatorLocator;
 
     private $twig;
 
     /**
      * @param ContainerInterface $generatorLocator
      */
-    public function __construct($generatorLocator, \Twig_Environment $twig)
+    public function __construct($view, $generatorLocator, \Twig_Environment $twig)
```
dbu: this parameter is not used. i think you need to assign it.

and i'd append it at the end of the list and default to the default template, for BC in case somebody extended the controller or something.

you could use a constant on this class for the default template name, so you can use it both in the constructor and as the default value in the Configuration class.

magnetik:
Nice ! I have missed it but it looks perfect :)

PR #1360, opened 2018-07-12
CollectionType not working in example tab

imajim:
Removed just line 179 and collectionType now working in example tab

### DIFF
--- a/ModelDescriber/FormModelDescriber.php
+++ b/ModelDescriber/FormModelDescriber.php
@@ -91,10 +91,8 @@ final class FormModelDescriber implements ModelDescriberInterface, ModelRegistry
      *
      * @param FormConfigBuilderInterface $config
      * @param                            $property
-     *
-     * @return bool
      */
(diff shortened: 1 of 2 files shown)


GuilhemN:
Thanks for contributing!

You need to update the test suite to remove the occurrences of the `example` field removed. According to https://travis-ci.org/nelmio/NelmioApiDocBundle/jobs/402961433, Nelmio\ApiDocBundle\Tests\Functional\FunctionalTest::testFormSupport needs to be updated.

ModelDescriber/FormModelDescriber.php
```diff
@@ -175,7 +175,7 @@ private function findFormType(FormConfigBuilderInterface $config, $property): bo
                 $itemsProp = $property->getItems();
 
                 if (!$this->findFormType($subForm->getConfig(), $itemsProp)) {
```
GuilhemN: Please remove the if structure and only keep the condition

GuilhemN:
Thank you:

PR #1363, opened 2018-07-18
[WIP] [JMSModelDescriber] Infer integer indexed arrays as "array" instead of "object"

phansys:
|Q            |A     |
|---          |---   |
|Branch       |master|
|Bug fix?     |yes   |
|New feature? |no    |
|BC breaks?   |no    |
|Deprecations?|no    |
|Tests pass?  |yes   |
|Fixed tickets|n/a   |
|License      |MIT   |
|Doc PR       |n/a   |

### DIFF
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -93,13 +93,13 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
             }
 
             if ($nestedType = $this->getNestedTypeInArray($item)) {
-                list($type, $isHash) = $nestedType;
-                if ($isHash) {
+                list($type, $isObject) = $nestedType;
+                if ($isObject) {
                     $property->setType('object');
 
                     $typeDef = $this->findPropertyType($type, $groups);
 
-                    // in the case of a virtual property, set it as free object type
+                    // set it as object type
                     $property->merge(['additionalProperties' => $typeDef ?: []]);
 
                     continue;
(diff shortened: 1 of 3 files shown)


ModelDescriber/JMSModelDescriber.php
```diff
@@ -193,20 +193,16 @@ private function registerPropertyType(array $typeDef, $property)
...
-        if ('array' !== $item->type['name'] && 'ArrayCollection' !== $item->type['name']) {
+        if (('array' !== $item->type['name'] && 'ArrayCollection' !== $item->type['name']) || !isset($item->type['params'][0]['name'])) {
             return null;
         }
 
-        // array<string, MyNamespaceMyObject>
         if (isset($item->type['params'][1]['name'])) {
-            return [$item->type['params'][1]['name'], true];
+            // array<string, MyNamespaceMyObject>
+            return [$item->type['params'][1]['name'], 'integer' !== $item->type['params'][0]['name']];
```
dbu: are you sure? thing is, when the integers are not consecutive, the hashmap needs to be represented as object. only 0 indexed consecutive php arrays are arrays in the understanding of json and other programming languages. does jms remove "holes" and make the array 0-indexed in this case?

phansys: You're right. I'll do the test ASAP. I found the "issue" with consecutive integers case while I was playing with the serializer in order to get a valid OpenAPI definition for a property with nested arrays (I'll tell you about this in another issue).

dbu: i guess in many cases, its the right interpretation. the "problem" is that PHP does not really distinguish plain array vs hashmap, while every other programming language does. its convenient, but when building json it makes a difference. not sure if jms itself has anything to handle it explicitly, or if not if we should add something additional to this bundle?

ftr, the problem we are talking about is:
```
php > var_dump(json_encode([0 => 'a', 2 => 'b']));
string(17) "{"0":"a","2":"b"}"
php > var_dump(json_encode([0 => 'a', 1 => 'b']));
string(9) "["a","b"]"
```

phansys: I was able to confirm that for a case like `[0 => 'a', 2 => 'b']`, the notation `array<integer, string>` gives an object output (`{"0":"a","2":"b"}`), while `array<string>` reindex the keys, generating an array output (`[0 => 'a', 1 => 'b']`). So, I'll close this PR by now. Thank you very much.

PR #1365, opened 2018-07-20
[JMSModelDescriber] Update `JMSModelDescriber` in order to iterate over nested array definitions

### LINKED ISSUE #1364: [JMSModelDescriber] Serialization config for nested arrays generates invalid definitions
I need to generate documentation for a property which holds values like `[[0, 0.5], [0, 0.8]]`, so I have the following serializer configuration which works fine for serialization purposes:
```yml
# config/serializer/MyObject.yml
MyObject:
    exclusion_policy: ALL
    properties:
        data:
            expose: true
            inline: false
            type: array<array<float>>
```
The problem is that config doesn't generate a valid OpenAPI definition (note the missing "items" property for `MyObject.properties.data.items`):
```yml
definitions:
  MyObject:
    properties:
      data:
        items:
          type: array
        type: array
    type: object
```
It should look like:
```yml
definitions:
  MyObject:
    properties:
      data:
        items:
          items:
              type: number
          type: array
        type: array
    type: object
```
So, https://editor.swagger.io/# shows me the following error:
```
Semantic error at definitions.MyObject.properties.data.items
Schemas with 'type: array', require a sibling 'items: ' field
```
I was trying to play with `JMSModelDescriber::describe()` in order to let the describer to follow nested array definitions, but I couldn't reach my goal.
Even, as a workaround I also tried to override the definition for `MyObject` at `nelmio_api_doc.documentation.definitions.MyObject`, but this causes the original definition to be renamed to "MyObject2", keeping the invalid schema.
I'm able to create a WIP PR with some of my attempts if this report seems right. Thank you in advance.

phansys:
|Q            |A     |
|---          |---   |
|Branch       |master|
|Bug fix?     |yes   |
|New feature? |no    |
|BC breaks?   |no    |
|Deprecations?|no    |
|Tests pass?  |yes   |
|Fixed tickets|#1364 |
|License      |MIT   |
|Doc PR       |n/a   |

Solution based on @GuilhemN's suggestion (see https://github.com/nelmio/NelmioApiDocBundle/issues/1364#issuecomment-406713022).

**TODO:**
- [x] Check if is possible to use `PropertyMetadata` for nested items at `describeItem()` and `getNestedTypeInArray()`.

Closes #1364.

### DIFF
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -14,7 +14,6 @@ namespace Nelmio\ApiDocBundle\ModelDescriber;
 use Doctrine\Common\Annotations\Reader;
 use EXSyst\Component\Swagger\Schema;
 use JMS\Serializer\Exclusion\GroupsExclusionStrategy;
-use JMS\Serializer\Metadata\PropertyMetadata;
 use JMS\Serializer\Naming\PropertyNamingStrategyInterface;
 use JMS\Serializer\SerializationContext;
 use Metadata\MetadataFactoryInterface;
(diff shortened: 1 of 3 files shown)


ModelDescriber/JMSModelDescriber.php
```diff
@@ -186,25 +160,39 @@ private function registerPropertyType(array $typeDef, $property)
         }
     }
 
-    /**
-     * @param PropertyMetadata $item
-     *
-     * @return array|null
-     */
-    private function getNestedTypeInArray(PropertyMetadata $item)
+    private function describeItem(array $type, Schema $property, array $groups = null): void
```
GuilhemN: Please remove the return type, it makes the tests fail :)

phansys:
Since `PropertyMetadata` is intended to be used as a representation for a class property, I think there is no chance for using it with standalone array values.

GuilhemN:
Thank you @phansys!

PR #1366, opened 2018-07-25
Jms serializer groups can be nested

goetas:
jms/serializer can be nested, not just plain strings


GuilhemN:
That's right, thank you!

PR #1367, opened 2018-07-25
Jms serializer default groups should not generate new definitions when possible

goetas:
jms/serializer default groups should not generate new definitions when possible.

This PR handles two cases
- When the default group is used
- When nested groups are used but on other properties that should not influence the current object

### DIFF
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -76,7 +76,7 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
                 $groups = array_filter($groups, 'is_scalar');
             }
 
-            if ([GroupsExclusionStrategy::DEFAULT_GROUP] == $groups) {
+            if ([GroupsExclusionStrategy::DEFAULT_GROUP] === $groups) {
                 $groups = null;
             }
 
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -76,6 +76,10 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
                 $groups = array_filter($groups, 'is_scalar');
             }
 
+            if ([GroupsExclusionStrategy::DEFAULT_GROUP] == $groups) {
+                $groups = null;
+            }
+
             // read property options from Swagger Property annotation if it exists
             if (null !== $item->reflection) {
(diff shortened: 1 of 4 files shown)


ModelDescriber/JMSModelDescriber.php
```diff
@@ -72,6 +72,12 @@ public function describe(Model $model, Schema $schema)
             $groups = $model->getGroups();
             if (isset($groups[$name]) && is_array($groups[$name])) {
                 $groups = $model->getGroups()[$name];
+            } elseif (is_array($groups)) {
+                $groups = array_filter($groups, 'is_scalar');
+            }
+
+            if ([GroupsExclusionStrategy::DEFAULT_GROUP] == $groups) {
```
phansys: Strict comparison could be used.

ModelDescriber/JMSModelDescriber.php
```diff
@@ -72,6 +72,12 @@ public function describe(Model $model, Schema $schema)
             $groups = $model->getGroups();
             if (isset($groups[$name]) && is_array($groups[$name])) {
                 $groups = $model->getGroups()[$name];
+            } elseif (is_array($groups)) {
+                $groups = array_filter($groups, 'is_scalar');
```
goetas: this is used to remove nested group definitions if the current property is not matched by any of them.

(equivalent to this code in the serializer lib https://github.com/schmittjoh/serializer/blob/942b566c801b2c567193f006ebf95b027e5bd439/src/JMS/Serializer/Exclusion/GroupsExclusionStrategy.php#L85)

ModelDescriber/JMSModelDescriber.php
```diff
@@ -72,6 +72,12 @@ public function describe(Model $model, Schema $schema)
...
+            }
+
+            if ([GroupsExclusionStrategy::DEFAULT_GROUP] == $groups) {
+                $groups = null;
```
goetas: equivalent to https://github.com/schmittjoh/serializer/blob/942b566c801b2c567193f006ebf95b027e5bd439/src/JMS/Serializer/Exclusion/GroupsExclusionStrategy.php#L19

ModelDescriber/JMSModelDescriber.php
goetas: fixed

GuilhemN:
Thanks @goetas!

goetas:
thanks to you! now I can complete https://github.com/nelmio/NelmioApiDocBundle/pull/1311 :)

PR #1368, opened 2018-07-25
[JMSModelDescriber] [WIP] Allow to use free form objects from JMS serialization

phansys:
|Q            |A     |
|---          |---   |
|Branch       |master|
|Bug fix?     |no    |
|New feature? |no    |
|BC breaks?   |no    |
|Deprecations?|no    |
|Tests pass?  |yes   |
|Fixed tickets|n/a   |
|License      |MIT   |
|Doc PR       |n/a   |

Depends on https://github.com/GuilhemN/swagger/pull/10.

**TODO**
- [ ] Confirm with reviewers if `"additionalProperties": true` will be accepted;
- [ ] Squash commits.

### DIFF
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -165,11 +165,17 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
         }
     }
 
+    /**
+     * @param array      $type
+     * @param Schema     $property
+     * @param array|null $groups
+     */
     private function describeItem(array $type, $property, array $groups = null)
     {
         if (list($nestedType, $isHash) = $this->getNestedTypeInArray($type)) { // @ todo update a bit getNestedTypeInArray and describe ($type = $item->type)
             if ($isHash) {
                 $property->setType('object');
+                $additionalProperties = $property->getAdditionalProperties();
                 // in the case of a virtual property, set it as free object type
                 $property->merge(['additionalProperties' => []]);
 
@@ -183,6 +189,10 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
         }
 
         if ($typeDef = $this->findPropertyType($type['name'], $groups)) {
+            if (isset($typeDef['type']) && 'array' === $typeDef['type'] && null === $property->getItems()->toArray()) {
+                unset($typeDef['type']);
+            }
+
             $this->registerPropertyType($typeDef, $property);
         }
     }
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -173,9 +173,7 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
                 // in the case of a virtual property, set it as free object type
                 $property->merge(['additionalProperties' => []]);
 
-                if ('array' !== $nestedType['name']) {
-                    $this->describeItem($nestedType, $property->getAdditionalProperties(), $groups);
-                }
+                $this->describeItem($nestedType, $property->getAdditionalProperties(), $groups);
 
                 return;
             }
@@ -185,6 +183,10 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
         }
 
         if ($typeDef = $this->findPropertyType($type['name'], $groups)) {
+            if (isset($typeDef['type']) && 'array' === $typeDef['type'] && null === $property->getItems()->toArray()) {
+                unset($typeDef['type']);
+            }
+
             $this->registerPropertyType($typeDef, $property);
         }
     }
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -165,17 +165,11 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
         }
     }
 
-    /**
-     * @param array      $type
-     * @param Schema     $property
-     * @param array|null $groups
-     */
-    private function describeItem(array $type, $property, array $groups = null)
+    private function describeItem(array $type, Schema $property, array $groups = null)
     {
         if (list($nestedType, $isHash) = $this->getNestedTypeInArray($type)) { // @ todo update a bit getNestedTypeInArray and describe ($type = $item->type)
             if ($isHash) {
                 $property->setType('object');
-                $additionalProperties = $property->getAdditionalProperties();
                 // in the case of a virtual property, set it as free object type
                 $property->merge(['additionalProperties' => []]);
 
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -173,7 +173,9 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
                 // in the case of a virtual property, set it as free object type
                 $property->merge(['additionalProperties' => []]);
 
-                $this->describeItem($nestedType, $property->getAdditionalProperties(), $groups);
+                if ('array' !== $nestedType['name']) {
+                    $this->describeItem($nestedType, $property->getAdditionalProperties(), $groups);
+                }
 
                 return;
             }
@@ -183,10 +185,6 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
         }
 
         if ($typeDef = $this->findPropertyType($type['name'], $groups)) {
-            if (isset($typeDef['type']) && 'array' === $typeDef['type'] && null === $property->getItems()->toArray()) {
-                unset($typeDef['type']);
-            }
-
             $this->registerPropertyType($typeDef, $property);
         }
     }
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -173,8 +173,12 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
                 // in the case of a virtual property, set it as free object type
                 $property->merge(['additionalProperties' => []]);
 
-                $this->describeItem($nestedType, $property->getAdditionalProperties(), $groups);
+                // this is a free form object (as nested array)
+                if ('array' === $nestedType['name'] && !isset($nestedType['params'][0])) {
+                    return;
+                }
 
+                $this->describeItem($nestedType, $property->getAdditionalProperties(), $groups);
                 return;
             }
 
@@ -183,10 +187,6 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
         }
 
         if ($typeDef = $this->findPropertyType($type['name'], $groups)) {
-            if (isset($typeDef['type']) && 'array' === $typeDef['type'] && null === $property->getItems()->toArray()) {
-                unset($typeDef['type']);
-            }
-
             $this->registerPropertyType($typeDef, $property);
         }
     }
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -138,9 +138,7 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
 
             $property->setType('array');
             $this->describeItem($nestedType, $property->getItems(), $groups);
-        }
-
-        if ('array' === $type['name'] && null === $property->getItems()->toArray()) {
+        } elseif ('array' === $type['name']) {
             $property->setType('object');
             $property->merge(['additionalProperties' => []]);
         } elseif (in_array($type['name'], ['boolean', 'string', 'array'], true)) {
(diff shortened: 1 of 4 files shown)


Tests/Functional/Entity/JMSUser.php
```diff
@@ -132,6 +132,12 @@ class JMSUser
      */
     private $latLonHistory;
 
+    /**
+     * @Serializer\Type("array<string, array>")
```
goetas: this is really not recommended to be used in jms/serializer

mixed type arrays are not really supported. the resulting serialized string is purely fully managed by json_encode and not by jms serializer anymore

phansys: How do you suggest to represent [free form objects](https://swagger.io/docs/specification/data-models/dictionaries/#free-form) within the serialization?

goetas: this trick works, i'm not saying that. 
im saying that is just risky to use (and the data contained in the values of the array are not in the control of the serializer anymore), but from the swagger doc is perfectly valid.

GuilhemN:
I'm not sure we need this change here, it will work fine as well with `[]`, right ?

About @goetas comment, using this trick is also risky on our side because it might break when we will migrate to OpenApi 3.0, I would rather suggest using `@SWG\Property(additionalProperties=[])`

phansys:
About the valid schema, `{}` or `true` should have the same meaning. I tried with `true` in order to avoid `[]` to be exposed. `definitions.Pet.properties.data.additionalProperties.type` should not be included:
```json
{
    "definitions":
        "Pet": {
            "type": "object",
            "properties":
                "data": {
                    "additionalProperties": {
                        "type": "array"
                    },
                    "type": "object"
                }
            }
        }
    }
}
```

phansys:
Now that the changes at `exsyst/swagger` were released under `0.4.1` the build is passing.
IIUC, the definition for Open API 3.0 supports `"additionalProperties": true` in the same way as `"additionalProperties": {}`: https://swagger.io/docs/specification/data-models/dictionaries/. If you think we must use `{}`, I can make another PR to update [`Schema`](https://github.com/GuilhemN/swagger/blob/2802550e131777e39c8709640c918cfd1dc9b996/src/Schema.php#L104) again.

Thank you in advance!

ModelDescriber/JMSModelDescriber.php
```diff
@@ -165,11 +165,17 @@ private function registerPropertyType(array $typeDef, $property)
...
+     * @param array      $type
+     * @param Schema     $property
+     * @param array|null $groups
+     */
     private function describeItem(array $type, $property, array $groups = null)
     {
         if (list($nestedType, $isHash) = $this->getNestedTypeInArray($type)) { // @ todo update a bit getNestedTypeInArray and describe ($type = $item->type)
             if ($isHash) {
                 $property->setType('object');
+                $additionalProperties = $property->getAdditionalProperties();
```
goetas: is this some new change in swagger? I remember that when I have implemented this feature, `$additionalProperties` was NULL if not set via `merge....`

goetas: and this var looks never used

ModelDescriber/JMSModelDescriber.php
```diff
@@ -183,6 +189,10 @@ private function describeItem(array $type, $property, array $groups = null)
         }
 
         if ($typeDef = $this->findPropertyType($type['name'], $groups)) {
+            if (isset($typeDef['type']) && 'array' === $typeDef['type'] && null === $property->getItems()->toArray()) {
```
goetas: actually this check can be moved just few lines before, after line 177. if the condition is satisfied just `return`. the implementation should be simpler and you do not need to check `$property->getItems()` or to unset type

Tests/Functional/Entity/JMSUser.php
goetas: the test will have to be changed to `additionalProperties=[]`

phansys: If I change the test, it will fail since[`Schema`](https://github.com/GuilhemN/swagger/blob/a02984db5edacdce2b4e09dae5ba8fe17a0e449e/tests/PetstoreTest.php#L130) is currently using `true` instead of `[]`. Are you proposing to modify that?

ModelDescriber/JMSModelDescriber.php
GuilhemN: What's the point of this change ? The definition is still invalid with this, right ?

phansys: No, the definition is fine with this change since it generates `"additionalProperties": true` when the type is "array" with no defined items.

goetas:
Here (https://github.com/phansys/NelmioApiDocBundle/pull/1) the "simplification" I was trying to explain in https://github.com/nelmio/NelmioApiDocBundle/pull/1368#discussion_r205781418

goetas:
Style CI complains (my fault :smiley: )

phansys:
After merging #1, Travis is failing.

phansys:
I've added a new test for definitions like `@Type("array")`, since properties declared simply as "array" (without specifying item types) should be treated as free form objects too, given the describer doesn't know how to populate the "items" property, which is required for "array" types. The problem is I don't know how to achieve the desired result from `JMSModelDescriber`, so any hint is welcome :grin:.

GuilhemN:
This case is managed [here](https://github.com/phansys/NelmioApiDocBundle/blob/6ae52346f4c0b33db8aceac3dfb47cc82de31159/ModelDescriber/JMSModelDescriber.php#L130). So `array` management should be done in a separate block and to simplify things, I would merge `findPropertyType` and `registerPropertyType`

phansys:
The requested changes are almost ready, just the "additionalProperties" definition is missing.

phansys:
Tests are green now. I don't like to call `Schema::merge()` from `registerPropertyType()` and `describeItem()`; but I didn't find an alternative to avoid this with the current approach.

goetas:
:-1: for the latest changes , will try to have a look at in the next days

ModelDescriber/JMSModelDescriber.php
```diff
@@ -118,61 +118,50 @@ public function supports(Model $model): bool
...
-        $typeDef = [];
-        if (in_array($type, ['boolean', 'string', 'array'])) {
-            $typeDef['type'] = $type;
-        } elseif (in_array($type, ['int', 'integer'])) {
-            $typeDef['type'] = 'integer';
-        } elseif (in_array($type, ['double', 'float'])) {
-            $typeDef['type'] = 'number';
-            $typeDef['format'] = $type;
+        if ('array' === $type) {
+            if (null === $property->getItems()->toArray()) {
```
GuilhemN: I'd rather add an `else` [here](https://github.com/nelmio/NelmioApiDocBundle/pull/1368/files#diff-2f796c92a7c91a1dd87dabc9cd030aa9R174) instead of this check.

ModelDescriber/JMSModelDescriber.php
```diff
@@ -118,61 +118,50 @@ public function supports(Model $model): bool
...
     }
 
-    /**
-     * @param string     $type
-     * @param array|null $groups
-     *
-     * @return array|null
-     */
-    private function findPropertyType(string $type, array $groups = null)
+    private function registerPropertyType(Schema $property, string $type, array $groups = null)
```
GuilhemN: Sorry to bother but I'm not sure we need this function anymore. Wouldn't it be clearer if it was inlined in `describeItem`?

phansys:
`registerPropertyType()` was merged into `describeItem()`.

ModelDescriber/JMSModelDescriber.php
```diff
@@ -171,7 +140,29 @@ private function describeItem(array $type, Schema $property, array $groups = nul
             $this->describeItem($nestedType, $property->getItems(), $groups);
         }
 
-        $this->registerPropertyType($property, $type['name'], $groups);
+        if ('array' === $type['name'] && null === $property->getItems()->toArray()) {
```
GuilhemN: `elseif ('array' === $type['name'])` to fix my second comment :)

phansys: Do you mean something like this?
```php
if ('array' === $type['name']) {
    if (null !== $property->getItems()->toArray()) {
        $property->setType($type['name']);
    } else {
        $property->setType('object');
        $property->merge(['additionalProperties' => []]);
    }
} elseif (...
```

goetas:
:-1:  `describeItem` become much more complex, the initial version (the one after https://github.com/phansys/NelmioApiDocBundle/pull/1) was much simpler (and was working anyway).

`additionalProperties=[]` or `additionalProperties=true` is equivalent by definition, so the solution was IMO just to relax the test.

ModelDescriber/JMSModelDescriber.php
```diff
@@ -118,61 +118,19 @@ public function supports(Model $model): bool
...
+    private function describeItem(array $type, Schema $property, array $groups = null)
     {
         if (list($nestedType, $isHash) = $this->getNestedTypeInArray($type)) { // @ todo update a bit getNestedTypeInArray and describe ($type = $item->type)
             if ($isHash) {
                 $property->setType('object');
                 // in the case of a virtual property, set it as free object type
                 $property->merge(['additionalProperties' => []]);
 
+                // this is a free form object (as nested array)
+                if ('array' === $nestedType['name'] && !isset($nestedType['params'][0])) {
```
GuilhemN: This seems to be dead code to me, isn't it?

phansys:
The version after phansys#1 is not working because these changes (which produces `additionalProperties=true`): https://github.com/GuilhemN/swagger/blob/a02984db5edacdce2b4e09dae5ba8fe17a0e449e/src/Schema.php#L104

GuilhemN:
> Do you mean something like this?

No I mean using an elseif with only one condition :) the second one becomes useless with an elseif since items is described only if the first case is entered.

@goetas it looks simpler to me :) it only lists the different types supported and removes the need of using intermediate arrays.

ModelDescriber/JMSModelDescriber.php
phansys: No, it prevents nested free form objects without defined items to be described.

GuilhemN: But this case is already filtered in https://github.com/nelmio/NelmioApiDocBundle/blob/1b2422eed0e16e5833bf780e04328330e61eb156/ModelDescriber/JMSModelDescriber.php#L174, right ?

phansys:
>No I mean using an elseif with only one condition

Sorry, I think I don't get your point. If `$type['name']` is "array" we must also decide which type to set based on the array items (if they are empty or not).

ModelDescriber/JMSModelDescriber.php
phansys: At this point, that check is performed over `$type`, not over `$nestedType`.

GuilhemN:
Sure but in fact if items is not empty, we know for sure that the first if was entered.

ModelDescriber/JMSModelDescriber.php
GuilhemN: Oops indeed, I remember why I didn't comment this earlier :smile:

phansys:
In order to have a clear understanding about what are you proposing, I need to know what you're considering "the first if".
In case you're referencing this
https://github.com/nelmio/NelmioApiDocBundle/blob/1b2422eed0e16e5833bf780e04328330e61eb156/ModelDescriber/JMSModelDescriber.php#L123
it isn't possible, remeber that this method has a recursive call inside that `if` statement:
https://github.com/nelmio/NelmioApiDocBundle/blob/1b2422eed0e16e5833bf780e04328330e61eb156/ModelDescriber/JMSModelDescriber.php#L134

GuilhemN:
@phansys yes the "first if" refers to https://github.com/nelmio/NelmioApiDocBundle/blob/1b2422eed0e16e5833bf780e04328330e61eb156/ModelDescriber/JMSModelDescriber.php#L123.

> it isn't possible, remeber that this method has a recursive call inside that if statement:

Precisely, this call fills `items` and makes `null !== $items` true. Actually I see only one case where my change would modify the behavior, it is when using a type name managed by a custom handler (and that's not a class), then additionalProperties will be used, but I think it's an unexpected edge case and if we truly want it, I'd prefer it being explicit.

ModelDescriber/JMSModelDescriber.php
```diff
@@ -118,72 +118,48 @@ public function supports(Model $model): bool
...
+
                 $this->describeItem($nestedType, $property->getAdditionalProperties(), $groups);
 
                 return;
             }
 
             $property->setType('array');
             $this->describeItem($nestedType, $property->getItems(), $groups);
-        }
+        } elseif ('array' === $type['name']) {
```
goetas: few lines below you have also `array`, you sure this if is ok?

GuilhemN: Yes, the `array` should be removed in fact :)

phansys:
These changes are fine for me, thank you so much!

GuilhemN:
Thank you! You're the one who made this possible :)

PR #1377, opened 2018-08-11
Added support for Doctrine MongoDB ODM 'document' form type

robholmes:
Adds support for the Doctrine MongoDB ODM `document` form type, which like `entity` extends `ChoiceType` and can be a multiple.

Also added test assertions for both `entity` and `document` with multiple enabled.

### DIFF
--- a/ModelDescriber/FormModelDescriber.php
+++ b/ModelDescriber/FormModelDescriber.php
@@ -177,25 +177,17 @@ final class FormModelDescriber implements ModelDescriberInterface, ModelRegistry
                 break;
             }
 
-            if ('entity' === $blockPrefix) {
+            // The DocumentType is bundled with the DoctrineMongoDBBundle
+            if ('entity' === $blockPrefix || 'document' === $blockPrefix) {
                 $entityClass = $config->getOption('class');
 
                 if ($config->getOption('multiple')) {
                     $property->setFormat(sprintf('[%s id]', $entityClass));
                     $property->setType('array');
+                    $property->getItems()->setType('string');
                 } else {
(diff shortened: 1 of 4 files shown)


ModelDescriber/FormModelDescriber.php
```diff
@@ -190,6 +190,16 @@ private function findFormType(FormConfigBuilderInterface $config, $property)
 
                 break;
             }
+
+            if ('document' === $blockPrefix) {
```
GuilhemN: Could you please add a note specifying this form type is bundled in https://github.com/doctrine/DoctrineMongoDBBundle please ? 
It wasn't obvious to me :)

ModelDescriber/FormModelDescriber.php
```diff
@@ -190,6 +190,16 @@ private function findFormType(FormConfigBuilderInterface $config, $property)
...
+                if ($config->getOption('multiple')) {
+                    $property->setType('array');
```
GuilhemN: That's invalid in fact (also for entity), we should add `$property->getItems()->setType('string');`

GuilhemN:
Thank you @robholmes!

bobvandevijver:
Does this mean that the release is coming? 😄

GuilhemN:
I'll fix a few more issues but then yes 😅

robholmes:
@GuilhemN no problem at all, glad I could help, and thank you for making the changes before I had a chance to 👍

GuilhemN:
3.3 is released :) See https://github.com/nelmio/NelmioApiDocBundle/releases/tag/v3.3.0

PR #1387, opened 2018-08-28
master: RepeatedType added

eaunitrust:


ModelDescriber/FormModelDescriber.php
```diff
@@ -164,6 +164,14 @@ private function findFormType(FormConfigBuilderInterface $config, $property)
                 break;
             }
 
+            if ('repeated' === $blockPrefix) {
+                $subType = $config->getOption('type');
+
+                $property->setType($subType);
```
GuilhemN: This looks wrong to me: if I understood correctly how this form type works, it adds two sub fields. So we should instantiate these two sub fields (like done for collections) and document them separately as object properties:
```php
$prop->setType('object');
$this->findFormType(/*first sub form type*/... , $prop->getProperties()->get($typeOptions['first_name']));

// 2nd sub form type
```

Btw could you add a test case for this change please?

GuilhemN:
Thank you @eaunitrust!

PR #1396, opened 2018-09-11
[Bug] Respect property name for constraints

scaytrase:
Currenly `@NotNull` or `@NotBlank` constraints utilize reflection property name, which is incorrect for cases where naming strategy is involved.

- [x] Created test (+model and controller) to reflect the bug present
- [x] Fixed the test


scaytrase:
@GuilhemN @dbu ?

GuilhemN:
Thank you @scaytrase!

PR #1402, opened 2018-09-24
Restore jms nested groups feature

goetas:
As said in https://github.com/nelmio/NelmioApiDocBundle/pull/1311#issuecomment-423981494 ,  #1384 broke the jms nested group feature implemented with https://github.com/nelmio/NelmioApiDocBundle/pull/1311.

This PR re integrates it. To do so was necessary to allow to not auto-register aliases. This is mainly because of how the nested groups work, is fundamental to know which was the previous group.

This PR fixes also few issues in the merge commit https://github.com/nelmio/NelmioApiDocBundle/commit/63cee64e6b02ef6f3685ee4a568a81444936d2fe

I'm open to hear better proposal since `reserve_name` to me sounds as a workaround for disabling  https://github.com/nelmio/NelmioApiDocBundle/pull/1384, otherwise will be impossible to achieve this feature.

### DIFF
--- a/Model/ModelRegistry.php
+++ b/Model/ModelRegistry.php
@@ -40,32 +40,30 @@ final class ModelRegistry
     {
         $this->modelDescribers = $modelDescribers;
         $this->api = $api;
-        $this->alternativeNames = array_reverse($alternativeNames); // last rule wins
+        $this->alternativeNames = []; // last rule wins
+
+        foreach (array_reverse($alternativeNames) as $alternativeName => $criteria) {
+            $this->alternativeNames[] = $model = new Model(new Type('object', false, $criteria['type']), $criteria['groups']);
+            $this->names[$model->getHash()] = $alternativeName;
+            $this->api->getDefinitions()->get($alternativeName);
+        }
     }
 
     public function register(Model $model): string
     {
-        return $this->doRegister($model);
-    }
-
-    /**
-     * Private method allowing to enforce the model name for alternative names.
-     */
-    private function doRegister(Model $model, string $name = null)
-    {
         $hash = $model->getHash();
-        if (isset($this->names[$hash])) {
-            return '#/definitions/'.$this->names[$hash];
+        if (!isset($this->models[$hash])) {
+            $this->models[$hash] = $model;
+            $this->unregistered[] = $hash;
+        }
+        if (!isset($this->names[$hash])) {
+            $this->names[$hash] = $this->generateModelName($model);
         }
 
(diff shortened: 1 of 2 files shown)


goetas:
Instead of registering the types in advance, does it make sense if a non matching alias exist when assigning the name?

Model/ModelRegistry.php
```diff
@@ -43,7 +43,9 @@ public function __construct($modelDescribers, Swagger $api, array $alternativeNa
         $this->alternativeNames = array_reverse($alternativeNames); // last rule wins
 
         foreach ($this->alternativeNames as $alternativeName => $criteria) {
-            $this->doRegister(new Model(new Type('object', false, $criteria['type']), $criteria['groups']), $alternativeName);
+            if (isset($criteria['reserve_name']) && true === $criteria['reserve_name']) {
```
GuilhemN: I think we should reserve it if the field doesn't exist:
`!isset($criteria['reserve_name']) || $criteria['reserve_name']`

GuilhemN:
What about instead reserving the name but actually registering them much later (when `registerModels` is called) so that they're processed last?

goetas:
@GuilhemN The latest implementation does not pre-register models but makes sure that a name is assigned only to the model that has the same signature as the alias.

Let me know what do you think

GuilhemN:
I'm not fund of it, I think it's nice to have them autoregistered to be able to just reference them. Why not adding them at the start of `registerModels`?

goetas:
> I'm not fund of it, I think it's nice to have them autoregistered to be able to just reference them. Why not adding them at the start of registerModels?

The problem is that the traversal order is very important for the nested groups (see https://github.com/nelmio/NelmioApiDocBundle/blob/master/ModelDescriber/JMSModelDescriber.php#L36). If the registration is done, even in registerModels, its order will be different and so the names.

GuilhemN:
I pushed a commit on your branch to show you how I see it :) (I tried to open a PR on your repository but it seems buggy 😅).

goetas:
@GuilhemN your solution worked pretty nice! I have rebased it fixing CS failures.

goetas:
Would never immagine a solution as what you did! :) I'm not that yet into the logic behind the model registry  :)

goetas:
> I tried to open a PR on your repository but it seems buggy

yeah, the master branch of my fork is  still nelmio-apidoc 2.x

goetas:
Thanks to you! great work!

PR #1403, opened 2018-09-24
Add boolean enums support to forms

goetas:
This adds support for boolean enums. 

previous response was:

```php
[
  'type' => 'string',
  'enum' => [true, false],
]
```

The enum is in contradiction with the filed type. Now is:

```php
[
  'type' => 'boolean',
  'enum' => [true, false],
]
```


GuilhemN:
Thank you :)

PR #763, opened 2015-12-03
Configurable views

goetas:
This PR adds configurable views: 
Each view can be configured.
```yaml
# app/config/config.yml
nelmio_api_doc:
    views:
      premium:
        include: ["code", "platinum", "secret"]
        include_empty: true
      simple:
        include: ["core", "secret"]
        include_empty: false
```    

* `include` will include into the `premium` view all the endpoints that belongs to "code", "platinum", "secret" views.
* `include_empty` will include into the `premium` all the endpoints has no specified views

PR #1404, opened 2018-09-25
Separated documentation for areas

bifidokk:


goetas:
Very nice feature! Similar to what I did in the past for 2.x https://github.com/nelmio/NelmioApiDocBundle/pull/763

DependencyInjection/NelmioApiDocExtension.php
```diff
@@ -103,6 +104,17 @@ public function load(array $configs, ContainerBuilder $container)
                     new Reference('logger'),
                 ])
                 ->addTag(sprintf('nelmio_api_doc.describer.%s', $area), ['priority' => -200]);
+
+            $documentation = count($config['areas'][$area]['documentation']) ? $config['areas'][$area]['documentation'] : $config['documentation'];
```
GuilhemN: I think it would be better to always load the default config (`nelmio_api_doc.documentation`) and use this area config as a complement (with a lower priority like 990 to override global defaults)

DependencyInjection/NelmioApiDocExtension.php
```diff
@@ -151,9 +163,6 @@ public function load(array $configs, ContainerBuilder $container)
                     ]);
             }
         }
-
-        // Import the base configuration
-        $container->getDefinition('nelmio_api_doc.describers.config')->replaceArgument(0, $config['documentation']);
```
GuilhemN: Should be reverted imo

Resources/config/services.xml
```diff
@@ -26,7 +26,7 @@
         <service id="nelmio_api_doc.describers.config" class="Nelmio\ApiDocBundle\Describer\ExternalDocDescriber" public="false">
             <argument type="collection" />
 
-            <tag name="nelmio_api_doc.describer" priority="1000" />
+            <tag name="nelmio_api_doc.describer" priority="-1000" />
```
GuilhemN: Should be reverted

Routing/FilteredRouteCollectionBuilder.php
```diff
@@ -26,6 +26,7 @@ public function __construct(array $options = [])
             ->setDefaults([
                 'path_patterns' => [],
                 'host_patterns' => [],
+                'documentation' => [],
```
GuilhemN: This addition doesn't really make sense. Can't we instead remove the `documentation` field before passing the config to this class ?

Tests/DependencyInjection/NelmioApiDocExtensionTest.php
```diff
@@ -92,6 +92,9 @@ public function testMergesRootKeysFromMultipleConfigurations()
         $extension = new NelmioApiDocExtension();
         $extension->load([
             [
+                'areas' => [
+                    'default' => [],
```
GuilhemN: What's this change for ?

bifidokk: yep, these changes are not needed

GuilhemN:
Great, thank you @bifidokk!

Issue #1397, opened 2018-09-11
Version 3.3.0 Symfony Choice annotation bug


caciobanu:
Hello,

The version 3.3.0 of the bundle does not work when in the Symfony Choice annotation you define only the callback method name without specifying the class: https://symfony.com/doc/current/reference/constraints/Choice.html#supplying-the-choices-with-a-callback-function

Error message: **Warning: call_user_func() expects parameter 1 to be a valid callback, function '***' not found or invalid function name**

Mairu:
Can we have a new release with this fix please?

PR #1325, opened 2018-05-18
fix(SymfonyConstraintAnnotationReader): fixed enum guessing in Assert…

fpalluel:
Hello, 

In Symfony, when using ``` Assert\Choice``` with ```callback```parameter, choices are not rendered in swagger documentation.
This PR proposes a small fix ;-)

PR #1405, opened 2018-09-26
Fix choice annotation with callback method name without class

piotrantosik:
Fix https://github.com/nelmio/NelmioApiDocBundle/issues/1397 added in https://github.com/nelmio/NelmioApiDocBundle/pull/1325


HEKET313:
@GuilhemN when are you planning to release this?

PR #1408, opened 2018-10-02
Display the FOS query param pattern

martijnvankempen:
The requirements for the FOS query parameters aren't displayed in the UI. This feature adds the regex pattern to the documentation page. See attached screenshot.
image

### DIFF
--- a/RouteDescriber/FosRestDescriber.php
+++ b/RouteDescriber/FosRestDescriber.php
@@ -12,13 +12,13 @@
 namespace Nelmio\ApiDocBundle\RouteDescriber;
 
 use Doctrine\Common\Annotations\Reader;
+use EXSyst\Component\Swagger\Parameter;
 use EXSyst\Component\Swagger\Swagger;
 use FOS\RestBundle\Controller\Annotations\QueryParam;
 use FOS\RestBundle\Controller\Annotations\RequestParam;
 use Symfony\Component\Routing\Route;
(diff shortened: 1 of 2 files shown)


goetas:
@martijnvankempen  I guess this needs some tests :)

martijnvankempen:
@goetas Done!

goetas:
hmmm, looks a bit of extreme-unit testing (but of course this is my opinion)

calbrecht:
No offence @martijnvankempen, the pattern is already added to the parameter further down in 
lines 68 - 71, wouldn't it be much more useful if you try to find out how to enable the display of the pattern within swagger-ui? 
```php
                $pattern = $this->getPattern($annotation->requirements);
                if (null !== $pattern) {
                    $parameter->pattern = $pattern;
                }
```
A quick search on their github repo yielded this for me https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md
```
showCommonExtensions | Boolean=false. Controls the display of extensions (pattern, maxLength, minLength, maximum, minimum) fields and values for Parameters.
```
So the next step you could take is to find out which version the NelmioApiDocBundle is using and if there is a way to either get the swagger-ui to the newer version that supports it (if the current is to old), (and) (or) find the file where it is configured within NelmioApiDocBundle and change the configuration accordingly. 

That would be a much more cleaner solution and all different describers that support pattern would benefit from it, as well as there would be less code to maintain.

PR #1409, opened 2018-10-02
Add ability to use response objects which defined in configuration

paillechat:
We can define responses in configuration, but can't use it in while we describe schema.


GuilhemN:
Thank you @paillechat!

Issue #1278, opened 2018-03-30
Update zircote/swagger-php to version 3


AndriiChymyrys:
Could you please update zircote/swagger-php to 3 version because I and maybe someone else need some futures from there?

GuilhemN:
It is not released yet, and we need to update https://github.com/GuilhemN/swagger first, help welcome this is a huge task :)

Jean85:
[zircote/swagger-php v3](https://github.com/zircote/swagger-php/releases/tag/3.0.0) has been released!

lucacavallaro:
Any update on this?

GuilhemN:
https://github.com/GuilhemN/swagger still needs to be updated but I can't dedicate it the time that would be necessary, contributions still welcome ;)

calbrecht:
To get into the topic i cloned the NelmioApiDocBundle and started poking around. Eventually this led me to the conclusion it would be better not to have the swagger-php functionality duplicated in swagger and therefore having to maintain a layer in between that needs to be updated before the update of the swagger-php version.

Now after i have removed the dependency to swagger by replacing its functionality with static methods in a Util class and adjusting the calls from within NelmioApiDocBundle classes, making all tests pass again and cleaning up here and there some bits, maybe you are willing to have a look at it?

https://github.com/mayflower/NelmioApiDocBundle/tree/refactor-swagger

I like to continue testing the rest of introduced functionality that isn't tested yet and then open a pull request here to get the changes merged, to be able to discuss the next refactoring on that basis, that would be to make the NelmioApiDocBundle work with the newer version of swagger-php.

GuilhemN:
That's a very interesting approach! I'm a bit worried about releasing a new major (3.x is not old...), but I guess it's worth it and I will most likely maintain 3.x for some time anyway.

I look forward to learning more about your proposal ☺

goetas:
If the update path from 3.x to 4.x is simple and clear, probably a new major release should be not a problem

goetas:
Checked now the changes... a lot of renames :(

calbrecht:
@goetas Yes, because that is what they are, the methods are named like *Definition, *Property.
The actual parameters to this methods are in fact instances of those Annotations named like so, why not name the variables to what they are?

goetas:
Naming is one of the most common BC breaks in most of the libraries... sadly they do not add any advantage to the library itself. But this is how life it goes :)

calbrecht:
[...]

PR #1416, opened 2018-10-13
Replace swagger dependency with static Util class

calbrecht:
This commit replaces the dependency to swagger.

The expected benefits from this refactoring are:
* Updates from the dependency swagger-php are easier to follow,
  because there is no external layer in between NelmioApiDocBundle
  and swagger-php that has to be maintained, see #1278.
  As the fundamental design of swagger-php does not change (seen
  in swagger-php v2 vs. swagger-php v3) and the static Util class is
  quite agnostic to the actual Annotation classes in swagger-php
  (beside Util::getNesting and convenience getter methods, that are
  easy to adapt), the effort is expected to be lower than keeping
  swagger up to date.
* NelmioApiDocBundle is expected to be less error prone, because it
  generates the Annotation objects of swagger-php directly. That
  removes the possibility of introducing bugs through the mapping with
  swagger and contributors will probably notice incompatibilities to
  swagger-php earlier. (examples beeing default=>null, type=>'string'
  values and int vs. string value types in test expectations and
  fixtures), also workarounds that existed because of the mapping
  layer as shown in SwaggerPhpDescriber (e.g. generating a new class
  and restricting ref resolution) are no longer needed.

For an explanation about the static Util class, see its own
documentation blocks.

Why the changes to:
* ./Controller/*
  - generate() no longer returns a swagger Instance but a swagger-php
    Swagger instance and there is no toArray method available.
* ./Describer/*
  - Change the classes to use swagger-php Anntoations and adapt to Util.
* ./Describer/ExternalDocDescriber
  - Save some cycles in case there is nothing to merge.
* ./Describer/SwaggerPhpDescriber
  - __constructor $overwrite parameter was never true in any existing
    test and has no efffect on the swagger annotation beside not
    filling up the _unmerged which is blacklisted and not serialized
    by Swagger->jsonSerialize anyway.
  - Removing AddDefaults from getProcessors because it duplicates
    functionality from DefaultDescriber and many test fail because of
    Swagger validation warning about existing properties (title or
    version) after refactoring.
    (Since i rely solely on the test suite i might be wrong as well,
    any hints appreciated.)
  - Removing the dynamic class extending Swagger because it is not
    needed any longer. Swagger is able to resolve references because
    all definitions, parameters, responses are present as nested
    childs of $api, i guess.
  - Removing method updateNestedAnnotations because all Annotations
    at this stage are nested already.
  - Rest of the changes is adapting to Util and cleanup for
    readability.
* ./Model/ModelRegistry
  - Adapt to Util
  - Method registerDefinitions
    - Add clarity through consistent naming.
    - The set($name) from the end of the loop is now handled
      in Util::getDefinition as well as the creation of the
      definition.
  - Method generateModelName
    - An existing name is now search through all api->definitions[]
      by its member api->definitions[]->definition.
  - Method typeToString
    - Deduplication of return 'mixed[]' and less nesting of control
      structures for readabilty.
* ./ModelDescriber/Annotations/*
  - Change the signature of updateDefinition and updateProperty to
    stricter typehinted classes, i guess the reason for the namimg
    was that swagger didn't have special classes for those but in
    swagger-php they exist as Annotation classes.
    (Since all the tests are passing i guess this did no harm.)
    Also this gives proper code completion within the IDE which
    cannot be accomplished in all cases by using the parent
    Schema class.
* ./ModelDescriber/Annotations/AnnotationsReader.php
  - Removed two members because they where not accessed anywhere but
    in the __constructor
* ./ModelDescriber/Annotations/SwgAnnotationsReader.php
  - Removed aliases in use statements because there is no conflict.
  - Changes from $property->merge to $property->mergeProperties
    because i could not get the tests green otherwise, so i think
    this is pretty safe and is doing the same as before.
* ./ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
  - setSchema added Schema type hint for strictness (maybe this can be
    set to Definition as well but the method name suggests Schema).
* ./ModelDescriber/FormModelDescriber.php
  - Method parseForm
    - again mergeProperties seems to do the job fine.
  - Method getBuiltinFormType
    - prevent duplicated call to $type->getBlockPrefix()
* ./ModelDescriber/JmsModelDescriber.php
  - Method supports
    - IDE suggested \ReflectionException is never thrown
  - Method describeItem
    - additionalProperties i'm very unsure if this is correct like
      it is now, but the tests are suggesting this is the way to go.
* ./ModelDescriber/ObjectModelDescriber.php
  - Added $type=$model->getType() to be reused in two calls.
* ./RouteDescriber/PhpDocDescriber.php
  - Nested control structure was not necessary.
* ./RouteDescriber/RouteDescriberTrait.php
  - \in_array enable opcode optimization,
    see https://github.com/Roave/FunctionFQNReplacer
* ./Tests/Functional/Controller/ApiController.php
  - All other response codes in tests are strings, and even
    the Response Annotation from swagger states in its doc
    block that response should be a string.
* ./Tests/Functional/*
  - All the models tested now contain a definition property,
    not sure if this is fine, but it is needed for indexing
    within the definitions collection anyway and it seems to
    do no harm otherwise.
* ./Tests/Functional/FunctionalTest.php
  - Method testApiPlatform
    - No need to assign
  - Method testUserModel
    - It is wanted that the declared default value
      is given even when it is null, i guess.
* ./Tests/Functional/JmsFunctionalTest.php
  - default => null, see above.
  - Method testModelComplexDocumentation
    - It is a good thing if a property of type integer
      is integer in fact.
* ./Tests/Functional/TestKernel.php
  - Need to define the type when generating swagger parameters
    from configuration. This will be troublesome, for some
    projects who rely on the lesser strictness.
* ./Tests/Functional/WebTestCase.php
  - Added assertion methods for convenience and easier testing.
* ./composer.json
  - zircote/swagger-php Parameter allowEmptyValue that is set within
    FosRestDescriber was first introduced in version 2.0.12
    https://github.com/zircote/swagger-php/releases/tag/2.0.12
  - jms/serializer because jms/serialzer-bundle uses it, and a
    bug in ExpressionPropertyMetadata that is hitting the
    VirtualProperty.php (don't understand why it did not show earlier)
    is fixed in version 1.13
    https://github.com/schmittjoh/serializer/releases/tag/1.13.0

BC issues:
  - Configuration with missing required properties is breaking now,
    e.g. ./Tests/Functional/TestKernel.php


calbrecht:
I did also check against a real life project https://github.com/etraxis/etraxis, which uses 3.3.0 of NelmioApiDocBundle, which i documented in https://gist.github.com/calbrecht/cb05e8968b341be87380c36dd0f74961 and the outcome is very promising in that there is only a persistent change within the definition objects, that seems negligible to me.

I'm happy to get questions and critique to the code, as i am eager to bring this into NelmioApiDocBundle. :)

calbrecht:
@GuilhemN to explain a bit more why i think it is good to actually replace the swagger dependency -- imho this had to be done much earlier with the introduction of swagger-php to this project. 
Given that swagger is much more convenient to work with when creating annotation objects and composing the whole swagger API description with it, it is also a duplication of concerns somehow since it does create the very similar structure like swagger-php but a little different.
It seems to me swagger is less strict in regards to the actual Swagger 2.0 specs than swagger-php which i think might be a source of confusion that leads to workarounds or bloated code to reverse the differences between swagger and swagger-php.

Model/ModelRegistry.php
```diff
@@ -142,14 +138,16 @@ private function typeToString(Type $type): string
...
-            } else {
-                return 'mixed[]';
             }
-        } else {
-            return $type->getBuiltinType();
+
+            return 'mixed[]';
         }
+
+        return $type->getBuiltinType();
```
GuilhemN: Please do not truly related changes in a different PR in a PR this big :) It makes reviewing slightly harder ;)

ModelDescriber/Annotations/SwgAnnotationsReader.php
```diff
@@ -35,36 +33,36 @@ public function __construct(Reader $annotationsReader, ModelRegistry $modelRegis
...
-        /** @var SwgDefinition $swgDefinition */
-        if (!$swgDefinition = $this->annotationsReader->getClassAnnotation($reflectionClass, SwgDefinition::class)) {
+        /** @var Definition $classDefinition */
+        if (!$classDefinition = $this->annotationsReader->getClassAnnotation($reflectionClass, Definition::class)) {
             return;
         }
 
         // Read @Model annotations
-        $this->modelRegister->__invoke(new Analysis([$swgDefinition]));
+        $this->modelRegister->__invoke(new Analysis([$classDefinition]));
```
GuilhemN: same, would be nicer in a different PR

ModelDescriber/JMSModelDescriber.php
```diff
@@ -120,59 +124,59 @@ public function supports(Model $model): bool
     {
         $className = $model->getType()->getClassName();
 
-        try {
-            if ($this->factory->getMetadataForClass($className)) {
-                return true;
-            }
-        } catch (\ReflectionException $e) {
```
GuilhemN: Why this change?

ModelDescriber/ObjectModelDescriber.php
```diff
@@ -37,19 +39,21 @@ public function __construct(
...
 
-    public function describe(Model $model, Schema $schema)
+    public function describe(Model $model, Definition $definition)
     {
-        $schema->setType('object');
-        $properties = $schema->getProperties();
+        $type = $model->getType();
+        $class = $type->getClassName();
+
+        $definition->type = $type->getBuiltinType();
```
GuilhemN: We're only describing objects, the `object` assignment was fine imo.

Tests/Functional/FunctionalTest.php
```diff
@@ -186,6 +164,7 @@ public function testUserModel()
                         'readOnly' => true,
                         'title' => 'userid',
                         'example' => 1,
+                        'default' => null,
```
GuilhemN: This is strange, do you know why this is added?

Tests/Functional/JMSFunctionalTest.php
```diff
@@ -63,6 +66,7 @@ public function testModelDocumentation()
                     'readOnly' => true,
                     'title' => 'userid',
                     'example' => 1,
+                    'default' => null,
```
GuilhemN: strange too

Tests/Functional/JMSFunctionalTest.php
```diff
@@ -216,10 +223,12 @@ public function testModelComplexDocumentation()
                     'title' => 'userid',
                     'description' => 'User id',
                     'readOnly' => true,
-                    'example' => '1',
+                    'example' => 1,
+                    'default' => null,
```
GuilhemN: same

GuilhemN:
Thanks for this PR and all the precise comments you added!
I agree this is better to focus on swagger-php (their classes are not really adapted for how we use them imo but it's still better than maintaining a whole other library). I think I will create a new branch for the next major to merge your PR.
Do you think it should only support Open Api 3.0 or both 2.0 and 3.0 ?

calbrecht:
Thanks for reviewing @GuilhemN, i will address this issues for sure -- it could be that it takes a little because i'm less frequent on my computer this day :)

ModelDescriber/ObjectModelDescriber.php
calbrecht: Good spot :) that slipped through. Changed and i will push when ready.

ModelDescriber/JMSModelDescriber.php
calbrecht: Unbelievable but true, this is a manifestation of the metadata serializer bug fixed in jms/serializer 1.13

If running the tests with < 1.13 and changing the ```\ReflectionExption``` to be a ```Exception``` i am able to break within the catch block. The change to ```Exception``` is needed because phpunit is catching the actual 'Undefined offset' warning that results from the unserialize call within jms/serializer that tries to assign to a list() construct and is missing the indexes because there are to few.

If running the tests with 1.13 the breakpoint within the catch block does not halt. So it is good to be removed, yes?

Model/ModelRegistry.php
calbrecht: Ok, promised :)

Tests/Functional/FunctionalTest.php
calbrecht: If you are too concerned about why (because of breaking things) i'll investigate this, but i'd rather blame swagger and keep it like so ;)

ModelDescriber/JMSModelDescriber.php
GuilhemN: Ok, thank for investigating :+1:

Tests/Functional/FunctionalTest.php
GuilhemN: Oh ok so it was most likely an issue on our side. No need to investigate more, your explanation makes sense :)

calbrecht:
> Thanks for this PR and all the precise comments you added!
Thanks for pointing the comments out, very much appreciated. :)

calbrecht:
This is ready from my side, i think. :)

GuilhemN:
I'll review your changes ASAP thanks :)

GuilhemN:
Great job, thank you! I merged your PR in `dev` to simplify future maintainance of two separate versions :)

PR #1419, opened 2018-10-15
Fix NelmioApiDocBundle view customization instructions

chteuchteu:
"Just create a file..." was not OK (missing views/ subdirectory, wrong SwaggerUi case)
The path showcased as code-block's first row was OK


GuilhemN:
You're right about the case but `views` is not needed, see https://symfony.com/doc/master/bundles/override.html#templates.

chteuchteu:
Alright! Should we completely strip it on both occurrences then?

GuilhemN:
Yes indeed, `views` should be here at all :+1:

chteuchteu:
@GuilhemN Done!

GuilhemN:
Thanks @chteuchteu! :)

PR #1420, opened 2018-10-17
faq.rst: enhance readability in symfony.com docs

chteuchteu:
## Issue
Here's how it's displayed [on symfony.com](https://symfony.com/doc/current/bundles/NelmioApiDocBundle/faq.html#asset-files-not-loaded): the last word of the first line gets printed on the next row, while the two next lines have a nice indentation
:
frequently asked questions -faq- - nelmio apidoc bundle docs- 10-17-2018 9-42-47 am

## Solution
After the fix, it will look like this:
frequently asked questions -faq- - nelmio apidoc bundle docs- 10-17-2018 9-47-08 am
No nice indentation under to emphasis the "A: " thing, but it looks a bit more readable

An alternative for that would be to ensure that the "or" doesn't get wrapped by fine-tuning on which word we wrap the line in the RST file, but it does not feel like the right thing to do

### DIFF
--- a/Resources/doc/faq.rst
+++ b/Resources/doc/faq.rst
@@ -45,7 +45,7 @@ Optional Path Parameters
 ------------------------
 
 Q: I have a controller with an optional path parameter. In swagger-ui, the parameter is required - can I make it
-   optional? The controller might look like this::
+optional? The controller might look like this::
 
     /**
      * Get all user meta or metadata for a specific field.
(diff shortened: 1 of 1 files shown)


GuilhemN:
Good catch! I think your solution is best, making sure we don't wrap the line on `or` looks more hacky.

Thanks :)

chteuchteu:
Thank you for merging this! :-)

PR #1421, opened 2018-10-17
[bugfix] jms serializer group names are on property names

goetas:
jms serializer group names are based on the internal property name, not on the serialized name


GuilhemN:
Thank you @goetas!

PR #1429, opened 2018-11-04
FIX #1428 Make sure FormInterfaceConfig has a 'documentation' option …

### LINKED ISSUE #1428: Unable to use @Model on a FormType adding SubmitType field
### The issue
Since migrating from 2.x to 3.0, I'm having an issue with my Form adding a submit button.
```
class MyType extends AbstractType
{
    /**
     * @param FormBuilderInterface $builder
     * @param array                $options
     */
    {
        $builder
            ->add('save', SubmitType::class)
        ;
    }
}
```
Defining the parameter like this
```
     * @SWG\Parameter(
     *     name="user",
     *     in="body",
     *     @Model(type=MyType::class)
     * ),
```
Leads to this error:

> {
    "code": 500,
    "message": "Argument 1 passed to EXSyst\\Component\\Swagger\\Schema::mergeDescription() must be of the type array, null given, called in C:\\Dev\\SupervisorHS\\vendor\\exsyst\\swagger\\src\\Schema.php on line 88"
}

### Where it comes from
It has to do with **FromModelDescriber.php**
https://github.com/nelmio/NelmioApiDocBundle/blob/67ff6bd1e11e070be202622ef908a65a0fd35271/ModelDescriber/FormModelDescriber.php#L78

```
    private function parseForm(Schema $schema, FormInterface $form)
    {
        $properties = $schema->getProperties();

        foreach ($form as $name => $child) {
            $config = $child->getConfig();
            $property = $properties->get($name);

            if ($config->getRequired()) {
                $required = $schema->getRequired() ?? [];
                $required[] = $name;

                $schema->setRequired($required);
            }

            $property->merge($config->getOption('documentation'));  <------------------- here
            if (null !== $property->getType()) {
                continue; // Type manually defined
            }

            $this->findFormType($config, $property);
        }
    }
```

It would be worth checking that there's a 'documentation' option
```
            if ($config->hasOption('documentation')) {
                $property->merge($config->getOption('documentation'));
            }
```

At least I don't get an error anymore, but the 'save' field is still appearing in the Model description.
What do you guys think?
I'm surprised I'm the first one to report this

neFAST:
…before getting it

FIX #1428

### DIFF
--- a/ModelDescriber/FormModelDescriber.php
+++ b/ModelDescriber/FormModelDescriber.php
@@ -74,8 +74,11 @@ final class FormModelDescriber implements ModelDescriberInterface, ModelRegistry
 
                 $schema->setRequired($required);
             }
-
-            $property->merge($config->getOption('documentation'));
+ 
+            if ($config->hasOption('documentation')) {
+                $property->merge($config->getOption('documentation'));
+            }
+ 
             if (null !== $property->getType()) {
                 continue; // Type manually defined
             }
--- a/ModelDescriber/FormModelDescriber.php
+++ b/ModelDescriber/FormModelDescriber.php
@@ -75,9 +75,7 @@ final class FormModelDescriber implements ModelDescriberInterface, ModelRegistry
                 $schema->setRequired($required);
             }
 
-            if ($config->hasOption('documentation')) {
-                $property->merge($config->getOption('documentation'));
-            }
+            $property->merge($config->getOption('documentation'));
             if (null !== $property->getType()) {
                 continue; // Type manually defined
             }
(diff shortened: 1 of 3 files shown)


neFAST:
@GuilhemN what do you think?

GuilhemN:
This is a strange issue that probably requires to dig a bit more, I'm not sure this is the right fix : this should not happen since `documentation` is supposed to default to `[]` (see https://github.com/nelmio/NelmioApiDocBundle/blob/master/Form/Extension/DocumentationExtension.php). Could you submit a failing test here please ?

neFAST:
Do you need more than my example in #1428 ?

neFAST:
@GuilhemN Ok so the tests do fail when I revert my commit, with the same error message that I'm getting in #1429

GuilhemN:
Ok, I found out where the issue comes from: not all form types inherits from `BaseType` (see https://symfony.com/doc/current/form/create_form_type_extension.html#generic-form-type-extensions). So I guess your solution is fine, sorry for the delay, we tend to forget about issues when we merge temp fixes without really understanding what's wrong.

GuilhemN:
Thanks a lot @neFAST !

PR #1430, opened 2018-11-08
add type hints

mmoll:


ApiDocGenerator.php
```diff
@@ -33,6 +33,8 @@
     /**
      * @param DescriberInterface[]|iterable      $describers
      * @param ModelDescriberInterface[]|iterable $modelDescribers
+     * @param CacheItemPoolInterface|null        $cacheItemPool
+     * @param string|null                        $cacheItemId
```
dbu: its a question of preference, but i am :-1: on doc comments that do not add any information. the type declaration in the signature is telling exactly the same thing. i switchted to only add param annotations if either the type can not be fully specified (array of specific type) or if there is any additional explanation for the parameter.

DependencyInjection/Configuration.php
```diff
@@ -16,7 +16,7 @@
 
 final class Configuration implements ConfigurationInterface
 {
-    public function getConfigTreeBuilder()
+    public function getConfigTreeBuilder(): TreeBuilder
```
dbu: as this implements [the interface](https://github.com/symfony/symfony/blob/8d277ce3e594c196594170e4fbbc37be98ba4f73/src/Symfony/Component/Config/Definition/ConfigurationInterface.php#L26) which does not have the return type declaration, i think we can't add that here. or is return type restriction allowed in php 7?

SwaggerPhp/Util.php
```diff
@@ -118,7 +118,7 @@ public static function getDefinition(Swagger $api, $definition): Definition
      * @see \Swagger\Annotations\Response::$schema
      * @see \Swagger\Annotations\Parameter::$schema
      *
-     * @param Response|Parameter $annotation
+     * @param AbstractAnnotation $annotation
```
dbu: are you sure that it can be other annotations than those 2?

Tests/Functional/Entity/SymfonyConstraints.php
```diff
@@ -84,79 +84,79 @@ class SymfonyConstraints
...
+    public function setPropertyChoice(int $propertyChoice)
     {
         $this->propertyChoice = $propertyChoice;
     }
 
     /**
      * @param int $propertyChoiceWithCallback
      */
-    public function setPropertyChoiceWithCallback(int $propertyChoiceWithCallback): void
+    public function setPropertyChoiceWithCallback(int $propertyChoiceWithCallback)
```
dbu: it looks like travis-ci is happy with these even when building with 7.0. this is quite weird because indeed the void return declaration has been introduced in 7.1

calbrecht:
@dbu hi, because you were mentioning styleci, i'd like to express my bad experience with this service here that i encountered when doing the swagger refactoring. 
It seemed very unreliable to me because it would not detect all issues on every run (i guess i fixed all warned places, just to find out in the next run there are more warnings). 
Then the styleci service gives no line numbers in the output, that is frustrating to find the correct places to fix. Also i felt like not being able to predict what the correct style would be. 

Is there a way to run the checks locally before submitting? 
And if not, would it be a welcome contribution to add the phpcs code sniffer with adequate ruleset to this project?

dbu:
@calbrecht its @GuilhemN call to decide if we want an additional phpcr sniffer. in an other project, we switched to pretty-ci which runs php-cs-sniffer with the same configuration that you can also use locally. that might be an option: https://github.com/php-http/HttplugBundle/pull/278

dbu:
> OK, but then I'd suggest to switch the minimun requirement to PHP 7.1 in the dev brnach, to get nullable return types

i am all in favor of that, but i don't see the direct connection. if we want to support PHP 7.0 its ok for me to keep the `@param|@return` for nullable parameters / return types. lets see what guilhelm thinks, if he likes the annotations i will shut up about them anyways ;-)

Form/Extension/DocumentationExtension.php
```diff
@@ -32,7 +32,10 @@ public function configureOptions(OptionsResolver $resolver)
             ->setAllowedTypes('documentation', ['array']);
     }
 
-    public function getExtendedType()
+    /**
+     * @return string
```
GuilhemN: Should be removed

Model/Model.php
```diff
@@ -20,6 +20,7 @@
     private $groups;
 
     /**
+     * @param Type          $type
```
GuilhemN: Same

GuilhemN:
I agree with @dbu I also prefer not adding unecessary annotations. It's also fine for me to move to 7.1, even though I'm not sure nullable types are worth it.

About StyleCI I think the issue is that we disabled `unalign_equals`, I guess we can add it back (needs to be checked maybe we aligned them somewhere). To fix cs locally you should install `php-cs-fixer` and run `php-cs-fixer fix` at the root of the repo.

Thanks for all these catches!

mmoll:
I opened GH-1434 against dev to drop PHP 7.0 and GH-1435 (dev) and GH-1436 (master) to fix the php-cs-fixer violations.

Model/Model.php
mmoll: I left that in, as the other parameter wouldn't be properly documented otherwise (`string[]|null` only translates to `?array`)

mmoll:
The current result of auto-fixing, please note that I didn't test ths with a real project...

PR #1439, opened 2018-11-16
Added TimeType field

LaurentMarquet:
Added TimeType as supported field.
I've also wished to modify the format used for date as it displays "string" and add the one for time, but I can't find where to do...

### DIFF
--- a/ModelDescriber/FormModelDescriber.php
+++ b/ModelDescriber/FormModelDescriber.php
@@ -132,6 +132,13 @@ final class FormModelDescriber implements ModelDescriberInterface, ModelRegistry
                 break;
             }
 
+            if ('time' === $blockPrefix) {
+                $property->setType('string');
+                $property->setFormat('time');
+
+                break;
+            }
+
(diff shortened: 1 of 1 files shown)


ModelDescriber/FormModelDescriber.php
```diff
@@ -132,6 +132,13 @@ private function findFormType(FormConfigBuilderInterface $config, $property)
                 break;
             }
 
+            if ('time' === $blockPrefix) {
+                $property->setType('string');
+                $property->setFormat('time');
```
GuilhemN: `datetime`? `time` is not an official format

Laurent3170: Maybe, I don't know. In the `date` blockPrefix (just above), it's specified 'date', that's why I've set 'time'. But I don't know neither where that formats are formatted. Date should display 'YYYY-MM-DD' and time 'H:i:s' or something like that, but `date` currently displays "string".

DjordyKoert:

PR #1446, opened 2018-12-06
Fix deprecation for symfony/config 4.2+

dirkkredler:
Fix the "A tree builder without a root node is deprecated since Symfony 4.2 and will not be supported anymore in 5.0" deprecation; same as https://github.com/doctrine/DoctrineBundle/pull/853


dirkkredler:
Already fixed :+1:

PR #1447, opened 2018-12-10
Change validation to allow fields described as DateTimeInterface.

AyrtonRicardo:
My models are described using interfaces, but when I tried to use "DateTimeInterface" wasn't working, only after change do DateTime or other custom DateTime format the documentation works.

The error:
```
The PropertyInfo component was not able to guess the type of DateTimeInterface::$timezone
```


GuilhemN:
Good catch, thank you @AyrtonRicardo!

PR #1456, opened 2018-12-27
Cache authorization in local storage.

cyberemissary:
Currently user has to provide authorization every time page is refreshed. During development this becomes extremely annoying as after every change you need to refresh the page and thus re-authorize.

This PR hooks into the swagger authorization mechanism to store the authorization in local storage and restore it on page reload (essentially providing the behavior of API Key storage like it was in V2)

### DIFF
--- a/Resources/public/init-swagger-ui.js
+++ b/Resources/public/init-swagger-ui.js
@@ -21,5 +21,30 @@ window.onload = () => {
     layout: 'StandaloneLayout'
   });
 
+  const storageKey = 'nelmio_api_auth';
+
+  // if we have auth in storage use it
+  if (sessionStorage.getItem(storageKey)) {
+    try {
+      ui.authActions.authorize(JSON.parse(sessionStorage.getItem(storageKey)));
+    } catch (ignored) {
+      // catch any errors here so it does not stop script execution
+    }
+  }
+
+  // hook into authorize to store the auth in local storage when user performs authorization
+  const currentAuthorize = ui.authActions.authorize;
+  ui.authActions.authorize = function (payload) {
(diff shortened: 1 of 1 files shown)


Resources/public/init-swagger-ui.js
```diff
@@ -21,5 +21,30 @@ window.onload = () => {
     layout: 'StandaloneLayout'
   });
 
+  const storageKey = 'nelmio_api_auth';
+
+  // if we have auth in storage use it
+  if (localStorage.getItem(storageKey)) {
```
GuilhemN: What do you think about using sessionstorage since this is quite sensible data?

cyberemissary: Sessions storage would definitely be more secure, but the only drawback with that is, session storage does not get shared across browser tabs (at least in chrome). So if you authorize in one tab then open a second tab, you will have to authorize again even if the first tab is still open.

I usually find myself working in a single tab with apidoc, so either storage engine would be fine for me personally. I believe most developers would be working with this on their private machines, which would make it fine to use `localStorage`.

It would be nice to hear some feedback from community on this.

In either case, let me know your decision. I'll change to `sessionStorage` if you say so.

GuilhemN: Sorry for the late answer, I haven't gotten much time to review PRs lately.
Despite the multiple tabs drawbacks, with no complaints about it (I think it is an acceptable limitation) I'd prefer to stick to using session storage

cyberemissary: Sorry took a while to changes this, I totally forgot about this PR. Changed to session storage.

GuilhemN:
Thank you @cyberemissary! :)

PR #1457, opened 2018-12-31
Fix: Deprecation: A tree builder without a root node is deprecated since symfony 4.2

mantis:


DependencyInjection/Configuration.php
```diff
@@ -18,9 +18,8 @@
 {
     public function getConfigTreeBuilder()
     {
-        $treeBuilder = new TreeBuilder();
-        $treeBuilder
-            ->root('nelmio_api_doc')
+        $treeBuilder = new TreeBuilder('nelmio_api_doc');
+        $treeBuilder->getRootNode()
```
dbu: this is missing the layer for supporting symfony before 4.2. you need to do something like here: https://github.com/php-http/HttplugBundle/blob/852f7f9fd71075a3d1ae41a3f8f46616447a9ddb/DependencyInjection/Configuration.php#L52-L58

mantis: @dbu - it looks like @GuilhemN has also started picking this up, so I think will leave to him - see https://github.com/nelmio/NelmioApiDocBundle/pull/1449 . I didn't realise when I submitted this PR that others had also done similar...

That we he can decide at what point to introduce this, and whether to do a separate version for >4.2 only etc; I note there's a separate issues/requests for jms/serializer v2 - I've implemented the reflection change locally (atlance's comment in 1437). jms v2 required php 7.2 so will need to be a decision as to what to support...

GuilhemN: I wasn't able to fix the tests so far, if you provide green tests I'd be glad to merge this PR :)

mantis: Wonder if this works

mantis:
@MrMitch The version in #1449 wasn't passing the test suite; so either wasn't quite right or the test suite is wrong. Same applies to the code here - this passes the test suite, however is it definitely working :)

MrMitch:
@mantis thanks, this seems OK, I'm waiting for this to be merged and released.
@dbu and @mantis : thanks for your work on this !

GuilhemN:
Great, thanks!

phansys:
Is there a chance to backport this FC layer to `2.2`?
Based on your answer, I can cherry-pick these commits and provide a PR.

/cc @GuilhemN.

GuilhemN:
@phansys sure thing, a PR would be great :)

Issue #1246, opened 2018-03-01
Add an annotation to specify the area(s) of a route


babaorum:
I would like to use the areas to specify multiple pages.
The problem with the **path_patterns** is I would need some complicated regular expressions.
Plus, if possible, I would like to abstract from the path to determine where a route should be displayed/described.

I think the easiest way would be to have an `@Area()` or `@Areas()` annotation?
I'm interested in contributing if you approve this feature, but I would love some guidance on the subject.

GuilhemN:
We discussed this in https://github.com/nelmio/NelmioApiDocBundle/pull/1207 and I'm personally in favor of this new annotation, I'm not sure what name it should take though.

I'd say the only thing to do is to add a check in https://github.com/nelmio/NelmioApiDocBundle/blob/master/Routing/FilteredRouteCollectionBuilder.php#L40-L45 ensuring that annotated routes only get matched when presenting the correct annotation.

kadryjanek:
I'm also for adding `@Area` annotation as the functionality is called: `Areas`.

dbu:
just stumbled over needing this as well ;-)

as routes can be in several areas, the annotation should probably allow to specify a whole list of areas and thus be called `@Areas`, right? anybody up for writing a pull request with this feature?

kadryjanek:
I think, I can implement it :)

babaorum:
I started working on it this weekend.
It takes, a little more time than I thought... Right now, FilteredRouteCollectionBuilder only concern is the path matching. So the context is changing a bit.

todoricMDPI:
I also need this feature. @babaorum, any news? :+1:

PR #1460, opened 2019-01-08
add a way to filter areas on route by annotation

babaorum:
Fix https://github.com/nelmio/NelmioApiDocBundle/issues/1246

### Description

A new boolean configuration is available in the areas: `with_annotation`.
By setting it to true, every route will be filtered by annotation.
Every route to match will need to add the `@Areas` with the area name.

### Example

the `default` area would only match the testAction(), while the `area2` will match both.

```yaml
areas:
    default:
        path_patterns:
            - '^/api/'
        with_annotation: true
    area2:
        with_annotation: true
```

```php
class ApiController
{
    /**
     * @Areas({'default', 'area2'})
     */
    public function testAction()
    {
    }

    /**
     * @Areas({'area2'})
     */
    public function test2Action()
    {
    }
}
```

### DIFF
--- a/Annotation/Areas.php
+++ b/Annotation/Areas.php
@@ -1,0 +1,50 @@
+<?php
+
+/*
+ * This file is part of the NelmioApiDocBundle package.
+ *
+ * (c) Nelmio
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Nelmio\ApiDocBundle\Annotation;
+
+/**
+ * @Annotation
+ */
+final class Areas
+{
+    /** @var string[] */
+    private $areas;
+
+    public function __construct(array $properties)
+    {
+        if (!array_key_exists('value', $properties) || !is_array($properties['value'])) {
+            throw new \InvalidArgumentException('An array of areas was expected');
+        }
+
+        $areas = [];
+        foreach ($properties['value'] as $area) {
+            if (!is_string($area)) {
+                throw new \InvalidArgumentException('An area must be given as a string');
+            }
+
+            if (!in_array($area, $areas)) {
+                $areas[] = $area;
+            }
+        }
+
+        if (0 === count($areas)) {
+            throw new \LogicException('At least one area is expected');
+        }
+
+        $this->areas = $areas;
+    }
+
+    public function has(string $area): bool
+    {
+        return in_array($area, $this->areas, true);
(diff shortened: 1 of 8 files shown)


Annotation/Areas.php
```diff
@@ -0,0 +1,50 @@
...
+ */
+final class Areas
+{
+    /** @var string[] */
+    private $areas;
+
+    public function __construct(array $properties)
+    {
+        if (!array_key_exists('value', $properties) || !is_array($properties['value'])) {
+            throw new \InvalidArgumentException('An array of areas were expected');
```
GuilhemN: `was`?

Routing/FilteredRouteCollectionBuilder.php
```diff
@@ -74,4 +97,29 @@ private function matchHost(Route $route): bool
...
     }
+
+    private function matchAnnotation(Route $route): bool
+    {
+        if (false === $this->options['with_annotation']) {
+            return true;
+        }
+
+        try {
+            $method = $this->routeReflector->controllerReflectionMethod($route);
```
GuilhemN: I don't think this is a worth an util class, could we inline it please ?

DependencyInjection/NelmioApiDocExtension.php
```diff
@@ -100,15 +100,25 @@ public function load(array $configs, ContainerBuilder $container)
...
                     ->setPublic(false);
             } else {
                 $container->register(sprintf('nelmio_api_doc.routes.%s', $area), RouteCollection::class)
                     ->setPublic(false)
                     ->setFactory([
                         (new Definition(FilteredRouteCollectionBuilder::class))
-                            ->addArgument($areaConfig),
+                            ->setArguments(
+                                [
+                                    new Reference('annotations.cached_reader'),
```
GuilhemN: `annotation_reader` instead, caching is managed by symfony

babaorum:
I fixed everything according to your comments.
Let me know if you see anything else.

GuilhemN:
Thanks you @babaorum!

ikleiman:
Thanks @babaorum 

Worked like a charm, but had to use it with double quotes instead of single ones (as explained in https://stackoverflow.com/questions/6338178/troubleshooting-syntax-error-expected-plainvalue-got):
```

    /**
     * @Areas({"area2"})
     */
    public function test2Action()
    {
    }

```

PR #1463, opened 2019-01-11
Fix documentation decoration with API Platform

### LINKED ISSUE #1462: Unable to use documentation decorator with API Platform
We wanted to decorate the `api_platform.swagger.normalizer.documentation` service as documented at https://api-platform.com/docs/core/swagger#overriding-the-swagger-documentation to define some recurring parameters and easily reference them in all our resources.

The service is defined like this:
```
  App\Serializer\SwaggerDocumentationDecorator:
    autoconfigure: false
    decorates: 'api_platform.swagger.normalizer.documentation'
    arguments:
      $decorated: '@AppSerializer\SwaggerDocumentationDecorator.inner'

```

This works with the default Swagger UI shipped with API Platform, but unfortunately throws an exception when using the NelmioApiDoc:

```php
Argument 2 passed to Nelmio\ApiDocBundle\Describer\ApiPlatformDescriber::__construct() must be an instance of 
ApiPlatform\Core\Swagger\Serializer\ApiGatewayNormalizer or 
ApiPlatform\Core\Swagger\Serializer\DocumentationNormalizer, 
App\Serializer\SwaggerDocumentationDecorator given.
```

So the problem is, the constructor of \Nelmio\ApiDocBundle\Describer\ApiPlatformDescriber only allows exactly those 2 normalizers and fails on everything else.

Is there another way to define parameters in the global parameters section of OpenAPI?

elkangaroo:
Fixes #1462
Adjust ApiPlatformDescriber to allow any implementation of NormalizerInterface that supports normalization of the ApiPlatform Documentation class.

Also added a unit test for the class.

### DIFF
--- a/Describer/ApiPlatformDescriber.php
+++ b/Describer/ApiPlatformDescriber.php
@@ -12,19 +12,17 @@
 namespace Nelmio\ApiDocBundle\Describer;
 
 use ApiPlatform\Core\Documentation\Documentation;
-use ApiPlatform\Core\Swagger\Serializer\ApiGatewayNormalizer;
-use ApiPlatform\Core\Swagger\Serializer\DocumentationNormalizer;
-use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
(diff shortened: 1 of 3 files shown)


GuilhemN:
Great, thank you @elkangaroo!

PR #1469, opened 2019-02-01
Update to make trouble shooting easier

Jeroenkeesom:
Update to make trouble shooting easier with json error output enabled.


GuilhemN:
Thank you!

Issue #1470, opened 2019-02-13
Error in "php bin/console debug:autowiring"


peterprochaska:
When i execute `php bin/console debug:autowiring` i get the following error:

```
Fatal error: Class 'Symfony\Component\Form\AbstractTypeExtension' not found in C:\www\comeback\vendor\nelmio\api-doc-bundle\Form\Extensio
n\DocumentationExtension.php on line 22
```
Am I doing something wrong?
I'am using Symfony 4.2.3 and nelmio api doc bundle 3.3.1

GuilhemN:
See https://github.com/nelmio/NelmioApiDocBundle/pull/1472, the issue is that we autowire a possibly broken service (when forms are not installed).

peterprochaska:
Thank you very much for the fix

PR #1472, opened 2019-02-18
Do not autowire the form extension

GuilhemN:
Fixes https://github.com/nelmio/NelmioApiDocBundle/issues/1470

### DIFF
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -48,7 +48,7 @@
 
         <!-- Form Type Extensions -->
 
-        <service id="Nelmio\ApiDocBundle\Form\Extension\DocumentationExtension">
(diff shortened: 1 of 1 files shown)


Resources/config/services.xml
```diff
@@ -48,7 +48,7 @@
 
         <!-- Form Type Extensions -->
 
-        <service id="Nelmio\ApiDocBundle\Form\Extension\DocumentationExtension">
+        <service id="nelmio_api_doc.form.documentation_extension">
```
dbu: missing class="Nelmio\ApiDocBundle\Form\Extension\DocumentationExtension" attribute

GuilhemN: Indeed thanks, I pushed this change a bit too quickly :)

PR #1476, opened 2019-02-26
escaped regex

daimmo:


GuilhemN:
Since we use `{` as delimiter there is no need to escape `\`. What's the reasoning behing this change?

PR #1477, opened 2019-02-26
Fix schema in SwaggerDocblockConvertCommand

gesof:
The generated "schema" did not follow the standard resulting in the following error: 

> Argument 1 passed to EXSyst\Component\Swagger\Schema::mergeExtensions() must be of the type array, string given

### DIFF
--- a/UPGRADE-3.0.md
+++ b/UPGRADE-3.0.md
@@ -123,7 +123,7 @@ class SwaggerDocblockConvertCommand extends ContainerAwareCommand
 
             if ('POST' !== $apiDoc->getMethod()) {
                 $annotation .= ',
-     *         schema=""';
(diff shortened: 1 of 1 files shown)


UPGRADE-3.0.md
```diff
@@ -123,7 +123,7 @@ class SwaggerDocblockConvertCommand extends ContainerAwareCommand
 
             if ('POST' !== $apiDoc->getMethod()) {
                 $annotation .= ',
-     *         schema=""';
+     *         @SWG\Schema(type="'.$this->determineDataType($parameter).'")';
```
GuilhemN: The type definition (a few lines above) should be omited in fact when we're in this case.
Could you put it in a else statement please ?

GuilhemN:
Thank you @gesof!

PR #1480, opened 2019-02-27
Fix for missing basePath

netbull:
If it is used without `#/definitions/` the documentation gives an error: `Could not resolve reference because of: Tried to resolve a relative URL, without having a basePath..`


GuilhemN:
Good catch, thanks !

PR #1482, opened 2019-03-11
Use jms/serializer groups detection logic

goetas:
This PR moves outside of NelmioApiDocBundle the logic for groups detection.

In this way jms/serializer can change the groups logic without affection NelmioApiDocBundle code.

Another advantage or this solution is that will be possible to introduce easily other exclusion strategies (as versioning) without changing too much code.

TODO:

- [x] concept (works on my machine on relatively complex project)
- [x] update tests
- [x] release updates on the jms/serializer side

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,8 @@
         "api-platform/core": "^2.1.0",
         "friendsofsymfony/rest-bundle": "^2.0",
         "willdurand/hateoas-bundle": "^1.0|^2.0",
-        "jms/serializer-bundle": "^2.0|^3.0"
+        "jms/serializer-bundle": "^2.0|^3.0",
+        "jms/serializer": "^1.14|^3.0"
     },
     "suggest": {
         "api-platform/core": "For using an API oriented framework.",
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,6 @@ matrix:
 before_install:
     - phpenv config-rm xdebug.ini
 
-install: composer update $COMPOSER_FLAGS --prefer-stable
+install: composer update $COMPOSER_FLAGS
 
 script: ./phpunit
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -13,6 +13,7 @@ namespace Nelmio\ApiDocBundle\ModelDescriber;
 
 use Doctrine\Common\Annotations\Reader;
 use EXSyst\Component\Swagger\Schema;
+use JMS\Serializer\Context;
 use JMS\Serializer\Exclusion\GroupsExclusionStrategy;
 use JMS\Serializer\Naming\PropertyNamingStrategyInterface;
 use JMS\Serializer\SerializationContext;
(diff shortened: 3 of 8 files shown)


goetas:
@GuilhemN If you are OK with moving that logic outside of nelmio-apidoc, I can complete this

GuilhemN:
Indeed, it probably makes more sense to include this in JMS 👍

goetas:
Great! will let you know as soon as i am done!

composer.json
```diff
@@ -44,7 +44,8 @@
         "api-platform/core": "^2.1.0",
         "friendsofsymfony/rest-bundle": "^2.0",
         "willdurand/hateoas-bundle": "^1.0|^2.0",
-        "jms/serializer-bundle": "^2.0|^3.0"
+        "jms/serializer-bundle": "^2.0|^3.0",
+        "jms/serializer": "^1.0@dev|^2.0@dev"
```
goetas: this is just temporary, before the release

goetas:
@GuilhemN excluding the composer.json changes, this is ready for review (probably a first round).

Preamble: I'm going to release jms/serializer 3.0, that reverts the nested group strategy as it was in 1.x (see https://github.com/schmittjoh/serializer/pull/1071). 2.x will not be supported anymore. 
Users using 1.x, are encouraged to upgrade directly to 3.x.

Because of the jms/serializer changes, I think also that naelmio/api-doc should just drop support for jms/serializer 2.x, and focus only on 1.x and 3.x. 

This decision allows to not have to maintain anymore v1 and v2 tests, so I have just deleted the v2 tests.

Regarding the other changes, not jms/serializer exposes  `getGroupsFor()` that calculates the groups for a given object. To make it work I had to simulare a graph-traversal (since the nested groups depends on the class graph).

composer.json
```diff
@@ -44,7 +44,8 @@
         "api-platform/core": "^2.1.0",
         "friendsofsymfony/rest-bundle": "^2.0",
         "willdurand/hateoas-bundle": "^1.0|^2.0",
-        "jms/serializer-bundle": "^2.0|^3.0"
+        "jms/serializer-bundle": "^2.0|^3.0",
+        "jms/serializer": "dev-master as 2.99.0"
```
goetas: when this jms/serializer 3.0 will be tagged, this will be just `^1.14|^3.0`

composer.json
```diff
@@ -29,7 +29,7 @@
         "symfony/asset": "^3.4|^4.0",
         "symfony/console": "^3.4|^4.0",
         "symfony/config": "^3.4|^4.0",
-        "symfony/validator": "^3.4|^4.0",
+        "symfony/validator": "^3.4|~4.1.0",
```
goetas: this is a temporary fix for https://github.com/symfony/symfony/issues/31152

goetas:
@GuilhemN  this is ready! jms/serializer v3 released 

the failure is related to https://github.com/symfony/symfony/issues/31152

GuilhemN:
Great, thank you @goetas :)

Congrats for releasing JMS v3 😃

PR #1492, opened 2019-04-02
Provide default host to generated .json documentation

AyrtonRicardo:
In our scenario we can have multiples `urls` leading to the same project, and in the docs we can only provide one host to the complete documentation. To swagger ui it's not problem because the javascript if the `host` it's not provided, the `window.location` is used instead. So this pr is just to add the same behaviour to the .json generation. 

Another idea is provide a configuration to say something like `auto_discover_host`.

What do you think about it?

about build:
I don't from where this error is coming from.

### DIFF
--- a/Controller/DocumentationController.php
+++ b/Controller/DocumentationController.php
@@ -48,8 +48,13 @@ final class DocumentationController
         }
 
         $spec = $this->generatorLocator->get($area)->generate()->toArray();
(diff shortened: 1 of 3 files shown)


GuilhemN:
Your solution is fine 👍  no need to add a config option for that imo.

Thank you!

PR #1495, opened 2019-04-09
Allow model describers on jms custom types

goetas:
Currently is not possible to describe at all custom jms type handlers (they are pretty common in complex apis). This change allows to describe via custom describers jms virtual types.

The change in behavior from the current status is substantial.
The current implementation will just not define the type for a property, while with this change sill throw an exception saying something as "Schema of type "...." can't be generated, no describer supports it".
This obliges the user to have all the types defined. Not sure if it is the best approach.

Solution used here is to add a fallback describer, defined for all the types, that just does nothing and emulates the current behavior.
The only drawback from this solution is that will create empty types jms virtual types... that could be a not bad thing.

@GuilhemN what do you think?
If you are ok with this changes I can add tests for this.

### DIFF
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -177,11 +177,6 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
             $property->setType('string');
             $property->setFormat('date-time');
         } else {
-            // we can use property type also for custom handlers, then we don't have here real class name
-            if (!class_exists($type['name'])) {
-                return null;
-            }
-
             $model = new Model(new Type(Type::BUILTIN_TYPE_OBJECT, false, $type['name']), $groups);
             $property->setRef($this->modelRegistry->register($model));
 
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -46,6 +46,10 @@
             <tag name="nelmio_api_doc.model_describer" />
         </service>
 
+        <service id="nelmio_api_doc.model_describers.object_fallback" class="Nelmio\ApiDocBundle\ModelDescriber\FallbackObjectModelDescriber" public="false">
+            <tag name="nelmio_api_doc.model_describer" priority="-1000" />
+        </service>
+
         <!-- Form Type Extensions -->
 
         <service id="nelmio_api_doc.form.documentation_extension" class="Nelmio\ApiDocBundle\Form\Extension\DocumentationExtension">
--- a/DependencyInjection/NelmioApiDocExtension.php
+++ b/DependencyInjection/NelmioApiDocExtension.php
@@ -173,6 +173,8 @@ final class NelmioApiDocExtension extends Extension implements PrependExtensionI
                         new Reference('nelmio_api_doc.model_describers.jms.inner'),
                     ]);
             }
+        } else {
+            $container->removeDefinition('nelmio_api_doc.model_describers.object_fallback');
         }
 
         // Import the base configuration
--- a/ModelDescriber/FallbackObjectModelDescriber.php
+++ b/ModelDescriber/FallbackObjectModelDescriber.php
@@ -1,0 +1,28 @@
+<?php
+
+/*
+ * This file is part of the NelmioApiDocBundle package.
+ *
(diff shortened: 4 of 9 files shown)


GuilhemN:
How does Swagger-ui react to empty definitions?
Doesn't this create new issues by creating an invalid documentation?

goetas:
> How does Swagger-ui react to empty definitions?
Doesn't this create new issues by making the documentation invalid?

It is not a problem, is equivalent to an object with no props.
From the UI, it looks as this: 

image

GuilhemN:
Ok, then it's fine for me to do this change  :+1:

goetas:
Great!

goetas:
Done! Tests added.

it looks that something changed in the latest releases of API platform (using `--prefer-lowest` the problem does not occur.)
and in the phpunit bridge cant be installed anymore.

composer.json
```diff
@@ -55,6 +55,11 @@
             "Nelmio\\ApiDocBundle\\": ""
         }
     },
+    "autoload-dev": {
+        "psr-4": {
+            "Nelmio\\ApiDocBundle\\Tests\\": "Tests"
```
GuilhemN: What is this change for?

Resources/config/services.xml
```diff
@@ -46,6 +46,10 @@
             <tag name="nelmio_api_doc.model_describer" />
         </service>
 
+        <service id="nelmio_api_doc.model_describers.object_fallback" class="Nelmio\ApiDocBundle\ModelDescriber\FallbackObjectModelDescriber" public="false">
```
GuilhemN: In fact I wonder if this is makes sense when using symfony's serializer. Shouldn't it be added only when using JMS?

GuilhemN:
I fixed the Api-platform issue in https://github.com/nelmio/NelmioApiDocBundle/pull/1498, thanks for the report :)

composer.json
goetas: My bad, did not notice, will revert it

Resources/config/services.xml
goetas: Hmm. It could be. This change is needed because now the ObjectModelDescriber does not accept all the class types but only the classes that actually exists.
But I can change it and add this describer only when jms is enabled.

GuilhemN: I think this would make more sense :)
We should throw when we get an unexistent class without jms

ModelDescriber/JMSModelDescriber.php
```diff
@@ -177,11 +177,6 @@ private function describeItem(array $type, $property, array $groups = null, arra
             $property->setType('string');
             $property->setFormat('date-time');
         } else {
-            // we can use property type also for custom handlers, then we don't have here real class name
-            if (!class_exists($type['name'])) {
-                return null;
-            }
-
             $model = new Model(new Type(Type::BUILTIN_TYPE_OBJECT, false, $type['name']), $groups);
             $property->setRef($this->modelRegistry->register($model));
```
goetas: @GuilhemN 
A problem that this PR does not solve is how to allow custom handlers to be mapped to simple types (as strings integers or enums...).

Currently `JMSModelDescriber` calls `setRef` on the schema object. This forces that any custom type is an "object". 
Any idea on how it could be possible to allow calling `setType` or `setEnum` on the `$property` object?

Resources/config/services.xml
goetas: Done in https://github.com/nelmio/NelmioApiDocBundle/pull/1495/commits/f023a9bbba16d180127cbd6694c6232fe162d3af

goetas:
@GuilhemN I have done all the changes requested.
If you want to postpone the decision for https://github.com/nelmio/NelmioApiDocBundle/pull/1495#discussion_r275222348 , then this PR is ready from my side.

ModelDescriber/JMSModelDescriber.php
GuilhemN: The current design won't allow doing this easily, I'd say let's wait and see if we have a real case :)

GuilhemN:
Thank you @goetas!

goetas:
thanks! :)

Issue #1465, opened 2019-01-16
"Could not render this component, see the console" error for QueryParam with map=true


spolischook:
The `@SWG\Parameter` works ok, but QueryParam produce an error in api doc
![selection_283](https://user-images.githubusercontent.com/1978331/51243135-64de9200-198a-11e9-9138-92b566aca95e.png)
Here is code for produce this result:
```php
class ApiController
{
    /**
     * @SWG\Response(
     *     response=200,
     *     description="List of products"
     * )
     * @SWG\Parameter(
     *     name="cat",
     *     in="query",
     *     type="array",
     *     collectionFormat="multi",
     *     @SWG\Items(
     *         type="string"
     *     ),
     *     description="The field used to order rewards"
     * )
     * @QueryParam(map=true, name="categories", requirements="\d+", description="Id of categories to filter.")
     */
    public function getProductsAction()
    {
        return [];
    }
}
```
for the code example see https://github.com/spolischook/api-doc-test

spolischook:
@GuilhemN can you reproduce this issue?

GuilhemN:
Yes, sorry for the delay I missed the notification...

Thank you for the precise report, I opened https://github.com/nelmio/NelmioApiDocBundle/pull/1506/files which should solve your issue.
Could you give it a try?

Issue #1490, opened 2019-04-01
Property collectionFormat multi bug


lukepass:
Hello,
I encountered a minor bug using collection format. This is my action:

    /**
     * Creates a new product.
     *
     * @Rest\Post("/create")
     * @SWG\Response(
     *     response=200,
     *     description="Created product"
     * )
     * @SWG\Response(
     *     response=400,
     *     description="901 - Invalid input"
     * )
     * @SWG\Parameter(
     *     name="code",
     *     in="formData",
     *     type="string",
     *     required=true
     * )
     * @SWG\Parameter(
     *     name="image",
     *     in="formData",
     *     type="file"
     * )
     * @SWG\Parameter(
     *     name="machineries[]",
     *     in="formData",
     *     type="array",
     *     @SWG\Items(type="integer"),
     *     collectionFormat="multi",
     * )
     */
    public function createAction(Request $request)
    {
    }

This is the generated call:

`curl -X POST "http://mysite.local/app_dev.php/api/products/create" -H "accept: application/json" -H "Content-Type: multipart/form-data" -F "code=12345" -F "machineries[]="123","321""`

I would have expected the `machineries[]` parameter repeated 2 times instead of a single parameter with multiple values:

https://swagger.io/docs/specification/2-0/describing-parameters/#array-and-multi-value-parameters

Thanks.

lukepass:
Hi @GuilhemN, did you have time to check the submitted bug? Thanks!

GuilhemN:
This part is managed by [swagger-ui](https://github.com/swagger-api/swagger-ui), and I don't know its internals.
Have you tried posting a message on one of symfony's support channels ? (see https://symfony.com/support)

GuilhemN:
Just in case, have you tried using type ``string`` instead of `int`?

chrisguitarguy:
Closing this due to age and OpenAPI 2.X use.

For those that land here: https://swagger.io/docs/specification/describing-parameters/#query-parameters

The `explode` keyword is probably what you want in the parameter definition:

```php
use OpenApi\Annotations as OA;

/**
 * @OA\Parameter(
 *     in="query",
 *     name="example",
 *     schema=@OA\Schema(type="array", items=@OA\Items(type="integer")),
 *     explode=true,
 * )
 */
```

PR #1506, opened 2019-04-20
Fix map support of FOSRestBundle

GuilhemN:
Fixes https://github.com/nelmio/NelmioApiDocBundle/issues/1465

### DIFF
--- a/RouteDescriber/FosRestDescriber.php
+++ b/RouteDescriber/FosRestDescriber.php
@@ -61,16 +61,16 @@ final class FosRestDescriber implements RouteDescriberInterface
                 if (null !== $parameter->getType()) {
                     continue;
                 }
-                        
+
                 if (null === $parameter->getDescription()) {
                     $parameter->setDescription($annotation->description);
                 }
-                
+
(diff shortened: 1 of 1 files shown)


spolischook:
If almost works
Selection_063
it generates parameters as `id=1465,145` but should `id[]=1465&id[]=145`

GuilhemN:
This is most likely a bug in swagger-ui (see https://github.com/nelmio/NelmioApiDocBundle/issues/1490). I'm not sure what to do about this one, maybe it is fixed in the latest versions of swagger-ui and we only need to bump it

spolischook:
why this closed?

GuilhemN:
The correct fix was https://github.com/nelmio/NelmioApiDocBundle/pull/1509, right?

spolischook:
it was changes to your changes in this PR - so this PR was updated and should be merged as well

spolischook:
I have choose your branch as target for my PR

GuilhemN:
Oops, you're right I didn't realize your PR wasn't targetting master :)
Thanks for the reminder!

PR #1509, opened 2019-04-23
Fixed names and multi for array query params

spolischook:


spolischook:
@GuilhemN this is the fix. It works, I've checked.

GuilhemN:
Thank you @spolischook!

PR #1510, opened 2019-04-24
Support typed embedded relation with willdurand/hateoas 3.0

goetas:
This "documents" the type on an embedded relation in hateoas 3.0

Previously the embedded relations were just simple `object` types.
As of hateoas 3.0 is possible to declare via metadata the type of an embedded relation, so that type can be used in the api doc.

image

### DIFF
--- a/ModelDescriber/JMSModelDescriber.php
+++ b/ModelDescriber/JMSModelDescriber.php
@@ -118,6 +118,9 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
         $context->popClassMetadata();
     }
 
+    /**
+     * @internal
+     */
     public function getSerializationContext(Model $model): SerializationContext
     {
         if (isset($this->contexts[$model->getHash()])) {
@@ -178,6 +181,10 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
         return false;
     }
 
+    /**
+     * @internal
+     * @return void
+     */
     public function describeItem(array $type, $property, Context $context)
     {
         $nestedTypeInfo = $this->getNestedTypeInArray($type);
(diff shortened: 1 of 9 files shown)


ModelDescriber/JMSModelDescriber.php
```diff
@@ -178,7 +178,7 @@ public function supports(Model $model): bool
         return false;
     }
 
-    private function describeItem(array $type, $property, Context $context)
+    public function describeItem(array $type, $property, Context $context)
```
GuilhemN: Should it be internal?

goetas: Definitely yes. Will update the pr

goetas: done

Tests/Functional/TestKernel.php
```diff
@@ -88,6 +90,12 @@ protected function configureRoutes(RouteCollectionBuilder $routes)
 
         if ($this->useBazinga) {
             $routes->import(__DIR__.'/Controller/BazingaController.php', '/', 'annotation');
+
+            try {
+                new \ReflectionMethod(Embedded::class, 'getType');
```
GuilhemN: Is this to replace `method_exists`?

ModelDescriber/JMSModelDescriber.php
```diff
@@ -118,7 +118,7 @@ public function describe(Model $model, Schema $schema)
         $context->popClassMetadata();
     }
 
-    private function getSerializationContext(Model $model): SerializationContext
+    public function getSerializationContext(Model $model): SerializationContext
```
GuilhemN: Should be internal too I think

goetas: My bad, now is fixed

goetas:
`@return void` looks forbidden by StyleCI, but IMO should be allowed as NelmioApiDocBundle targets PHP 7.0 that does not allow the `void` return type-hint

GuilhemN:
Thank you @goetas :)

goetas:
:tada:

PR #1521, opened 2019-05-24
Excluded tests from classmap

samnela:


GuilhemN:
Sorry for the delay, let's do it!
Thank you @samnela :)

PR #1530, opened 2019-07-03
Fix Controllers used as Services in Symfony >=4.1

DemigodCode:
In Symfony 4.1 the route name is also built with "::" instead of ":" when using services.

### DIFF
--- a/Util/ControllerReflector.php
+++ b/Util/ControllerReflector.php
@@ -100,7 +100,7 @@ class ControllerReflector
             $method = '__invoke';
         } else {
             // Has to be removed when dropping support of symfony < 4.1
-            if (Kernel::VERSION_ID < 40100 && preg_match('#(.+):([\w]+)#', $controller, $matches)) {
+            if (preg_match('#(.+):([\w]+)#', $controller, $matches)) {
                 $controller = $matches[1];
                 $method = $matches[2];
             }
--- a/Util/ControllerReflector.php
+++ b/Util/ControllerReflector.php
@@ -83,8 +83,6 @@ class ControllerReflector
             $controller = $this->controllerNameParser->parse($controller);
         }
 
-
-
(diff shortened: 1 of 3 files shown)


DemigodCode:
What's the preferred way? We can disable the deprecations to support lower symfony versions, or we have to drop symfony < 4.2 support. Any other ideas?

Util/ControllerReflector.php
```diff
@@ -85,11 +86,21 @@ private function getClassAndMethod(string $controller)
...
+                    if (class_exists(ClassUtils::class)) {
+                        $class = ClassUtils::getRealClass($class);
+                    }
+                }
+            }
         } elseif (class_exists($controller)) {
             $class = $controller;
             $method = '__invoke';
         } else {
-            if (preg_match('#(.+):([\w]+)#', $controller, $matches)) {
```
GuilhemN: Support of this syntax is deprecated in sf > 4.1, not removed, right? 
I think we should keep this part as it is.

DemigodCode:
I tried to get rid of the failing tests. But it doesn't work. I can revert them, if you want :)

GuilhemN:
Thank you @DemigodCode!

Util/ControllerReflector.php
```diff
@@ -85,10 +86,20 @@ private function getClassAndMethod(string $controller)
         if (preg_match('#(.+)::([\w]+)#', $controller, $matches)) {
             $class = $matches[1];
             $method = $matches[2];
+            // Since symfony 4.1 routes are defined like service_id::method_name
+            if (Kernel::VERSION_ID >= 40100 && !class_exists($class)) {
+                if ($this->container->has($class)) {
+                    $class = get_class($this->container->get($class));
+                    if (class_exists(ClassUtils::class)) {
```
maxhelias: `ClassUtils` is not imported. We will never go into this if

Issue #1537, opened 2019-08-04
Command line availability


mantis:
Hello,

Are there any plans to have a command line option to dump out a swagger.json documentation file ?

This would basically be similar to DocumentationController, with $spec, passed through json_encode( JSON_PRETTY_PRINT ); 

(or is it preferred to download the file from the webui via a different method?)

Thanks

GuilhemN:
This hasn't be considered but it would be definitely be great to have.

Would you like to submit a PR?

mantis:
I've had a stab at a sample command with #1540

GuilhemN:
Fixed in #1540 :)

PR #1540, opened 2019-08-14
Add dump command to print out swagger JSON. (#1537)

mantis:
Command outputs to stdout, so can be redirected to a file.
Pretty prints output by default.
--no-pretty can be passed to remove whitespace from json output if it does not need to be human readable

### DIFF
--- a/Command/DumpCommand.php
+++ b/Command/DumpCommand.php
@@ -18,8 +18,6 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\DependencyInjection\ServiceLocator;
-use Symfony\Component\HttpFoundation\JsonResponse;
-use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 
 class DumpCommand extends Command
@@ -60,7 +58,7 @@ class DumpCommand extends Command
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $area = $input->getOption( 'area' );
+        $area = $input->getOption('area');
 
         if (!$this->generatorLocator->has($area)) {
             throw new BadRequestHttpException(sprintf('Area "%s" is not supported.', $area));
@@ -68,10 +66,10 @@ class DumpCommand extends Command
 
         $spec = $this->generatorLocator->get($area)->generate()->toArray();
 
-        if( $input->hasParameterOption(['--no-pretty']) ) {
-            $output->writeln( json_encode( $spec ) );
+        if( $input->hasParameterOption(['--no-pretty'])) {
+            $output->writeln(json_encode($spec));
         } else {
-            $output->writeln( json_encode( $spec, JSON_PRETTY_PRINT ) );
+            $output->writeln(json_encode($spec, JSON_PRETTY_PRINT));
         }
     }
 }
--- a/Command/DumpCommand.php
+++ b/Command/DumpCommand.php
@@ -26,22 +26,8 @@ class DumpCommand extends Command
 
     private $generatorLocator;
 
-    /**
-     * @param ContainerInterface $generatorLocator
-     */
-    public function __construct($generatorLocator)
+    public function __construct(ContainerInterface $generatorLocator)
     {
-        if (!$generatorLocator instanceof ContainerInterface) {
-            if (!$generatorLocator instanceof ApiDocGenerator) {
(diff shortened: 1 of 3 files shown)


b-durand:
@mantis any update or vote to merge this PR?

mantis:
I can look to see if it needs updating and rebase, but the bundle author updated the commits so i'd assumed that he was going to handle it from there and decide whether it was suitable.. Maybe I was wrong, but hadn't realise it was back with me to progress further as a concept.

mantis:
have done some minor updates to this to clear the StyleCI warnings and add some phpdoc comments

mantis:
@FilipBenco I looked at updating this PR earlier today, and whilst I think it should probably also have an option to save to file (not just dump to console) - am I correct in thinking this PR wouldn't be affected by the OpenApi changes?

Command/DumpCommand.php
```diff
@@ -0,0 +1,81 @@
...
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $area = $input->getOption('area');
+
+        if (!$this->generatorLocator->has($area)) {
+            throw new InvalidArgumentException(sprintf('Area "%s" is not supported.', $area));
+        }
+
+        $spec = $this->generatorLocator->get($area)->generate()->toArray();
```
FilipBenco, line 73: In OA3 update you cannot convert generated specs to array using `toArray()` method. Generated Specs object is jsonserializable, so you can just use returned value from `$this->generatorLocator->get($area)->generate()` as first parameter of `json_encode`. But this works only of OA3, not swagger.

Command/DumpCommand.php
```diff
@@ -0,0 +1,81 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class DumpCommand extends Command
+{
+    /**
+     * @var string command name
+     */
+    protected static $defaultName = 'app:doc:dump';
```
FilipBenco, line 26: here you have `app:doc:dump` and in services.yaml you specify it as `api:doc:dump`. Maybe we can unify it :)

FilipBenco:
@mantis I commented on proposed code and also mentioned think, that should be changed after OA3 is merged.

Command/DumpCommand.php
mantis: I believe that block can just be deleted instead ;)

mantis:
Update for latest master. Seems to work fine here ;)

Resources/config/services.xml
```diff
@@ -4,6 +4,12 @@
     xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
     <services>
+        <!-- Commands -->
+        <service id="nelmio_api_doc.command.dump" class="Nelmio\ApiDocBundle\Command\DumpCommand" public="true">
+            <argument type="service" id="nelmio_api_doc.generator_locator" />
+            <tag name="console.command" command="api:doc:dump" />
```
GuilhemN, line 10: I wonder if it shouldn't be named something like `nelmio:doc:dump`, don't you think?

Command/DumpCommand.php
```diff
@@ -45,7 +45,7 @@ protected function configure()
         $this
             ->setDescription('Dumps documentation in OpenAPI JSON format')
             ->addOption('area', '', InputOption::VALUE_OPTIONAL, '', 'default')
-            ->addOption('--no-pretty', '', InputOption::VALUE_NONE, 'Do not pretty format output')
+            ->addOption('no-pretty', '', InputOption::VALUE_NONE, 'Do not pretty format output')
```
mantis, line 48: would the 2nd parameter allow np to be specified ? for a shorter version?

Resources/config/services.xml
mantis: you are correct unfortunately ;/ - is there a sensible way that a user could configure the command name for their app.

nelmio:doc:dump would ensure no clashes with someone's application; where they might already use api:* . At the same time, api:doc probably makes more sense to a user and more memorable.

In any case, where nelmio has other modules e.g. cores - should probably be nelmio:api:dump or the more verbose nelmio:apidoc:dump.  Just thinking about if you added a console command for some of the other bundles one day...

In any case, I think you are correct using api:doc isn't the best idea for the module.

Command/DumpCommand.php
GuilhemN: not sure that's needed, I guess this option should be used in Makefiles or similar intermediate and the verbosity doesn't hurt there in my opinion.

Resources/config/services.xml
GuilhemN: `nelmio:apidoc:dump` looks good as it is in sync with the bundle's name :)

GuilhemN:
Thank you @mantis!

PR #1552, opened 2019-11-05
Added describer that removes duplicate parameters when using $ref.

### LINKED ISSUE #1407: Referencing a parameter from the config duplicates it, resulting in broken Swagger docs
Let's say I have a controller action like this one:

```php
    /**
     * @param Request $request
     * @param string $id
     * @return \Symfony\Component\HttpFoundation\Response
     * @throws \Doctrine\ORM\NonUniqueResultException
     * @Rest\Get("/{id}")
     * @SWG\Response(
     *     response="200",
     *     description="Returns the given language for this contest",
     *     @Model(type=Language::class)
     * )
     * @SWG\Parameter(ref="#/parameters/id")
     */
    public function getSingleAction(Request $request, string $id)
    {
        // Some code here
    }
```

And assume the following in my config.yml:

```yaml
nelmio_api_doc:
    documentation:
        parameters:
            id:
                name: id
                in: path
                type: string
                description: The ID of the entity to get
                required: true
```

This will result in the API docs for the `getSingleAction` to have two parameters, one for the `ref` parameter and one for `id/path`. The first one is added by the `SwaggerPhpDescriber`, the second one by the `RouteDescriber`, specifically the `RouteMetadataDescriber`.

I think the `RouteMetadataSubscriber` should check `$operation` and loop over all parameters, checking if any of them is a `ref` parameter pointing to the same `$pathVariable` with an `in` of `path`. If so, it should skip this parameter instead of adding it again.

The Swagger UI is broken when not doing this, because it expects two required parameters with the same name.

I can create a PR if desired that fixes this.

For now, and for anyone looking for a temporary solution, I have created my own Describer class and added it as a describer with priority -999 that removes these duplicate parameters:

```php
<?php

namespace DOMJudgeBundle\ApiDescriber;

use EXSyst\Component\Swagger\Operation;
use EXSyst\Component\Swagger\Parameter;
use EXSyst\Component\Swagger\Path;
use EXSyst\Component\Swagger\Swagger;
use Nelmio\ApiDocBundle\Describer\DescriberInterface;

/**
 * Class ParameterRefMergeDescriber
 * @package DOMJudgeBundle\ApiDescriber
 */
class ParameterRefMergeDescriber implements DescriberInterface
{
    /**
     * @inheritdoc
     */
    public function describe(Swagger $api)
    {
        // This method removes paramaters that also have a ref version as they will be duplicated otherwise

        /** @var Path $path */
        foreach ($api->getPaths() as $path) {
            /** @var Operation $operation */
            foreach ($path->getOperations() as $operation) {
                $parametersToRemove = [];
                /** @var Parameter $parameter */
                foreach ($operation->getParameters() as $parameter) {
                    if ($parameter->getRef()) {
                        $ref = substr($parameter->getRef(), 13);
                        if (isset($api->getParameters()[$ref])) {
                            /** @var Parameter $refParameter */
                            $refParameter   = $api->getParameters()[$ref];
                            $refParameterId = sprintf('%s/%s', $refParameter->getName(), $refParameter->getIn());
                            if ($operation->getParameters()->has($refParameterId)) {
                                $parametersToRemove[] = $operation->getParameters()->get($refParameterId);
                            }
                        }
                    }
                }
                foreach ($parametersToRemove as $parameterToRemove) {
                    $operation->getParameters()->remove($parameterToRemove);
                }
            }
        }
    }
}
```

cyberemissary:
Fixes #1407


cyberemissary:
Rebased to latest master, tests with symfony 5 failing due to composer install, which was not affected by my changes.

GuilhemN:
I am not fund of removing parameters arbitrarily when the issue actually is in the `RouteMetadataDescriber`, I've reworked your PR to show you how I'd rather do it.
Are these changes acceptable to you ? Does it still correctly fix your issue ?

RouteDescriber/RouteMetadataDescriber.php
```diff
@@ -26,20 +29,47 @@ public function describe(Swagger $api, Route $route, \ReflectionMethod $reflecti
             $requirements = $route->getRequirements();
             $compiledRoute = $route->compile();
 
+            $globalParams = $api->getParameters();
+            $existingParams = [];
+            foreach ($operation->getParameters() as $id => $parameter) {
+                $ref = $parameter->getRef();
+                if (null === $ref) {
+                    $existingParams[$id] = true;
```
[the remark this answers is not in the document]
cyberemissary: I removed this in my refactored code. Please let me know if you think this really should be there, and I can put it back.

RouteDescriber/RouteMetadataDescriber.php
```diff
@@ -25,24 +30,59 @@ public function describe(Swagger $api, Route $route, \ReflectionMethod $reflecti
...
 
             // Don't include host requirements
             foreach ($compiledRoute->getPathVariables() as $pathVariable) {
                 if ('_format' === $pathVariable) {
                     continue;
                 }
 
-                $parameter = $operation->getParameters()->get($pathVariable, 'path');
+                $paramId = $pathVariable.'/path';
+                $parameter = $existingParams[$paramId] ?? $operation->getParameters()->get($pathVariable, 'path');
```
GuilhemN: We should just do nothing if there is a already a ref parameter in my opinion, it would be quite strange to update a global parameter due to a local constraint.

RouteDescriber/RouteMetadataDescriber.php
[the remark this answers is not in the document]
GuilhemN: It is supposed to be the parameter identifier (`name/in`) but it shouldn't be added to this array anyway so your removal is fine.

RouteDescriber/RouteMetadataDescriber.php
cyberemissary: The reason I did this here is, if they forget to specify it as required, we will have issues in the generated doc.

I agree that updating global ref would be *magic*, but maybe at least we should throw exception here if it is not set as required?

GuilhemN: Sure that sounds reasonnable

cyberemissary: Done

cyberemissary:
@GuilhemN, I have updated the code per our discussion and rebased. Hopefully this will be good to go now 😄

GuilhemN:
That's great, thank you :)

PR #1557, opened 2019-11-19
Remove deprecated notice about controller_name_converter

maxhelias:


GuilhemN:
Thank you @maxhelias !

PR #1558, opened 2019-11-20
Fix build travis

maxhelias:


composer.json
```diff
@@ -37,7 +37,8 @@
         "symfony/cache": "^3.4|^4.0",
         "symfony/phpunit-bridge": "^3.4.24|^4.0",
         "symfony/stopwatch": "^3.4|^4.0",
-        "sensio/framework-extra-bundle": "^3.0",
+        "symfony/routing": "^3.4|^4.0",
```
fabianoroberto: Add also `^5.0` version

maxhelias: This is the next step :)
First we need to have all continuous integration on green and then add some steps to check compatibility with Symfony 5

GuilhemN:
Whoa, that's great :)
Thank you @maxhelias!

PR #1564, opened 2019-11-22
Allow Symfony 5

maxhelias:
- I remove support for PHP 7.1 as she is no longer maintained and she don't support `iterable` type for compatibility with Symfony 5.

- Add PHP 7.3 and 7.4 tests

- Try different versions of Symfony

TODO :

- [ ] Fix Travis 😄


maxhelias:
@GuilhemN I think we can merge this PR even if the Sf5 test fails because of FOSRestBundle. It should automatically go green at the next release of FOSRest

Form/Extension/DocumentationExtension.php
```diff
@@ -37,7 +37,7 @@ public function getExtendedType()
         return self::getExtendedTypes()[0];
     }
 
-    public static function getExtendedTypes()
+    public static function getExtendedTypes(): iterable
     {
         return [FormType::class];
```
mantis: I've not tried this (was just looking at symfony 5 status), but should the return statement not be:

return array(FormType::class);

the addition of array() is in the upgrade doc's example?

maxhelias: I do not understand what you mean :/

mantis: I happened to read https://github.com/symfony/symfony/blame/master/UPGRADE-5.0.md#L153 about 30 seconds before reading your PR. The symfony guide 'after' example added array()

GuilhemN:
Thank you @maxhelias ! :)

Form/Extension/DocumentationExtension.php
ThibaultVlacich: @mantis `array()` and `[]` are the same thing. It's just two different ways of writing arrays in PHP. `[]` being more modern.

mantis: argh, sorry - I missed the []

PR #1575, opened 2019-12-13
Extensible property describing

FilipBenco:
I need som customization, so I can support also MyClabs\Enum . 

Because I want to support this object type I need to be able to hook in the part, that is describing object properties.

I refactored property describing into property describers, so anybody can hook in and add custom describers if wanted.

Here is also example of my ProperyDescriber for Enum:
```php
<?php

declare(strict_types=1);

namespace WebSupport\RestBundle\Nelmio\ApiDoc\PropertyDescriber;

use EXSyst\Component\Swagger\Schema;
use MyCLabs\Enum\Enum;
use Nelmio\ApiDocBundle\Describer\ModelRegistryAwareInterface;
use Nelmio\ApiDocBundle\Describer\ModelRegistryAwareTrait;
use Nelmio\ApiDocBundle\Model\Model;
use Nelmio\ApiDocBundle\PropertyDescriber\PropertyDescriberInterface;
use Symfony\Component\PropertyInfo\Type;

class EnumPropertyDescriber implements PropertyDescriberInterface, ModelRegistryAwareInterface
{
    use ModelRegistryAwareTrait;
    
    public function describe(Type $type, Schema $property, array $groups)
    {
        $type = new Type($type->getBuiltinType(), false, $type->getClassName());

        $property->setRef(
            $this->modelRegistry->register(new Model($type, []))
        );
    }

    public function supports(Type $type): bool
    {
        return Type::BUILTIN_TYPE_OBJECT === $type->getBuiltinType()
            && is_a($type->getClassName(), Enum::class, true);
    }

}
```


FilipBenco:
@GuilhemN could you please look at this PR? (1 job that is failing in pipeline is failing also in master branch).
What do you think about this feature? I would love to have it merged 👼 🙂

maxhelias:
Don't worry, the fail is related to friendsofsymfony/rest-bundle which is not yet compatible with Sf5

It's a nice feature for me and it looks good but I still don't have enough knowledge about this part to validate

PropertyDescriber/ArrayPropertyDescriber.php
```diff
@@ -0,0 +1,57 @@
...
+use Symfony\Component\PropertyInfo\Type;
+
+class ArrayPropertyDescriber implements PropertyDescriberInterface, ModelRegistryAwareInterface
+{
+    use ModelRegistryAwareTrait;
+
+    /** @var PropertyDescriberInterface[] */
+    private $propertyDescribers;
+
+    public function __construct($propertyDescribers = [])
```
GuilhemN: Couldn't we add an `iterable` type hint here ?

GuilhemN:
I'm a bit uncomfortable with the fact that this is only supported when using sf serializer (and not when using jms serializer), what do you think @goetas ?

Except that interrogation that is more about how we want the bundle to evolve, your changes look all good to me.

PropertyDescriber/PropertyDescriberInterface.php
```diff
@@ -0,0 +1,22 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Nelmio\ApiDocBundle\PropertyDescriber;
+
+use EXSyst\Component\Swagger\Schema;
+use Symfony\Component\PropertyInfo\Type;
+
+interface PropertyDescriberInterface
```
goetas: I'm not sure we need this interface. Is basically identical to `ModelDescriberInterface`.

PropertyDescriber/PropertyDescriberInterface.php
```diff
@@ -0,0 +1,22 @@
...
+{
+    public function describe(Type $type, Schema $property, array $groups = null);
+
+    public function supports(Type $type): bool;
```
goetas: The `supports()` method and the way is used in this PR (and in many parts of NelmioApiDocBundle) is a modified version of the chain of responsibility pattern.
But this modification, requires an additional class (in our case the `ArrayPropertyDescriber`)... if you drop the `supports()` method and just let the `describe()` method skip the not supported classes you get the same result and you need at all `ArrayPropertyDescriber`.

PropertyDescriber/ArrayPropertyDescriber.php
```diff
@@ -0,0 +1,57 @@
...
+        $type = $type->getCollectionValueType();
+        if (null === $type) {
+            throw new \LogicException(sprintf('Property "%s:%s" is an array, but its items type isn\'t specified. You can specify that by using the type `string[]` for instance or `@SWG\Property(type="array", @SWG\Items(type="string"))`.', $class, $propertyName));
+        }
+
+        $property->setType('array');
+        $property = $property->getItems();
+
+        foreach ($this->propertyDescribers as $propertyDescriber) {
+            if ($propertyDescriber instanceof ModelRegistryAwareInterface) {
```
goetas: The `*AwareInterface` interfaces have been proven to be a bad practice (even symfony is discouraging the use of ContainerAware interfaces and similar), I do not see a reason to propagate this bad practice in new code. If a propery describer needs it, should be able to get it via DI.

PropertyDescriber/PropertyDescriberInterface.php
FilipBenco: changed

PropertyDescriber/ArrayPropertyDescriber.php
FilipBenco: updated

PropertyDescriber/ArrayPropertyDescriber.php
FilipBenco: Fixed

PropertyDescriber/PropertyDescriberInterface.php
FilipBenco: Difference is in the `describe` definition. Unfortunately I don't know how should I merge these 2 interfaces. In this case a prefer it to have those interfaces separated.

PropertyDescriber/PropertyDescriberInterface.php
FilipBenco: I was trying to change it chain of responsibility pattern as you described.

In that case, I would create `AbstractHandler`  and maybe some additional handler interface: https://refactoring.guru/design-patterns/chain-of-responsibility/php/example
Also I would have to code method, that will compose whole chain of describers. I would rather do this in separate PR, where we could change it for both Property and Model describers.

PropertyDescriber/ArrayPropertyDescriber.php
FilipBenco: I had to revert this back. Problem is, that ModelDescriber cannot be passed trough Symfony DI container, as it is created during runtime (see `ApiDocGenerator` class)

FilipBenco:
Hi @goetas and @GuilhemN , sorry I didn't have time to get to this sooner. I updated some parts and gave response to other conversations. Please check them and give me some feedback. :) Looking forward to it :)

goetas:
Same here, :+1: 

> The use of a supports method is inspired by the practices in Symfony, I'm not sure it's worth changing it.

agree! I just wanted to point it out.

GuilhemN:
Thank you @FilipBenco :)

FilipBenco:
No problem, Happy to help. When do you plan to release these changes as next version?

GuilhemN:
I just released v3.6.0 :) https://github.com/nelmio/NelmioApiDocBundle/releases/tag/v3.6.0

PR #1577, opened 2019-12-19
fix compatibility with symfony 5

smoench:
FOSRestBundle prevents installing some symfony components in version 5 as it is not compatible yet. I removed the dev requirement locally on FOSRestBundle and got a few incompatibility issues.

-  `Booting the kernel before calling %s() is not supported, the kernel should only be booted once.` on some functional tests
- removed/missing `Symfony\Bundle\FrameworkBundle\Controller\ControllerNameParser`-class

### DIFF
--- a/Tests/Describer/RouteDescriberTest.php
+++ b/Tests/Describer/RouteDescriberTest.php
@@ -41,11 +41,20 @@ class RouteDescriberTest extends AbstractDescriberTest
         $this->routes = new RouteCollection();
         $this->describer = new RouteDescriber(
             $this->routes,
-            new ControllerReflector(
-                new Container(),
-                $this->createMock(ControllerNameParser::class)
-            ),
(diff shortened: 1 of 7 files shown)


maxhelias:
It's look good to me :) 

BTW, I wondering if we shouldn't depreciate the support of FOSRestBundle. The bundle is less and less maintained to favour API-Platform and this would allow this bundle to move forward in a major version without FOSRest. WDYT @GuilhemN ?

GuilhemN:
Thank you @smoench :)

GuilhemN:
@maxhelias about fosrest, except the fact that it doesn't support sf 5 yet, it doesn't constraint us much and its support doesn't cost much so I'd rather keep it for now.

smoench:
@GuilhemN Well, the master build fails ;-) But there you go another PR #1578

PR #1583, opened 2020-01-27
Delete outdated documentation

markuspoerschke:
The CHANGELOG was not updated in the past and therefore it is outdated now.
To avoid confusion about which is the latest version and which changes are made to a version,
the outdated documentation is removed from the repository.


GuilhemN:
It should not be removed imo, it is incomplete but not empty therefore it still is a bit usefull.
I think we should just remove the unreleased comment in it.

PR #1588, opened 2020-02-19
PHP 7.4 is already released

phansys:


maxhelias:
The failure is a false positive caused by fos-rest-bundle which is not yet Symfony 5 compatible

GuilhemN:
Thanks!

Issue #1585, opened 2020-02-05
QueryParam key property not used


lukepass:
Hello, in the `QueryParam` annotation for FOSRestBundle the `key` param is used to change a parameter's name (in query string):

```
* @Rest\QueryParam(name="itemsPerPage", key="items_per_page", requirements="\d+", default=PaginatedResult::DEFAULT_ITEMS_PER_PAGE, description="Number of items per page")
```

Unfortunately Nelmio is not "using" it and in the documentation it shows `itemsPerPage` as parameter name, which is not correct (and it leads to the default being used).

Any way to fix this (even manually in annotations)?

Thanks.

GuilhemN:
Could you review https://github.com/nelmio/NelmioApiDocBundle/pull/1589 please?

lukepass:
It's now working fine. Thanks!

PR #1589, opened 2020-02-21
Use FOSRest *Param::$key when defined

GuilhemN:
Fix https://github.com/nelmio/NelmioApiDocBundle/issues/1585


GuilhemN:
Thanks for the reviews ;)

Issue #1601, opened 2020-03-13
Processing Assert\Choice enums like arrays not objects


katin-dev:
Hello! 
I have a model: 
```php
class User
{
    const STATUS_NAMES = [
        1  => 'active',
        2 => 'blocked',
    ];

    /**
     * @var string Имя
     * @Assert\NotBlank()
     * @Assert\Choice(choices=UserDto::STATUS_NAMES)
     * @Type("string")
     */
    public $status;
}
```

Generating swagger doc by this model results in example value:

```
{
  "status": {
    "1": "active",
    "2": "blocked"
  }
}
```

But must be
```
{
  "status": "active"
}
```

The reason is in  `SymfonyConstraintAnnotationReader`. I think it should use `array_values` when processing `Assert\Choice`.

What do You think?

PR #1602, opened 2020-03-13
Fix generating example for Assert\Choice when choices are object not array

katin-dev:
https://github.com/nelmio/NelmioApiDocBundle/issues/1601


GuilhemN:
Could you add a test case to ensure this works as expected please?

katin-dev:
@GuilhemN I have added the test

GuilhemN:
Thank you @katin-dev! Sorry for the delay...

PR #1617, opened 2020-04-16
Exception message fix for array properties

svetlyi:
I didn't see a way here to get a class and property names, let it be at least title


GuilhemN:
Thanks @svetlyi :)

PR #1619, opened 2020-04-28
Help opcache.preload to load Swagger annotation class

katin-dev:
I faced a problem when turned on opcache.preloader with Symfony 5.0: PHP failed with error about Anonymous class which extends non-preloaded `SWG\Swagger` class.


GuilhemN:
Thank you @katin-dev!

PR #1622, opened 2020-05-07
Add support for JMS Serializer inline property feature

goetas:
https://jmsyst.com/libs/serializer/master/reference/annotations#inline

When this feature is used, properties are declare in the parent class


goetas:
@GuilhemN ping  :)

GuilhemN:
Thank you @goetas! :)

goetas:
😊

PR #1555, opened 2019-11-16
Upgrade to OpenAPI 3

namoscato:
**⚠️ This is currently a work-in-progress.**

Resolves https://github.com/nelmio/NelmioApiDocBundle/issues/1443

## Details

* Upgrades [zircote/swagger-php](https://github.com/zircote/swagger-php) from `^2.0.12` to `^3.0.2`
* Upgrades [api-platform/core](https://github.com/api-platform/core) from `^2.1.0` to `^2.4.2` (for added OpenAPI 3 support in `2.4.0`)

## Next Steps

- [ ] Update todos in `RouteMetadataDescriber::describe`
- [ ] Update controller action annotations / functional tests (ideally this can be split into more explicit, parallel items)
- [ ] Audit OA `UNDEFINED` checks (i.e. `FormModelDescriber::parseForm` todo, `WebTestCase`)

PR #1623, opened 2020-05-07
OpenApi 3 Support

FilipBenco:
This PR adds support for OpenApi 3 and drops Swagger 2. I was inspired by `Util` class created by @calbrecht & initial try at OA3 support made by @namoscato in this PR: #1555

After I failed with picking up work, where @namoscato ended in his PR & also merging nelmo master branch into that namoscatos OA3 branch, I decided to start over from current master branch.

All tests are green. And this branch should generate valid OpenApi 3 specs.

if this PR is merged, that it should be also new Major Version as it completely changes the bundle output.

Even though this update can generate OA3 and pipeline should be green, there are still few tweaks, that I want to implement, before it can be merged.

I'll dig into those this weekend or next week. In the meantime, if you could give me any feedback, what else should I improve so it can be released. I would appreciate it really a lot :)


GuilhemN:
Wow that's really impressive!
Thanks a lot for taking over, this is an enormous job.

I haven't finished reviewing your changes but I wanted to congrats you as soon as possible to let you now I really appreciate your work :)

OpenApiPhp/ModelRegister.php
```diff
@@ -52,31 +55,48 @@ public function __invoke(Analysis $analysis, array $parentGroups = null)
...
+                    case 'json':
+                        $modelAnnotation = new OA\JsonContent($properties);
+
+                        break;
+                    case 'xml':
+                        $modelAnnotation = new OA\XmlContent($properties);
+
+                        break;
+                    default:
+                        throw new \InvalidArgumentException('Unsupported media type');
```
GuilhemN, line 76: This error could be improved a bit to something  like `sprintf("@Model annotation is not compatible with the media type '%s'. It must be one of 'json' or 'xml'.", $this->mediaType)`

Describer/OpenApiPhpDescriber.php
```diff
@@ -0,0 +1,210 @@
...
+                    !$annotation instanceof OA\RequestBody &&
+                    !$annotation instanceof OA\Parameter &&
+                    !$annotation instanceof OA\ExternalDocumentation
+                ) {
+                    throw new \LogicException(sprintf('Using the annotation "%s" as a root annotation in "%s::%s()" is not allowed.', get_class($annotation), $method->getDeclaringClass()->name, $method->name));
+                }
+
+                foreach ($annotation->_unmerged as $unmergedAnnotation) {
+                    if (!$unmergedAnnotation instanceof OA\JsonContent && !$unmergedAnnotation instanceof OA\XmlContent) {
+                        continue;
```
namoscato, line 145: 👀 Interesting; this was definitely something I was missing.

Tests/Functional/Controller/ClassApiController.php
```diff
@@ -23,7 +24,9 @@ class ClassApiController
 {
     /**
      * @Route("/security/class")
-     * @SWG\Response(response="201", description="")
+     * @Operation(
```
smoench, line 27: Is there a reason the response annotation is now surrounded by an operation?

smoench:
This PR would result to a new release, so I would propose removing support for PHP 7.1 and Symfony <4.4. WDYT?

GuilhemN:
Well as long as we don't require new features and it's not a burden to keep supporting them I'm not fund of removing their support.

smoench:
Currently the pipeline with testing Symfony 3.4 is failing. Might be caused by requiring api-platform ^2.5. This could probably decreased to ^2.4 as the OpenApi v3 support was introduced there.

namoscato:
> Currently the pipeline with testing Symfony 3.4 is failing. Might be caused by requiring api-platform ^2.5.

I noticed this as well per https://github.com/nelmio/NelmioApiDocBundle/pull/1555#discussion_r389331228

OpenApiPhp/ModelRegister.php
```diff
@@ -52,31 +55,48 @@ public function __invoke(Analysis $analysis, array $parentGroups = null)
...
             }
 
-            $model = null;
-            foreach ($annotation->_unmerged as $unmerged) {
-                if ($unmerged instanceof ModelAnnotation) {
-                    $model = $unmerged;
-
-                    break;
+            // Implicit usages
+            if ($annotation instanceof OA\Parameter) {
```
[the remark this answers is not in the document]
FilipBenco: I removed the part, that was deprecated. These lines should be able to extract model into parameters or items if used with `ref=@Model` annotation.

Tests/Functional/Controller/ApiController.php
```diff
@@ -29,12 +28,12 @@
 class ApiController
 {
     /**
-     * @SWG\Response(
-     *     response="200",
-     *     description="Success",
-     *     @SWG\Schema(ref=@Model(type=Article::class, groups={"light"}))
+     * @OA\Parameter(ref="#/components/parameters/test"),
```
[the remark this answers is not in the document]
FilipBenco: No, it is irrelevant. I'll switch it back, I have no idea, why I changed this order :D

Tests/Functional/Controller/ApiController.php
```diff
@@ -46,48 +45,42 @@ public function fetchArticleAction()
      *
      * @Route("/swagger", methods={"GET", "LINK"})
      * @Route("/swagger2", methods={"GET"})
-     * @Operation(
```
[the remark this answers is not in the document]
FilipBenco: true that, I'll return it.

Describer/OpenApiPhpDescriber.php
FilipBenco: :) thanks

OpenApiPhp/Util.php
```diff
@@ -0,0 +1,519 @@
...
+ * generated swagger json decoded array.
+ */
+class Util
+{
+    /**
+     * All http method verbs as known by swagger.
+     *
+     * @var array
+     */
+    public static $operations = ['get', 'post', 'put', 'patch', 'delete', 'options', 'head'];
```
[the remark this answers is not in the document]
FilipBenco: I added `trace`, also, I agree, that this array should not be extendable, so I switched it to const.

Tests/Functional/Controller/ClassApiController.php
FilipBenco: nope, this was just my leftover. I switched it back.

FilipBenco:
I updated the api-platform constraint to `^2.4|^2.5`, as it should work with both versions. Regarding dropping support of SF3.4, I'll leave this decision up to maintainers as in our case I haven't made any specific thing to support also SF3.4..

FilipBenco:
I have to take it back. Api Platform v2.4 has few libraries that support only SF4 and above. Because of this I think we will have to drop support for SF3.

GuilhemN:
Ok no problem then, you may remove sf<4.4 support if that's needed :)

OpenApiPhp/ModelRegister.php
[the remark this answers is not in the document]
GuilhemN: Thanks for checking, you are right Indeed :)

composer.json
```diff
@@ -42,7 +42,7 @@
         "doctrine/annotations": "^1.2",
         "doctrine/common": "^2.4",
 
-        "api-platform/core": "^2.1.2",
+        "api-platform/core": "^2.4|^2.5",
```
smoench, line 45: No need for `^2.4|^2.5`. `^2.4` would be enough.

FilipBenco: fixed

FilipBenco:
Ok, so I dropped support for SF3 and now pipeline is green.
From my point of view there are 2 remaining tasks that I need to do:

- [ ]  Refactor how media type is handled, so developers can configure, which media types they want to support in their specs (right now, they can use either json or xml, but it is not yet configurable and they cannot use both at the same time)
- [ ] Update some nelmio docs (of course, if anybody want to help here, i would be happy :D )

Besides these two points, is there anything else that I've missed what I should do?

smoench:
I think this PR will result as a 4.0 release because there are some BC-Breaks.
I would recommend to create a 3.x branch from current master. So this PR could be merged on master and in another PRs we could figure how to resolve the remaining tasks of @FilipBenco has mentioned. Probably a 4.0-Release issue should be created where all remaining tasks could be listed.

grangeway:
Was just having a play to take a stab at the docs for this, should the following not be valid (this is the example in the existing docs):

```
     * @Route("/api/{user}/rewards", methods={"GET"})
     * @OA\Response(
     *     response=200,
     *     description="Returns the rewards of an user",
     *     @OA\Schema(
     *         type="array",
     *         @OA\Items(ref=@Model(type=Reward::class, groups={"full"}))
     *     )
     * )
     * @OA\Parameter(
     *     name="order",
     *     in="query",
     *     description="The field used to order rewards",
     *     @OA\Schema( 
     *       type="string"  
     *	   )
     * )
```

mantis:
Was just having a play with the docs, should the following not be valid (this was my attempt to update the example in the docs):  

```
     * @Route("/api/{user}/rewards", methods={"GET"})
     * @OA\Response(
     *     response=200,
     *     description="Returns the rewards of an user",
     *     @OA\Schema(
     *         type="array",
     *         @OA\Items(ref=@Model(type=Reward::class, groups={"full"}))
     *     )
     * ),
     * @OA\Parameter(
     *     name="order",
     *     in="query",
     *     description="The field used to order rewards",
	 *     @OA\Schema( 
	 *       type="string" 
	 *	   )
     * )
```

GuilhemN:
I updated the docs and added an upgrade guide. In my opinion your PR is ready to merge, the mime type configuration is not paramount for me since people can circumvent the issue by being more explicit (by using `@JsonContent` etc.).

@mantis `@OA\Schema` should be nested in an annotation expliciting the media type used, `@OA\JsonContent` for example.
Maybe we could add it automatically though, I'm not sure whether that would be a good practice, what do you think?

mantis:
I think there's better experts to comment on that then me - i'm a relative newbie in that regard ;).

The page at https://swagger.io/docs/specification/describing-parameters/ gives an example where you can use without (under schema vs content) - the examples i'd seen where from laravel and elsewhere where people tend to be entering OA/GET and wrapping it all.

GuilhemN:
@mantis I meant the `@Schema` under `@Response`, I would update it to:
```
     * @Route("/api/{user}/rewards", methods={"GET"})
     * @OA\Response(
     *     response=200,
     *     description="Returns the rewards of an user",
     *     @OA\JsonContent(@OA\Schema(
     *         type="array",
     *         @OA\Items(ref=@Model(type=Reward::class, groups={"full"}))
     *     ))
     * ),
     * @OA\Parameter(
     *     name="order",
     *     in="query",
     *     description="The field used to order rewards",
	 *     @OA\Schema( 
	 *       type="string" 
	 *	   )
     * )
```

The `@Schema` under `@Parameter` should be fine but I'm no expert either so maybe I'm wrong :)

FilipBenco:
Hi guys, sorry for my latter response, I was a but busy.

I am really happy, that you are helping with the docs!! :)
Regarding using `@OA\Schema` in `@OA\Parameter` is completely legit and you do not have to specify media type at all, as this is just a parameter.
`@OA\MediaType` should be used only in Responses.

Whether we add support for configuring MediaTypes or no is up to you @GuilhemN . I'll take a swing at it today. All I need to know is just the place in docs where we want to put it:
```yaml
media_type:
  - json (default)
  - xml
```
Then I'll just rearrange few constructors, so this settings reach its destination :)

FilipBenco:
I think it might be quite easy to implement, that when using: `@OA\Items(ref=@Model(type=Reward::class, groups={"full"}))` directly inside `@OA\Response` or `@OA\RequestBody`, we will convert it into:
```
     *     @OA\JsonContent(@OA\Schema(
     *         type="array",
     *         @OA\Items(ref=@Model(type=Reward::class, groups={"full"}))
     *     ))
```

This might be actually quite awesome, because it would also take into account configured media types.

What do you think?

mantis:
@FilipBenco @GuilhemN - i've updated to Guilhem's version of the comment i.e. add OA\Jsoncontent and was still getting the same message - so not sure if i'm missing something here

FilipBenco:
@mantis what messege are you getting?

mantis:
ErrorException:
User Warning: Multiple @OA\Schema() with the same schema="DocParser":
  \Doctrine\Common\Annotations\DocParser->Annotation() in E:\nelmio\test\vendor\doctrine\annotations\lib\Doctrine\Common\Annotations\DocParser.php on line 827
  \Doctrine\Common\Annotations\DocParser->Annotation() in E:\nelmio\test\vendor\doctrine\annotations\lib\Doctrine\Common\Annotations\DocParser.php on line 827

  at E:\nelmio\test\vendor\zircote\swagger-php\src\Logger.php:39
  at OpenApi\Logger->OpenApi\{closure}()
  at call_user_func()
     (E:\nelmio\test\vendor\zircote\swagger-php\src\Logger.php:61)
  at OpenApi\Logger::warning()
     (E:\nelmio\test\vendor\zircote\swagger-php\src\Annotations\AbstractAnnotation.php:395)
  at OpenApi\Annotations\AbstractAnnotation->validate()
     (E:\nelmio\test\vendor\zircote\swagger-php\src\Annotations\AbstractAnnotation.php:481)
  at OpenApi\Annotations\AbstractAnnotation::_validate()
     (E:\nelmio\test\vendor\zircote\swagger-php\src\Annotations\AbstractAnnotation.php:451)
  at OpenApi\Annotations\AbstractAnnotation->validate()
     (E:\nelmio\test\vendor\zircote\swagger-php\src\Annotations\OpenApi.php:135)
  at OpenApi\Annotations\OpenApi->validate()
     (E:\nelmio\test\vendor\zircote\swagger-php\src\Analysis.php:487)
  at OpenApi\Analysis->validate()
     (E:\nelmio\test\vendor\nelmio\api-doc-bundle\Describer\OpenApiPhpDescriber.php:53)
  at Nelmio\ApiDocBundle\Describer\OpenApiPhpDescriber->describe()
     (E:\nelmio\test\vendor\nelmio\api-doc-bundle\ApiDocGenerator.php:78)
  at Nelmio\ApiDocBundle\ApiDocGenerator->generate()
     (E:\nelmio\test\vendor\nelmio\api-doc-bundle\Controller\DocumentationController.php:53)
  at Nelmio\ApiDocBundle\Controller\DocumentationController->__invoke()
     (E:\nelmio\test\vendor\symfony\http-kernel\HttpKernel.php:157)
  at Symfony\Component\HttpKernel\HttpKernel->handleRaw()
     (E:\nelmio\test\vendor\symfony\http-kernel\HttpKernel.php:79)
  at Symfony\Component\HttpKernel\HttpKernel->handle()
     (E:\nelmio\test\vendor\symfony\http-kernel\Kernel.php:191)
  at Symfony\Component\HttpKernel\Kernel->handle()
     (E:\nelmio\test\public\index.php:25)

Controller:

<?php

namespace App\Controller;

use Symfony\Component\HttpFoundation\JsonResponse;
use Symfony\Component\Routing\Annotation\Route;
use Nelmio\ApiDocBundle\Annotation\Model;
use Nelmio\ApiDocBundle\Annotation\Security;

/**
 * Admin / System Utilities Controller
 *
 * @Route("/api/system")
 */
class AdminController  {

/**
     * List the rewards of the specified user.
     *
     * This call takes into account all confirmed awards, but not pending or refused awards.
     *
     * @Route("/api/{user}/rewards", methods={"GET"})
     * @OA\Response(
     *     response=200,
     *     description="Returns the rewards of an user",
     *     @OA\JsonContent(@OA\Schema(
     *         type="array",
     *         @OA\Items(ref=@Model(type=Reward::class, groups={"full"}))
     *     ))
     * ),
     * @OA\Parameter(
     *     name="order",
     *     in="query",
     *     description="The field used to order rewards",
	 *     @OA\Schema( 
	 *       type="string" 
	 *	   )
     * )
	 *
     * @OA\Tag(name="rewards")
     * @Security(name="Bearer")
	 * 
     */
    public function status()
    {


        return new JsonResponse([]);
    }

}

class Reward {
	public string $bar;
}

FilipBenco:
Oh, I found out where is the problem. Just leave out the @OA\Schema in JsonContent. I have exactly this case in the functional tests and here it is working correctly. 

This should work:
```
     *     @OA\JsonContent(
     *         type="array",
     *         @OA\Items(ref=@Model(type=Reward::class, groups={"full"}))
     *     )
```

FilipBenco:
So I added new configuration property: `media_types` which is an array, that can contain `json` or `xml` or both.

I checked also updated docs, & they seems fine.

So, I removed WIP flag as I think this PR can be merged to master :)

mantis:
@FilipBenco not sure if you saw, but i sent a possible PR to your repo - one fix to the docs that was missed, and my suggestion of a reword. Not sure how one is supposed to colloborate on a PR ;)

FilipBenco:
sorry @mantis somehow I overlooked that PR :) I just merged it. Thanks for your help :)

Resources/doc/index.rst
```diff
@@ -186,7 +191,7 @@ This annotation has two options:
 * ``type`` to specify your model's type::
 
     /**
-     * @SWG\Response(
+     * @OA\Response(
      *     response=200,
      *     @Model(type=User::class)
```
BafS, line 196:
```suggestion
     *     ref=@Model(type=User::class)
```

Don't you need `ref` ?

GuilhemN: Thanks for pointing that out! I actually believe there was a change that should not have been done here, in my opinion using the `$ref` field `@Response` doesn't make sense semantically as it is supposed to be a ref to a `response` object, I think we should still use `@Model` directly under `@Response`s.
I pushed a commit to reflect my vision, what do you think a bout it?

GuilhemN:
> I think it might be quite easy to implement, that when using: `@OA\Items(ref=@Model(type=Reward::class, groups={"full"}))` directly inside `@OA\Response` or `@OA\RequestBody`, we will convert it into:
> 
> ```
>      *     @OA\JsonContent(@OA\Schema(
>      *         type="array",
>      *         @OA\Items(ref=@Model(type=Reward::class, groups={"full"}))
>      *     ))
> ```
> 
> This might be actually quite awesome, because it would also take into account configured media types.
> 
> What do you think?

That could be a nice shortcut, but then we should also do it for `@Schema` in my opinion (i.e. `@Response(@Schema(...))` could be automatically translated into `@Response(@JsonContent(...))`.

However, I don't think it should be done in this PR, it is already quite big and this might create unwanted behaviors so we should take some time to think carefully about it.

Resources/doc/index.rst
BafS: I think it's great yes and consistent 👍 thanks

GuilhemN:
And it's merged! :smiley: 

Thanks a lot again @FilipBenco, that was really a huge work.
A big thank you to @calbrecht and @namoscato as well! And thank you to @smoench, @mantis and @BafS for your reviews :D

It took quite some time (and rightfully regarding the amount of changes required) but here we are! I'll release a 4.0 beta very soon, please give it a try if you can :)

mantis:
argh, @FilipBenco @GuilhemN 

If nelmio_api_doc.yaml does not have a components: section, then RouteMetaDataDescribe.php - line 33 tries to call ->parameters on $api->components - where the components is defined as OA\Undefined which is a string.

That probably needs fixing, also as I might try and tweak the upgrade guide to expand on the components section a bit more, as I suspect it will be a common issue (e.g. SecurityDefinitions is not components -> SecuritySchemes

mantis:
The order in AnnotationsReader for updating properties is wrong: 

The following leads to an error about multiple properties (title):

   /**
     * Custom Field Name
     *
     * @OA\Property(
     *     title = "Field Name",
     *     description = "Descriptive name given to the Custom Field by the user",
     * )

For description/title, I think it's just case of swapping the order in updateProperty - not sure about summary.

PR #1629, opened 2020-05-28
Fix undefined function in OA\Schema

BafS:
`$property->getTitle()` should be `$property->title`


GuilhemN:
Good catch! Thanks @BafS :)

PR #1633, opened 2020-05-29
Fix undefined property `$this->mediaType`

BafS:
Fix undefined property `$this->mediaType`, should be mediaTypes


GuilhemN:
Thank you! :)

PR #1634, opened 2020-05-29
Add return type hints

BafS:
Add some return type hints

### DIFF
--- a/ApiDocGenerator.php
+++ b/ApiDocGenerator.php
@@ -50,7 +50,7 @@ final class ApiDocGenerator
         $this->cacheItemId = $cacheItemId;
     }
 
-    public function setAlternativeNames(array $alternativeNames)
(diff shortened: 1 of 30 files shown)


GuilhemN:
I don't think we should update our public signatures, this would create additional BC breaks that may harden the upgrade to v4 while not bringing huge benefits imo.

BafS:
Okay, let's close this PR

GuilhemN:
Thanks for the submission still :)

BafS:
You're welcome !
Off topic but instead of creating a new issue; what do you think about having Redoc (https://github.com/Redocly/redoc) as frontend ? Or both options (redoc is basically a single js file)

GuilhemN:
Sure, that would be nice to have both options :+1: For reference, it seems ApiPlatform added its support a while ago now: https://github.com/api-platform/core/pull/2384.

Issue #1348, opened 2018-06-22
No documentation generated for non persisted column


darame07:
I have an issue when I try to format my response doc with a model class but with one non persisted column (expiresAt)

`
@SWG\Response(
     *         response="200",
     *         @Model(type=AuthToken::class, groups={"auth-token"}),
     *         description="Returned when successful"
     *     )
`

`
    /**
     * @ORM\Column(type="string")
     * @SWG\Property(description="Token d'authentification")
     */
    protected $value;

    /**
     * @ORM\Column(type="datetime")
     * @var \DateTime
     * @SWG\Property(description="Date de création de token")
     */
    protected $createdAt;

    /**
     * @var \DateTime
     * @SWG\Property(description="Date d'expiration de token")
     */
    protected $expiresAt;

    /**
     * @ORM\Column(type="string")
     * @var string
     * @SWG\Property(description="Login de l'annuaire LDAP")
     */
    protected $uid;
`

It generate this following response
`
{
  "value": "string",
  "createdAt": "2018-06-22T09:46:06.923Z",
  "uid": "string"
}
`

Is there any way to add the expiresAt column keeping the @Model annotation ?

Thanks in advance

GuilhemN:
I believe that's an issue in sf property info component, not sure what we could do here to fix it. Imo the issue is that the `DoctrineExtractor` implements `getProperties`, it shouldn't be the case imo.

@dunglas, do you know if this extractor implements `getProperties` for a good reason?

dunglas:
Because sometimes only this one is available (and it is fast).
For this specific issue, you can change the priority of the `ReflectionExtractor` to be executed before the Doctrine one when extracting the properties list.

magnetik:
This might be related: it seems that doctrine Embedded fields are not working at the moment.

Should I open another bug?

magnetik:
I knew this reminded me of something: https://github.com/symfony/symfony/issues/21367 

Maybe we should chain the extractors and stop when enough information have been found?

GuilhemN:
> This might be related: it seems that doctrine Embedded fields are not working at the moment.

@magnetik What are you referring to exactly? `@OneToOne`, `@OneToMany`, ... relations?

> I knew this reminded me of something: symfony/symfony#21367

Indeed, it's a similar issue. Maybe the `DoctrineExtractor::getProperties()` method should simply be removed after all, there is already a cache layer regarding performance issues.

[...]

PR #1639, opened 2020-05-31
Fix #1616 - property describing doctrine error

### LINKED ISSUE #1616: Bad mapping with @var with annotation array @ORM\Column(type="array", nullable=true)
Hello,

i get error when i use 

 ```
/**
     * @var string[]
     * @ORM\Column(type="array", nullable=true)
     */
    private $alerts;
```

But like this work
```
/**
     * @var string[]
     */
    private $alerts;
```
For fix this i have to use swagger property but i would like use @var is more simple. This feature exist?

FilipBenco:
PR Fixes #1616 by changing priority of php doc type_extractor. I added also test, which should validate this fix

### DIFF
--- a/DependencyInjection/NelmioApiDocExtension.php
+++ b/DependencyInjection/NelmioApiDocExtension.php
@@ -139,6 +139,10 @@ final class NelmioApiDocExtension extends Extension implements PrependExtensionI
         $container->getDefinition('nelmio_api_doc.model_describers.object')
             ->setArgument(3, $config['media_types']);
 
+        // Make PhpDoc Property Type Extractor take precedence over Doctrine
+        $container->getDefinition('property_info.php_doc_extractor')
+            ->addTag('property_info.type_extractor', ['priority' => -998]);
+
         // Import services needed for each library
         $loader->load('php_doc.xml');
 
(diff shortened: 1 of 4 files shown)


GuilhemN:
I've done some research to find where this extractor was registered and why this priority was used. For the reference, it was introduced in https://github.com/doctrine/DoctrineBundle/pull/502 but there was no debate about the priorities. I believe the reason is the same as given https://github.com/nelmio/NelmioApiDocBundle/issues/1348, it probably was done with regards to performance but it should be mitigated by now as the PropertyInfo caches results.

To me, the [DoctrineExtractor](https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/Doctrine/PropertyInfo/DoctrineExtractor.php) is never more explicit than the PhpDocExtractor so this change should be fine.

However I'm not fund of changing the priorities of services that do not belong to us, imo this should be done directly in the concerned libraries (either in [Symfony](https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L1559) or in [DoctrineBundle](https://github.com/dunglas/DoctrineBundle/blob/fee8d3b26851e82015b573333a335c2b1b4a191c/DependencyInjection/DoctrineExtension.php#L788)) though due to the priorities choosen in Symfony this change can't be done in DoctrineBundle... so it can only be done in Symfony itself.

PR #1651, opened 2020-06-04
Add support for compound properties

FilipBenco:
This PR adds support for Compound types like: `@var Object1|Object2` in Object properties.
It leverages OA3 `oneOf` polymophism feature

### DIFF
--- a/ModelDescriber/ObjectModelDescriber.php
+++ b/ModelDescriber/ObjectModelDescriber.php
@@ -43,7 +43,7 @@ class ObjectModelDescriber implements ModelDescriberInterface, ModelRegistryAwar
     public function __construct(
         PropertyInfoExtractorInterface $propertyInfo,
         Reader $reader,
-        $propertyDescribers,
+        iterable $propertyDescribers,
         array $mediaTypes,
         NameConverterInterface $nameConverter = null
     ) {
@@ -99,7 +99,7 @@ class ObjectModelDescriber implements ModelDescriberInterface, ModelRegistryAwar
 
             $types = $this->propertyInfo->getTypes($class, $propertyName);
             if (null === $types || 0 === count($types)) {
-                throw new \LogicException(sprintf('The PropertyInfo component was not able to guess the type of %s::$%s. You may need to add a `@var` annotation or use `@SWG\Property(type="")` to make its type explicit.', $class, $propertyName));
+                throw new \LogicException(sprintf('The PropertyInfo component was not able to guess the type of %s::$%s. You may need to add a `@var` annotation or use `@OA\Property(type="")` to make its type explicit.', $class, $propertyName));
             }
 
             $this->describeProperty($types, $model, $property, $propertyName);
--- a/ModelDescriber/ObjectModelDescriber.php
+++ b/ModelDescriber/ObjectModelDescriber.php
@@ -58,6 +58,10 @@ class ObjectModelDescriber implements ModelDescriberInterface, ModelRegistryAwar
     {
         $schema->type = 'object';
 
+        if ($model->getType()->isNullable()) {
+            $schema->nullable = true;
+        }
+
         $class = $model->getType()->getClassName();
         $schema->_context->class = $class;
 
@@ -118,7 +122,7 @@ class ObjectModelDescriber implements ModelDescriberInterface, ModelRegistryAwar
             if ($propertyDescriber->supports($types)) {
                 $propertyDescriber->describe($types, $property, $model->getGroups());
 
-                if (1 === count($types) && $types[0]->isNullable()) {
+                if (1 === count($types) && $types[0]->isNullable() && Type::BUILTIN_TYPE_OBJECT !== $types[0]->getBuiltinType()) {
                     $property->nullable = true;
                 }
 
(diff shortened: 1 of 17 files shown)


PropertyDescriber/PropertyDescriberInterface.php
```diff
@@ -16,7 +16,17 @@
 
 interface PropertyDescriberInterface
 {
-    public function describe(Type $type, Schema $property, array $groups = null);
+    /**
+     * @param Type[] $types
+     *
+     * @return bool
```
GuilhemN, line 22: It returns nothing actually :)

PropertyDescriber/CompoundPropertyDescriber.php
```diff
@@ -0,0 +1,65 @@
...
+        if (Type::BUILTIN_TYPE_ARRAY === $types[0]->getBuiltinType()) {
+            $property->type = 'array';
+            $property = Util::getChild($property, OA\Items::class);
+        }
+
+        $property->oneOf = OA\UNDEFINED !== $property->oneOf ? $property->oneOf : [];
+
+        foreach ($types as $type) {
+            $ref = $this->modelRegistry->register(new Model(
+                new Type($types[0]->getBuiltinType(), false, $type->getClassName(), $type->isCollection(), $type->getCollectionKeyType(), $type->getCollectionValueType()),
```
GuilhemN, line 36: why not calling the property describers here with `$type`?

GuilhemN:
I pushed a commit fixing my second comment and adding a test case, hope it's ok for you :)

FilipBenco:
Oh that is great. Thanks for help!

FilipBenco:
Is there anything else that I should do in this PR? :)

GuilhemN:
It looks good to me, I'll do a final review during the weekend just to be sure :)

FilipBenco:
@GuilhemN I accidentally pushed here a feature that will allow `nullable` properties in models. Should I do it in another PR or can I leave it here? These are only 3 lines of code in ObjectModelDescriber.

ModelDescriber/ObjectModelDescriber.php
```diff
@@ -99,31 +103,34 @@ public function describe(Model $model, OA\Schema $schema)
...
         foreach ($this->propertyDescribers as $propertyDescriber) {
             if ($propertyDescriber instanceof ModelRegistryAwareInterface) {
                 $propertyDescriber->setModelRegistry($this->modelRegistry);
             }
-            if ($propertyDescriber->supports($type)) {
-                $propertyDescriber->describe($type, $property, $model->getGroups());
+            if ($propertyDescriber->supports($types)) {
+                $propertyDescriber->describe($types, $property, $model->getGroups());
+
+                if (1 === count($types) && $types[0]->isNullable() && Type::BUILTIN_TYPE_OBJECT !== $types[0]->getBuiltinType()) {
```
GuilhemN, line 125: I'd rather check the content of `$property->ref`, what do you think?
Besides, I don't think we need the check on the number of types.

GuilhemN: In fact I believe this should be done in a property describer with high priority, that would wrap the other describers (this way nullable would also be supported in compound types or sub types).

FilipBenco: You are correct. And also I realized, that how to handle nullable is specific to each property type. So I refactored it to `NullablePropertyTrait`, that can be used for properties, where you need it. What do you think?

FilipBenco:
Sorry to bother you @GuilhemN , would you have time to look at this? As usually, I would like to get rid of my fork :D

PropertyDescriber/ObjectPropertyDescriber.php
```diff
@@ -21,15 +21,13 @@ class ObjectPropertyDescriber implements PropertyDescriberInterface, ModelRegist
 {
     use ModelRegistryAwareTrait;
 
-    public function describe(Type $type, OA\Schema $property, array $groups = null)
+    public function describe(array $types, OA\Schema $property, array $groups = null)
     {
-        $type = new Type($type->getBuiltinType(), false, $type->getClassName(), $type->isCollection(), $type->getCollectionKeyType(), $type->getCollectionValueType()); // ignore nullable field
```
GuilhemN, line 26: Oops I remember now why this was there.
It is here to avoid avoid having 2 definitions, one for `MyClass` and one for my `MyClass|null`.

Couldn't we instead do something like:
```
oneof:
    - $ref: blabla
nullable: true
```

what do you think? not sure that would be a valid specification though :/

GuilhemN:
> Sorry to bother you @GuilhemN , would you have time to look at this? As usually, I would like to get rid of my fork :D

Sure, sorry I'm quite busy these days :sweat_smile:

PropertyDescriber/ObjectPropertyDescriber.php
FilipBenco: Good idea. I played with it a little and when using allOf, then also generated clients where able to generate very good SDKs :)

FilipBenco:
> > Sorry to bother you @GuilhemN , would you have time to look at this? As usually, I would like to get rid of my fork :D
> 
> Sure, sorry I'm quite busy these days 😅

Yeah :) I know what you mean :D I updated this PR to handle Nullable references better. Thanks for the suggestion :)

GuilhemN:
Awesome! Thanks a lot :D

PR #1657, opened 2020-06-11
Change the Expression's pattern from the message to the expression itself

### LINKED ISSUE #1641: Expression message in pattern
Hi guys,

I was implementing an API with a DTO, and I was using the Symfony's `Expression`  assertion. So I implemented the expression and the error message... Until I notice that the error message is displayed as a `pattern` in Nelmio.

So the property looks like this
image

And the result is that
image

I don't understand why this was implemented in the first place. For me, the message from the Expression is an error message where the pattern is a description of the expected input.

This is where the pattern is added: https://github.com/nelmio/NelmioApiDocBundle/blob/master/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php#L72

Thanks in advance for the answers :)

NicolasGuilloux:
### Fix #1641

I am still doubtful about the pattern. Is the expression maybe a little bit too technical to be displayed as it is in the documentation?

Or maybe it would be worth to create a configuration enable/disable annotations to let the user choose what should be displayed?

### DIFF
--- a/Tests/Functional/FunctionalTest.php
+++ b/Tests/Functional/FunctionalTest.php
@@ -371,7 +371,7 @@ class FunctionalTest extends WebTestCase
                 ],
                 'propertyExpression' => [
                     'type' => 'integer',
-                    'pattern' => 'If this is a tech post, the category should be either php or symfony!',
+                    'pattern' => 'this.getCategory() in [\'php\', \'symfony\'] or !this.isTechnicalPost()',
                 ],
                 'propertyRange' => [
                     'type' => 'integer',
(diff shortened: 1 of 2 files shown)


pratyam:
like a I say in my comment: https://github.com/nelmio/NelmioApiDocBundle/issues/1641#issuecomment-642969881

It doesn't fix the problem, in `pattern` you can only write a regex. The fix is do nothing or append the message to the description

Tests/Functional/FunctionalTest.php
```diff
@@ -371,7 +371,7 @@ public function testSymfonyConstraintDocumentation()
                 ],
                 'propertyExpression' => [
                     'type' => 'integer',
-                    'pattern' => 'If this is a tech post, the category should be either php or symfony!',
+                    'pattern' => 'this.getCategory() in [\'php\', \'symfony\'] or !this.isTechnicalPost()',
```
pratyam, line 374: showing symfony code inside a swagger documentation only makes things worse

NicolasGuilloux: The expression shouldn't append its message nor expression to the pattern, as the documentation states that the pattern field must only be Javascript regex.

This will not be corrected from this PR as it differs from the original topic. It should be the subject of another PR.

Discussion over here: https://github.com/nelmio/NelmioApiDocBundle/issues/1641

GuilhemN:
Closing in favor of #1670, thanks!

PR #1667, opened 2020-06-21
remove pattern added from the Expression Violation message.

pratyam:
This error message inside the pattern, confuses the API client showing a violation message instead of having a Regex. Any informatory message for the client should be placed in "description"


GuilhemN:
Thank you for the submission!

Could you target 3.x instead of master and add an entry to the CHANGELOG please? :)

pratyam:
> Thank you for the submission!
> 
> Could you target 3.x instead of master and add an entry to the CHANGELOG please? :)

I could a separate PR instead: https://github.com/nelmio/NelmioApiDocBundle/pull/1670

GuilhemN:
Thanks! closing in favor of #1670

PR #1670, opened 2020-06-22
Bugfix/remove pattern from property expression message error

pratyam:


GuilhemN:
Nice, thank you @pratyam!

PR #1681, opened 2020-07-12
Register the fallback object model describer only when using jms

GuilhemN:
I believe this should have been done when this class was introduced as this may hide issues when using the symfony serializer.

@goetas if you want to have a look at this :)


goetas:
Can you please explain the impact of this change? What was the propose of a catch-all model describer? (From what I remember, JMS will also catch any type for which is able to infer metadata)

GuilhemN:
oops in fact this PR is not useful, it seems you already dealt with this in https://github.com/nelmio/NelmioApiDocBundle/pull/1495#discussion_r275228740.

The catch-all model describer was introduced by you in order to avoid hard failures when using custom handlers :)

goetas:
:+1:

Issue #1412, opened 2018-10-03
[JMSModelDescriber] Title and Description missing on virtual properties.


cyberemissary:
When using VirtualProperty the title and description do not get populated in documentation:

![image](https://user-images.githubusercontent.com/25044236/46429887-86867700-c715-11e8-8662-7540ae49206f.png)

Results in:

![image](https://user-images.githubusercontent.com/25044236/46429023-7077b700-c713-11e8-8bb4-94954eb9c05a.png)

A current workaround is to create an actual property which will not be used and define its accessor to be the method used for virtual property like so:

![image](https://user-images.githubusercontent.com/25044236/46433034-d36e4b80-c71d-11e8-9524-ee166aa33da9.png)

Resulting in:

![image](https://user-images.githubusercontent.com/25044236/46429361-3955d580-c714-11e8-8847-493e17cdf501.png)

I have narrowed it down to this section of the code in `Nelmio\ApiDocBundle\ModelDescriber\JMSModelDescriber`

![image](https://user-images.githubusercontent.com/25044236/46429466-7cb04400-c714-11e8-8ed7-ef3053401b62.png)

In case of VirtualPropertyMetadata, the `$item->reflection` is null so `$annotationsReader->updateProperty()` (which parses the title/description) is never called. 

It should be possible to use method reflection to load this info.

goetas:
@GuilhemN I would like to start working on this but all the 
ModelDescriber/Annotations/SwgAnnotationsReader.php
ModelDescriber/Annotations/AnnotationsReader.php
ModelDescriber/Annotations/PropertyPhpDocReader.php

Always try to extract metadata from `ReflectionPropery`, but in this case we have `ReflectionMethod`, so the method signature do not allow to use them for methods but only for properties.

Are you ok if I duplicate the methods to support `ReflectionMethod`? or you prefer to remove the signatures from the exiting methods?

(I'm not happy with code duplication, but IMO we should duplicate the methods and handle `ReflectionMethod` independently)

GuilhemN:
It looks overkill to me to duplicate all these methods, I think removing the signature would be fine (and call the correct doctrine reader method).
Why do you think we need separate methods to handle `ReflectionMethod`?

goetas:
The names are also `updateProperty` or `getPropertyName` that gives me the idea of "properties" instead of methods.

[...]

PR #1682, opened 2020-07-18
Add annotations support at methods level when using the JMS serializer

GuilhemN:
Fixes https://github.com/nelmio/NelmioApiDocBundle/issues/1412


goetas:
Whoa! This is huge!

GuilhemN:
Thanks! Next step could be to leverage readOnly/write fields since they are more likely to be useful when using virtual properties :smile: 

Besides, do you think I should check `\property_exists($item, 'getter')` (or `$item instanceof JMS\Serializer\Metadata\PropertyMetadata`) since it is not enforced by the `ClassMetadata` typing?

goetas:
I think that doing instance of property metadata is good enough. Almost all jms code relies on that property being that class

GuilhemN:
Ok, thanks, let's merge it as is then :)

Issue #1676, opened 2020-07-08
UI/Bug After installing master branch


xpresas:
On master branch there is a bug where clicking on one API endpoint in the UI it opens up all of them.
In HTML all classes become ```opblock opblock-post is-open``` instead of only the one I just clicked.

GuilhemN:
Indeed, I was able to reproduce your issue.
Upgrading swagger ui doesn't fix this issue so I'm not sure whether this is a bug on their side or on ours...
The only solution I see is reverting the last commit bumping swagger-ui.

charliefancelli:
If we look on the html, the `id` is the same for all endpoint in the same "Category" (tag)

![Capture d’écran 2020-07-20 à 16 57 47](https://user-images.githubusercontent.com/800399/87962380-62955080-cab7-11ea-9937-cc42596cca29.png)

GuilhemN:
@charliefancelli thank you for debugging this!

I don't know why but swagger-php enabled a new processor by default in version 3.x (which automatically operation ids on every operation). I don't like it, and it's not working well with the custom contexts we use in this bundle so I submitted https://github.com/nelmio/NelmioApiDocBundle/pull/1685 to disable it.
Would this fix be acceptable to you?

charliefancelli:
@GuilhemN of course it's ok for me. I also don't understand why swagger-php did that.

PR #1685, opened 2020-07-20
Disable the OperationId processor

GuilhemN:
Fixes https://github.com/nelmio/NelmioApiDocBundle/issues/1676 by disabling the OperationId from zircote/swagger-php (see https://github.com/zircote/swagger-php/blob/91bfae956a5db9f777f5cd091f46c900928fc5e0/src/Processors/OperationId.php).

It's been enabled by default in swagger-php 3.x but it is not working well with the custom contexts we use, so I think it's better to shortcut this processor.

### DIFF
--- a/ApiDocGenerator.php
+++ b/ApiDocGenerator.php
@@ -15,6 +15,7 @@ use Nelmio\ApiDocBundle\Describer\DescriberInterface;
 use Nelmio\ApiDocBundle\Describer\ModelRegistryAwareInterface;
 use Nelmio\ApiDocBundle\Model\ModelRegistry;
 use Nelmio\ApiDocBundle\ModelDescriber\ModelDescriberInterface;
+use Nelmio\ApiDocBundle\OpenApiPhp\DefaultOperationId;
 use Nelmio\ApiDocBundle\OpenApiPhp\ModelRegister;
 use OpenApi\Analysis;
 use OpenApi\Annotations\OpenApi;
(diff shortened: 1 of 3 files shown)


BafS:
Could we have that as an option ? This break all the links in redoc, as it uses `operationId` to create anchors.

GuilhemN:
@BafS well I believe we would need to do more than just adding an option reverting this as the `operationId` was the same for almost all operations and it was completely broken with how we register annotations.

BafS:
Thanks for your answer, if it can help someone I did wrap it like that:

```php
    public function generate(string $area = ''): OpenApi
    {
        $area = $area ? ('.' . $area) : '';

        $serviceName = 'nelmio_api_doc.generator' . $area;

        if (!$this->container->has($serviceName)) {
            throw new ServiceNotFoundException(
                "Service `$serviceName` does not exists, please install nelmio/api-doc-bundle."
            );
        }

        /** @var \Nelmio\ApiDocBundle\ApiDocGenerator $apiDocGenerator */
        $apiDocGenerator = $this->container->get($serviceName);
        $openApi = $apiDocGenerator->generate();

        // https://github.com/nelmio/NelmioApiDocBundle/pull/1685
        $analysis = new Analysis();
        $analysis->addAnnotation($openApi, null);
        $allOperations = $analysis->getAnnotationsOfType(Operation::class);

        foreach ($allOperations as $operation) {
            if (null === $operation->operationId && $operation instanceof Operation) {
                $path = $operation->method . '_' . trim($operation->path ?? '', ' /');
                $operation->operationId = preg_replace('/[^a-zA-Z0-9_\{\}\@]+/', '-', $path);
            }
        }

        return $openApi;
    }
```

tdrbg:
@BafS That seems very helpful thanks but I'm confused at to where to put this in the code?

BafS:
@tdrbg I did my own command to generate the documentation (json/yaml output), this is the main method.

PR #1697, opened 2020-08-05
Fix nullable array properties

FilipBenco:
This PR fixes nullable array property descriptions:

```
/**
 * @var Class[]|null
 */
$property;
```

will be now described also with nullable flag in OA specs.

### DIFF
--- a/Tests/Functional/FunctionalTest.php
+++ b/Tests/Functional/FunctionalTest.php
@@ -195,6 +195,13 @@ class FunctionalTest extends WebTestCase
                             ['$ref' => '#/components/schemas/User'],
                         ],
                     ],
+                    'friends' => [
+                        'nullable' => true,
+                        'items' => [
+                            '$ref' => '#/components/schemas/User',
+                        ],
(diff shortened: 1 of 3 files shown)


FilipBenco:
@GuilhemN I haven't touched style config but it is failing for some reason. Could you please take a look at it? Otherwise, this PR is ready :)

Tests/Functional/FunctionalTest.php
```diff
@@ -460,7 +467,6 @@ public function testInvokableController()
     public function testDefaultOperationId()
     {
         $operation = $this->getOperation('/api/article/{id}', 'get');
-        var_dump($operation->operationId);
```
GuilhemN, line 463: Oh I was wondering where this was coming from :laughing: 
Thanks for catching this mistake!

GuilhemN:
Thank you @FilipBenco :) 

About the StyleCI failure, they changed their default config and thus we need to do some changes, it's not related to your changes there is no issue here.

Issue #1698, opened 2020-08-06
routes does not work when we use php config files instead of yaml for routes in Symfony


Dattaraj-Devne:
We need to use php configs instead of yaml in our Symfony project. So below yaml code is converted to php config file. But it does not work.

**YAML config**
```
app.swagger:
    path: /api/doc.json
    methods: GET
    defaults: { _controller: nelmio_api_doc.controller.swagger }
```

**PHP config**
```    
$routes->add('app.swagger', '/api/doc.json')
        ->controller('nelmio_api_doc.controller.swagger')
        ->methods(['GET']);
```

Above php config in routes.php throws an symfony exception for Type Error.

```Argument 1 passed to Nelmio\ApiDocBundle\Util\ControllerReflector::getReflectionMethod() must be of the type string, array given, called in vendor/nelmio/api-doc-bundle/Describer/SwaggerPhpDescriber.php on line 232```

composer looks like below:
```
"symfony/framework-bundle": "^5.1.3",
"nelmio/api-doc-bundle": "^3.6"
```

adirkuhn:
actually the problem happens when you specify an array 

`    $routes->add('health', '/v1/clients')
        ->controller([ClientsController::class, 'index'])`

GuilhemN:
Fixed in https://github.com/nelmio/NelmioApiDocBundle/pull/1701, thanks for the report!

PR #1699, opened 2020-08-06
Supports reflection for routes configured in PHP files.

adirkuhn:
PR for the issue #1698

By default the PHP configuration routes returns an array with the controller name and method, since the reflection class was typed for string only this was causing an exception when the user was not using YAML or XML configuration.
This changes removes the string type hint from the method and checks if it's an array or string to do the reflection.

### DIFF
--- a/Util/ControllerReflector.php
+++ b/Util/ControllerReflector.php
@@ -40,21 +40,31 @@ class ControllerReflector
      *
      * @return \ReflectionMethod|null
      */
-    public function getReflectionMethod(string $controller)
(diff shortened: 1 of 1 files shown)


GuilhemN:
Could you target branch 3.x instead please? (master is for version 4)
And if you're motivated enough a test case to prevent future regressions would be great :)

adirkuhn:
> Could you target branch 3.x instead please? (master is for version 4)
> And if you're motivated enough a test case to prevent future regressions would be great :)

Ok I need to close this PR and create a new one

PR #1701, opened 2020-08-07
Supports reflection for routes configured in PHP files.

adirkuhn:
PR for issue #1698

By default, the PHP configuration routes return an array with the controller name and method, since the reflection class was typed for string only this was causing an exception when the user was not using YAML or XML configuration.
This change removes the string type hint from the method and checks if it's an array or string to do the reflection.

### DIFF
--- a/Util/ControllerReflector.php
+++ b/Util/ControllerReflector.php
@@ -40,21 +40,31 @@ class ControllerReflector
      *
      * @return \ReflectionMethod|null
      */
-    public function getReflectionMethod(string $controller)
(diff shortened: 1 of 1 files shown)


GuilhemN:
Thank you @adirkuhn! :)

adirkuhn:
do you know the date for the next 3.x release?

GuilhemN:
I released it this morning :) https://github.com/nelmio/NelmioApiDocBundle/releases/tag/v3.7.0

PR #1712, opened 2020-09-01
Fix incorrect media types in FosRestDescriber

spajxo:
Hello, 

i came across this incorrect backslash in `application\json` media type, when i was using FOSRest @ReqeustParam annotation to define request body.


spajxo:
CI fail seems to be not related to my changes. Looks like problem with new version of JMS Serializer

GuilhemN:
Thank you @spajxo!
As you rightfully noted, the CI failure is not related to your changes :)

PR #1718, opened 2020-09-17
Remove the @internal tag to avoid deprecations on symfony 5.1 (#1673)

### LINKED ISSUE #1673: RouteDescriberTrait is now internal, which triggers a deprecation
Since I've migrated to Symfony 5.1, I have the following deprecation on every page : 

```
The "Nelmio\ApiDocBundle\RouteDescriber\RouteDescriberTrait" trait is considered internal. It may change without further notice. You should not use it from "App\UI\Rest\Nelmio\CommandQueryDescriber".
```

As RouteDescriberTrait (just as ModelRegistryAwareTrait) are listed in examples to create your own RouteDescriberInterface implementation, I don't see why one would be set as internal while the other does not (and found no explanation on this repo on why the change happened in the first place).

How should I deal with this ?

rishirajpurohit:
Fix #1673

Looking at the discussion of the issue, this was the proposed fix.

### DIFF
--- a/RouteDescriber/RouteDescriberTrait.php
+++ b/RouteDescriber/RouteDescriberTrait.php
@@ -15,9 +15,6 @@ use EXSyst\Component\Swagger\Operation;
 use EXSyst\Component\Swagger\Swagger;
 use Symfony\Component\Routing\Route;
 
-/**
- * @internal
- */
 trait RouteDescriberTrait
 {
     /**


rishirajpurohit:
@GuilhemN Can you take a look and let me know if there is something I should add/update in this. Thanks.

GuilhemN:
It's all good, thank you @rishirajpurohit! :)

PR #1722, opened 2020-09-28
Exclude development files from dist package

phansys:
Exclude development files from dist package.


GuilhemN:
Great, thank you @phansys! :)

PR #1727, opened 2020-10-02
(feat): add utf-8 charset to response

lucassabreu:
`swagger-ui-bundle` is very sensible to encoding changes because of the
RegExp performed.

ensuring UTF-8 on the response prevents end-user config to break it.

the annotations already needs to be UTF-8 compatible to generate the
JSON, so it should not break users applications.

when i was improving the tests to check the charset i saw that it was already tested using the `content-type` header, but using symfony 3.4.44 with the version 3.7.4 of this bundle the `getCharset` was returning `null`

(i dont know if i should open for this branch... but seemed the better option)

### DIFF
--- a/Controller/SwaggerUiController.php
+++ b/Controller/SwaggerUiController.php
@@ -65,10 +65,12 @@ final class SwaggerUiController
             $spec['basePath'] = $request->getBaseUrl();
         }
 
-        return new Response(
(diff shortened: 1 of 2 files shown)


oAllanWeb (approved):

GuilhemN:
Thank you @lucassabreu!

PR #1735, opened 2020-10-22
Fix Fos rest query param with map

maxperrimond:
I was facing some errors using our swagger in json because of `collectionFormat` with `multi` value is invalid since it's part of Open Api 2 spec.
It comes from `@Rest\QueryParam` with map option which generate this.

### DIFF
--- a/RouteDescriber/FosRestDescriber.php
+++ b/RouteDescriber/FosRestDescriber.php
@@ -58,6 +58,10 @@ final class FosRestDescriber implements RouteDescriberInterface
                         $parameter->description = $annotation->description;
                     }
 
+                    if ($annotation->map) {
+                        $parameter->explode = true;
+                    }
+
                     $schema = Util::getChild($parameter, OA\Schema::class);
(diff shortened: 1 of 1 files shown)


RouteDescriber/FosRestDescriber.php
```diff
@@ -58,6 +58,10 @@ public function describe(OA\OpenApi $api, Route $route, \ReflectionMethod $refle
                         $parameter->description = $annotation->description;
                     }
 
+                    if ($annotation->map) {
+                        $parameter->explode = true;
```
GuilhemN, line 62: Looking at the OpenApi v3 specification it seems this is the default behavior so does having this line really make any difference?

maxperrimond: True, query parameters are with form style by default and so explode is true too. Doesn't hurt to be explicit though

GuilhemN:
Thank you @maxperrimond! :)

Issue #1662, opened 2020-06-18
How to hide (or collapse) Models section?


lukepass:
Hello! I read that it should be possible to hide or collapse the Models section. How can I do it using NelmioApiDocBundle?

I found this small guide here: https://idratherbewriting.com/learnapidoc/pubapis_swagger.html

```
const ui = SwaggerUIBundle({
  url: "openapi_openweathermap.yml",
  dom_id: '#swagger-ui',
  defaultModelsExpandDepth: -1,
  deepLinking: true,
  presets: [
    SwaggerUIBundle.presets.apis,
    SwaggerUIStandalonePreset
  ],
  plugins: [
    SwaggerUIBundle.plugins.DownloadUrl
  ],
  layout: "StandaloneLayout"
})
```

And I read that this feature should now be possible: https://github.com/nelmio/NelmioApiDocBundle/pull/1358 and https://github.com/nelmio/NelmioApiDocBundle/pull/1357

Where I should I put the code or the options? I am just interested in the `defaultModelsExpandDepth: -1` option.

Thanks!

GuilhemN:
Unfortunately we don't have a simple mechanism to extend `init-swagger-ui.js`.

What you can do is overwrite our template (see https://symfony.com/doc/current/bundles/override.html#templates):
```
{# templates/bundles/NelmioApiDocBundle/SwaggerUi/index.html.twig #}
{# the special '!' prefix avoids errors when extending from an overridden template #}
{% extends "@!NelmioApiDoc/SwaggerUi/index.html.twig" %}

{% block swagger_initialization %}
   <script src="{{ asset('path to your own init-swagger-ui.js') }}"></script>
{% endblock %}
```

And copy `init-swagger-ui.js` in your project and update as you wish.

jonmnoj:
I have been trying to use this new configuration option as shown in the docs [here](https://symfony.com/doc/4.x/bundles/NelmioApiDocBundle/customization.html#overwrite-twig-template), however it throws a ReferenceError "loadSwaggerUI is not defined" had something changed with the release of 4.x that means this needs to be done in a different way?

GuilhemN:
Actually https://github.com/nelmio/NelmioApiDocBundle/pull/1737 (which implements `loadSwaggerUI`) is not released yet.
I'll do a release very soon.

jonmnoj:
Ah yes, this was only merged in shortly after what was the most recent release, silly me.
Thanks for the update and your great work on this.

[...]

PR #1737, opened 2020-10-23
Allow to pass options to the constructor of SwaggerUIBundle

magnetik:
Hi,

Related to #1330 and and #1662

Instead of having to rewrite the whole `init-swagger-ui.js`, this allows to use template overriding with like this:

```
{% extends "@!NelmioApiDoc/SwaggerUi/index.html.twig" %}

{% block swagger_initialization %}
    <script type="text/javascript">
        window.onload = loadSwaggerUI({
            defaultModelsExpandDepth: -1,
            deepLinking: true,
        });
    </script>
{% endblock %}
```

It might be a BC break (but maybe we can find a way around it).

### DIFF
--- a/Resources/doc/customization.rst
+++ b/Resources/doc/customization.rst
@@ -3,12 +3,11 @@ Customization
 
 The look and feel of the Swagger UI can be customized.
 
-
 Overwrite Twig Template
 -----------------------
 
 If you want to customize parts of the template, you can create your own Twig template.
-This allows to change the title, the header, add additional or replace existing styles or scripts.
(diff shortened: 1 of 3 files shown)


GuilhemN (approved):
That's wonderful, thank you! And I consider these changes are acceptable in a minor version.
Would you like to document this addition or should I merge as is?

magnetik:
I've added some documentation but no changelog. I let you handle this.

Thanks a lot :)

GuilhemN:
Thank you @magnetik! :)

PR #1741, opened 2020-11-02
collectionFormat is not part of the OA 3.0 specification

estheban:


GuilhemN:
This was fixed in https://github.com/nelmio/NelmioApiDocBundle/pull/1735.
Thanks for the PR still :)

PR #1745, opened 2020-11-19
Add php 8 support

GuilhemN:

PR #1750, opened 2020-11-27
support php 8

Vilsafur:


fabianoroberto:
Duplicate of #1745

GuilhemN:
Thank you @Vilsafur!
We will rather go with #1745, though I need to make sure the failures I get are not critical before merging it, any help checking that is welcome :)

PR #1777, opened 2021-02-01
Add Yaml Documentation

### LINKED ISSUE #1613: How to generate API documentation in yaml swagger format?
In Nelmio UI there is missing some yaml preview of generated API, is it possible to generate documentation in yaml format? Some tools for testing like soap ui are able to import yaml swagger file and then soap ui is able to work with this file and apis. Solution?

saleh199:
Fixes #1613

### DIFF
--- a/Controller/JsonDocumentationController.php
+++ b/Controller/JsonDocumentationController.php
@@ -18,7 +18,7 @@ use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 
-final class DocumentationController
+final class JsonDocumentationController
 {
     private $generatorLocator;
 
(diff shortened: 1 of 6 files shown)


Controller/JsonDocumentationController.php
```diff
@@ -18,7 +18,7 @@
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 
-final class DocumentationController
+final class JsonDocumentationController
```
GuilhemN, line 21: I'd rather avoid this renaming, it is a backward compatibility break that is not necessary and if you really want to do it, I'd rather have it in a separate PR with backward compatibility layers.

GuilhemN (commented):
Thank you for this PR!
I think we should avoid the renaming of the existing controller but except for this, it's a great addition :)

Controller/JsonDocumentationController.php
saleh199: Yes, you are right, I already reverted the change of the controller

GuilhemN:
Thank you very much @saleh199!

Issue #1780, opened 2021-02-10
Length Constraints Don't Account for a `null` Min or Max Value


chrisguitarguy:
So it's possible to get something like this in the OpenAPI schema:

```
"property": {
  "type":"string",
  "maxLength":0,
  "minLength":1
}
```

PR #1781, opened 2021-02-10
Check Min and Max in Length Constraints Before Setting in Schemas

chrisguitarguy:
Should fix https://github.com/nelmio/NelmioApiDocBundle/issues/1780

Sometimes folks will set a `min` length without a max, for instance and
the generated open api schema would previously have been nonsense:

```
"property": {
  "type":"string",
  "maxLength":0,
  "minLength":1
}
```


GuilhemN:
Good catch! Thank you @chrisguitarguy :)

Issue #1783, opened 2021-02-11
Choice Constraints with `multiple` Do Not Apply Enum to Array Items


chrisguitarguy:
Take something like this:

```php
/**
 * @Assert\Choice(multiple=true, choices={"one", "two", "three"})
 * @var string[]
 */
public $property;
```

I'd expect this to be generated in the OpenAPI Doc:

```
"property": {
  "type": "array",
  "items": {
    "type": "string",
    "enum": ["one", "two", "three"]
  }
}
```

Instead the `enum` lands under property instead of `items`:

```
"property": {
  "type": "array",
  "enum": ["one", "two", "three"],
  "items": {
    "type": "string"
  }
}
```

PR #1784, opened 2021-02-11
Apply `enum` from Choice Constraints to Items When Choice is Multiple

chrisguitarguy:
Should fix https://github.com/nelmio/NelmioApiDocBundle/issues/1783 (sick of me yet?! 🙃 )

Otherwise JSON schema like this is generated:

```
"property": {
  "type": "array",
  "enum": ["one", "two", "three"],
  "items": {
    "type": "string"
  }
}
```

With this change, however, this schema is generated:

```
"property": {
  "type": "array",
  "items": {
    "type": "string",
    "enum": ["one", "two", "three"]
  }
}
```

A possible downside here is that the symfony constraint stuff happens
before types are figured out from PHPDoc. So it's _possible_ to end up
with something that won't validated. Take something like this:

```
/**
 * @Assert\Choice(multiple=true, choices={"..."})
 * @var string
 */
```

This would generate:

```
"property": {
  "type": "string",
  "items": {
    "enum": ["..."]
  }
}
```

### DIFF
--- a/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
+++ b/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
@@ -11,9 +11,11 @@
 
 namespace Nelmio\ApiDocBundle\ModelDescriber\Annotations;
 
+use LogicException;
 use Doctrine\Common\Annotations\Reader;
 use OpenApi\Annotations as OA;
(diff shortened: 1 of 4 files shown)


GuilhemN:
Thank you @chrisguitarguy!
(And sorry for the delay, I couldn't find much time to review my github notifications 😅)

But despite the delay, your PRs are greatly appreciated! Thanks again

PR #1785, opened 2021-02-12
Allow for generation of JSON spec from Documentation Controller when injected as a service

### LINKED ISSUE #1782: Allow for injection of DocumentationController into another controller as a service
Hi,

I want to serve the api.json as part of a twig template when a request is made to a certain page in my app. Is there any easy way that this could be done? I couldn't find any documentation on this and most of the services are private. 

If there's no easy way to do this, and you think this feature could also be useful I could add it a helper function to the Documentation Controller (I think that would be the best place for it, but I may be wrong). The Documentation Controller could be injected as a service to any controller then and the spec could be served from any controller.

Thanks!

Edit: To give a better example, I was thinking something like this could be done in the documentation controller:

```php
public function __invoke(Request $request, $area = 'default')
{
    $this->generate($request, $area);
    return new JsonResponse($spec);
}

public function generate(Request $request, $area = 'default') {
    if (!$this->generatorLocator->has($area)) {
        throw new BadRequestHttpException(sprintf('Area "%s" is not supported as it isn\'t defined in config.', $area));
    }

    /** @var OpenApi $spec */
    $spec = $this->generatorLocator->get($area)->generate();

    if ('' !== $request->getBaseUrl()) {
        $spec->servers = [new Server(['url' => $request->getSchemeAndHttpHost().$request->getBaseUrl()])];
    }

    return $spec;
}
```

m-graham:
Hi,

I want to be able to generate the spec from within another controller and include it in a response from that controller. 

Currently, it's pretty easy to add configure the Documentation Controller so that it can be injected as a service:

config/packages/services.yaml:
```yaml
Nelmio\ApiDocBundle\Controller\DocumentationController: '@nelmio_api_doc.controller.swagger'
```

Controller:
```php
...
use Nelmio\ApiDocBundle\Controller\DocumentationController;
...
class SomeGenericController extends AbstractController
{
    public function __construct(
        DocumentationController $documentationController
    ) {
        $this->documentationController = $documentationController;
    }

    public function index(Request $request): Response
    {
        return $this->render(
            'abc/index.html.twig', [
                'spec' => $this->documentationController->generate($request, 'area')
            ]
        );
    }
}
```

I want to do this as I am actually generating code client side using the API spec.

Seeing as there isn't really much of a code change (beyond encapsulating the contents of the __invoke() function in another function I did not write any tests. The current tests that are in place do pass with this change.

I can add documentation if this pull request seems reasonable. 

closes #1782

### DIFF
--- a/Controller/DocumentationController.php
+++ b/Controller/DocumentationController.php
@@ -29,6 +29,11 @@ final class DocumentationController
 
     public function __invoke(Request $request, $area = 'default')
     {
+        return new JsonResponse($this->generate($request, $area));
+    }
+
+    public function generate(Request $request, $area = 'default')
+    {
         if (!$this->generatorLocator->has($area)) {
             throw new BadRequestHttpException(sprintf('Area "%s" is not supported as it isn\'t defined in config.', $area));
         }
@@ -40,6 +45,6 @@ final class DocumentationController
             $spec->servers = [new Server(['url' => $request->getSchemeAndHttpHost().$request->getBaseUrl()])];
         }
 
-        return new JsonResponse($spec);
+        return $spec;
     }
 }


GuilhemN:
I'm not sure I see the benefit of calling this controller instead of either using render in your templates, or calling directly the generator locator, or even simpler the generator of the default area if you're using just one area

Do you have examples of what would be the benefit?

PR #1788, opened 2021-03-11
Fix Open API doc for DateTime requirements

raziel057:
When using the following requierement:

```
@Rest\QueryParam(name="updatedFrom", strict=true, nullable=true, requirements=@Constraints\DateTime("Y-m-d\TH:i:s"))
```

According to the OpenAPI 3.0 documentation https://swagger.io/specification/, 

We should have ``"format": "date-time"`` rather than ``"format": "DateTime"`` in the generated doc

### DIFF
--- a/RouteDescriber/FosRestDescriber.php
+++ b/RouteDescriber/FosRestDescriber.php
@@ -18,6 +18,7 @@ use Nelmio\ApiDocBundle\OpenApiPhp\Util;
 use OpenApi\Annotations as OA;
 use Symfony\Component\Routing\Route;
 use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\Constraints\DateTime;
 use Symfony\Component\Validator\Constraints\Regex;
 
 final class FosRestDescriber implements RouteDescriberInterface
@@ -104,6 +105,11 @@ final class FosRestDescriber implements RouteDescriberInterface
     private function getFormat($requirements)
     {
         if ($requirements instanceof Constraint && !$requirements instanceof Regex) {
+            
+            if ($requirements instanceof DateTime) {
+                return 'date-time';
+            }
+            
             $reflectionClass = new \ReflectionClass($requirements);
 
             return $reflectionClass->getShortName();
--- a/RouteDescriber/FosRestDescriber.php
+++ b/RouteDescriber/FosRestDescriber.php
@@ -107,7 +107,7 @@ final class FosRestDescriber implements RouteDescriberInterface
         if ($requirements instanceof Constraint && !$requirements instanceof Regex) {
             if ($requirements instanceof DateTime) {
                 // As defined per RFC3339
-                if ('Y-m-d\TH:i:s' === $requirements->format) {
+                if ('Y-m-d\TH:i:s' === $requirements->format || 'c' === $requirements->format) {
                     return 'date-time';
                 } elseif ('Y-m-d' === $requirements->format) {
                     return 'date';
--- a/RouteDescriber/FosRestDescriber.php
+++ b/RouteDescriber/FosRestDescriber.php
@@ -107,7 +107,12 @@ final class FosRestDescriber implements RouteDescriberInterface
         if ($requirements instanceof Constraint && !$requirements instanceof Regex) {
             
             if ($requirements instanceof DateTime) {
-                return 'date-time';
+                // As defined per RFC3339
+                if ($requirements->format === 'Y-m-d\TH:i:s') {
+                    return 'date-time';
+                } else if ($requirements->format === 'Y-m-d') {
+                    return 'date';
+                }
             }
             
             $reflectionClass = new \ReflectionClass($requirements);
--- a/RouteDescriber/FosRestDescriber.php
+++ b/RouteDescriber/FosRestDescriber.php
@@ -105,7 +105,6 @@ final class FosRestDescriber implements RouteDescriberInterface
     private function getFormat($requirements)
     {
         if ($requirements instanceof Constraint && !$requirements instanceof Regex) {
-
             if ($requirements instanceof DateTime) {
                 // As defined per RFC3339
                 if ('Y-m-d\TH:i:s' === $requirements->format) {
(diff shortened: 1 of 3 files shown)


GuilhemN:
If we go this way, maybe we should somehow merge the logic with https://github.com/nelmio/NelmioApiDocBundle/blob/master/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php, what do you think?

phansys (commented):
Could you please add a test covering this fix?

RouteDescriber/FosRestDescriber.php
```diff
@@ -104,6 +105,16 @@ private function getPattern($requirements)
...
     {
         if ($requirements instanceof Constraint && !$requirements instanceof Regex) {
+
+            if ($requirements instanceof DateTime) {
+                // As defined per RFC3339
+                if ('Y-m-d\TH:i:s' === $requirements->format) {
+                    return 'date-time';
+                } elseif ('Y-m-d' === $requirements->format) {
+                    return 'date';
+                }
```
phansys, line 115:
```suggestion
                }
                
                if ('Y-m-d' === $requirements->format) {
                    return 'date';
                }
```

phansys (changes requested):

raziel057:
@phansys I just added the unit tests and fixed the CS

Tests/Functional/Controller/FOSRestController.php
```diff
@@ -28,6 +29,8 @@ class FOSRestController
      * @QueryParam(name="mapped", map=true)
      * @RequestParam(name="Barraa", key="bar", requirements="\d+")
      * @RequestParam(name="baz", requirements=@IsTrue)
+     * @RequestParam(name="datetime", requirements=@DateTime("Y-m-d\TH:i:s"))
+     * @RequestParam(name="date", requirements=@DateTime("Y-m-d"))
```
phansys, line 33: Could you please add test cases for other valid PHP date formats, like `"c"`?

raziel057:
@phansys  I just added support for 'c' format

phansys:
> @phansys I just added support for 'c' format

I don't know if we should support these formats (`"c"`, `\DateTimeInterface::ATOM`, etc).
Mainly, my suggestion is about covering the case where the user uses other formats which are valid in PHP, but exposing that not all of them are valid for the OAS schema.
BTW, I'd like also be consistent with the behavior around this between describers.

raziel057:
Maybe I'm wrong but for me if the provided format doesn't match the OA spec we should not set the ``date`` or ``date-time`` format.

raziel057:
> 
> 
> If we go this way, maybe we should somehow merge the logic with https://github.com/nelmio/NelmioApiDocBundle/blob/master/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php, what do you think?

For me this point is not directly related and can be worked on another PR.

raziel057:
When do you think this PR could be merged?

phansys:
>Maybe I'm wrong but for me if the provided format doesn't match the OA spec we should not set the date or date-time format.

I agree with that, is what I tried to say here:
>I don't know if we should support these formats ("c", \DateTimeInterface::ATOM, etc).
Mainly, my suggestion is about covering the case where the user uses other formats which are valid in PHP, but exposing that not all of them are valid for the OAS schema.

It would be nice to have test cases for scenarios where invalid OAS date formats (but valid for PHP) are passed, reflecting how the serialization behaves.

raziel057:
@phansys I just fixed the format that should match according to the RFC3339 and added a new Assert for date with format that is not matching the RFC3339. In that case rather than generating a ``format: DateTime`` we doesn't set any format. At first I had in mind to generate a pattern given the date format but it seems overkill.

RouteDescriber/FosRestDescriber.php
```diff
@@ -104,6 +105,17 @@ private function getPattern($requirements)
     private function getFormat($requirements)
     {
         if ($requirements instanceof Constraint && !$requirements instanceof Regex) {
+            if ($requirements instanceof DateTime) {
+                // As defined per RFC3339
+                if (\DateTime::RFC3339 === $requirements->format || 'c' === $requirements->format) {
```
phansys, line 110:
```suggestion
                if (\DateTimeInterface::RFC3339 === $requirements->format || 'c' === $requirements->format) {
```

RouteDescriber/FosRestDescriber.php
```diff
@@ -104,6 +105,17 @@ private function getPattern($requirements)
...
+                    return 'date-time';
+                } elseif ('Y-m-d' === $requirements->format) {
+                    return 'date';
+                } else {
+                    return null;
+                }
```
phansys, line 116:
```suggestion
                }
                
                if ('Y-m-d' === $requirements->format) {
                    return 'date';
                }

                return null;
```
BTW, I think we should keep the existing behavior if the provided format is not supported or throw an exception in that case.

phansys (changes requested):

RouteDescriber/FosRestDescriber.php
raziel057: \DateTimeInterface::RFC3339 is not supported by PHP 7.1 as you can see here:
https://github.com/nelmio/NelmioApiDocBundle/runs/2186348084

That's why I done the following change here:
https://github.com/nelmio/NelmioApiDocBundle/pull/1788/commits/d6a2d1240a853262a8005f6e4d8533127e1101f7

RouteDescriber/FosRestDescriber.php
raziel057: Whether we add a ``format: DateTime`` or not at all in the resulting OA doc, the Rest Client Generator (.net) will create a string parameter as the DateTime format is unknown. So for me it should not appear as not supported by https://swagger.io/specification/. I'm not for throwing an exception as if someone wants to provide an other PHP supported datetime format it should be possible but just ignored by Open API.

GuilhemN: Throwing an exception would be too bad for DX imo, ignoring the constraint in case we don't know what to do looks like a fine compromise to me.

GuilhemN:
Thank you @raziel057 and @phansys for you review!

PR #1789, opened 2021-03-12
Added support for constraint attributes

derrabus:
Since Symfony 5.2, it is possible to annotate constraints using PHP 8.0 attributes. When I turn a constraint annotation into an attribute however, it becomes invisible to NelmioApiDocBundle. With this PR, I'd like to fix this problem.

### DIFF
--- a/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
+++ b/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
@@ -14,6 +14,7 @@ namespace Nelmio\ApiDocBundle\ModelDescriber\Annotations;
 use Doctrine\Common\Annotations\Reader;
 use Nelmio\ApiDocBundle\OpenApiPhp\Util;
 use OpenApi\Annotations as OA;
+use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\Constraints as Assert;
 
 /**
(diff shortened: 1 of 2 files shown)


ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
```diff
@@ -150,4 +151,26 @@ private function applyEnumFromChoiceConstraint(OA\Schema $property, Assert\Choic
...
         $setEnumOnThis->enum = array_values($enumValues);
     }
+
+    private function getPropertyAnnotations(\ReflectionProperty $reflection): \Traversable
+    {
+        if (\PHP_VERSION_ID >= 80000) {
+            foreach ($reflection->getAttributes(Constraint::class, \ReflectionAttribute::IS_INSTANCEOF) as $attribute) {
+                yield $attribute->newInstance();
+            }
+        }
```
phansys, line 161: IMO, we could extract these lines to a new private method in order to reuse the same logic in `getPropertyAnnotations()` and `getMethodAnnotations()`.

phansys (approved):

ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
derrabus: On second thought, we could have a single method `getAnnotations()` and move the `if` block from `updateProperty()` into the method. I'll try that.

derrabus: Pushed, please have a look.

GuilhemN:
Wonderful, thank you very much @derrabus!

PR #1556, opened 2019-11-19
Use getReflectionMethod instead of getReflactionClassAndMethod

maxhelias:

PR #1792, opened 2021-03-12
Fix calls to nonexistent method `ControllerReflector::getReflectionClassAndMethod()`

phansys:
This issue was introduced in #1556.


maxhelias (approved):
Good catch!

GuilhemN:
Thank you! I'm not fund of mocks partly because of this :(

PR #1795, opened 2021-03-12
Use PHPUnit 9.4

phansys:


phansys:
Closing since currently there is no way to specify a single PHPUnit version for the range PHP 7.1-8.

PR #1801, opened 2021-03-15
Fix detection of what groups to use if object is empty but has discriminator map metadata

pokhodenko:
When describing empty abstract class with JMS Discriminator map describer fails to compute groups correctly.


pokhodenko:
I have logic for handling discriminator map in custom describer. If you are interested I can submit a PR to add this logic to JMSModelDescriber

GuilhemN:
If possible, could you add a small test case to ensure this is working correctly please?

DjordyKoert:

PR #1804, opened 2021-03-30
Display non native form type childs description and title properly

Arakmar:
The form describer wasn't showing description and title for non native form type childs.
I'm really not an OpenApi expert, there is probably a better way to fix that.

Before
image

After
image

I didn't see any tests related to the form describer, do I need to add some ?


GuilhemN:
Thanks for your contribution!

Actually we have some tests for forms in the Functional folder if you'd like to add one (here https://github.com/nelmio/NelmioApiDocBundle/blob/master/Tests/Functional/FunctionalTest.php#L235).
No worry if you'd rather not that's also ok but it's definitely better to have tests for new features.

GuilhemN:
Thank you @Arakmar!

PR #1810, opened 2021-04-19
Allow psr/cache versions 2 and 3

derrabus:
Follows #1790.

This library is able to consume implementations of `psr/cache` versions 2 and 3 as well, so it should not block their installation.


GuilhemN:
Thank you @derrabus!

PR #1820, opened 2021-05-23
Updates documentation to show a side effect of design decision regarding context groups with JMS serializer

vpassapera:
When using serializer context groups in JMS serializer, even if just using the Default group in one place explicitly and leaving it blank in another, they will be treated as different component schemas, even though to the serializer outcome they are the same.

Documenting behavior of decision with side effect.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -256,6 +256,28 @@ General PHP objects
     **If you're using the JMS Serializer**, the metadata of the JMS serializer are used by default to describe your
     models. Additional information is extracted from the PHP doc block comment,
     but the property types must be specified in the JMS annotations.
+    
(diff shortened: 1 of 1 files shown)


Resources/doc/index.rst
```diff
@@ -256,6 +256,28 @@ General PHP objects
...
+         *
+         * @return Collection|ItemResponse[]
+         */
+        public function getItems(): Collection|array
+        {
+            return $this->items;
+        }
+
+    .. code-block
+       @OA\Schema(ref=@Model(type=FeedItemResponse::class, groups=["Default"])),
```
GuilhemN, line 278: For clarity maybe we should use a fully qualified namespace for `FeedItemResponse` here, what do you think?

vpassapera: Sure, i can do that. Will update.

vpassapera:
Updated.

GuilhemN:
Thank you @vpassapera!

PR #1837, opened 2021-06-16
Fix code block not showing

vpassapera:
Fixes codeblock not showing in the docs.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -257,9 +257,10 @@ General PHP objects
     models. Additional information is extracted from the PHP doc block comment,
     but the property types must be specified in the JMS annotations.
     
-    NOTE: If you are using serialization contexts (e.g. Groups) each permutation will be treated as a separate Path. For example if you have the following two variations defined in different places in your code:
+    **If you are using serialization contexts (e.g. Groups)**, each permutation will be treated as a separate Path. For example if you have the following two variations defined in different places in your code:
     
     .. code-block:: php
+
         /**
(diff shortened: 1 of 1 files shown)


Resources/doc/index.rst
```diff
@@ -257,9 +257,10 @@ General PHP objects
     models. Additional information is extracted from the PHP doc block comment,
     but the property types must be specified in the JMS annotations.
     
-    NOTE: If you are using serialization contexts (e.g. Groups) each permutation will be treated as a separate Path. For example if you have the following two variations defined in different places in your code:
+    **If you are using serialization contexts (e.g. Groups)**, each permutation will be treated as a separate Path. For example if you have the following two variations defined in different places in your code:
```
GuilhemN, line 260: I don't think this is relevant, this note only applies when you're using the JMS serializer so it shouldn't appear to be some sort of another case in my opinion.

GuilhemN (commented):
Good catch, thank you for your PR!

Resources/doc/index.rst
vpassapera: It is indeed relevant to jms serializer contexts as far as I know. I can be explicit about that in the note, so I'll go ahead and update it.

GuilhemN:
I merged https://github.com/nelmio/NelmioApiDocBundle/pull/1851 and https://github.com/nelmio/NelmioApiDocBundle/pull/1846 instead as they are more minimalistic.

Thank you for your PR still!

PR #1840, opened 2021-06-22
Improve resolving symfony asserts

kirya-dev:
Resolving types by Type assert

### DIFF
--- a/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
+++ b/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
@@ -44,6 +44,10 @@ class SymfonyConstraintAnnotationReader
      */
     public function updateProperty($reflection, OA\Property $property): void
     {
+        if (null === $this->schema) {
+            throw new \InvalidArgumentException(__CLASS__ . ' working without schema impossible.');
+        }
+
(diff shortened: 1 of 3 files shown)


GuilhemN:
I'm not sure that's really safe, `@Type` can contain a lot of different values (see https://symfony.com/doc/current/reference/constraints/Type.html) and few are compatible with open api.

What about having an allow list to make sure we don't allow incompatible values?

kirya-dev:
> What about having an allow list to make sure we don't allow incompatible values?

Whats are your mean?

```php
class Dto
{
/**
 * @type int|string
 * @Assert\Type(["float", "string"])
 */
 private $value;
}

GuilhemN:
I mean the value wrapped by `@Assert\Type` is not necessarily a valid value for OpenApi so we must be cautious.

For instance, `["float", "string"]` is an accepted value for `@Assert\Type` but it's not for OpenApi which only accepts a string. In case of an array the best we could do in OpenApi would probably just be to use `mixed`.
And `@Assert\Type` can also contain class types which cannot be used directly in the OpenApi `type` field.

kirya-dev:
We can use `oneOf` field?
Im hasnt time to develop this. It have hard logic for me.

DjordyKoert:

PR #1851, opened 2021-07-16
[Doc] Fix some minor RST issues

javiereguiluz:
The real issue is the lack of `php` in the second `.. code-block::`.

The blank line added after the first `.. code-block::` is to keep it consistent with the rest of code blocks.


GuilhemN:
Oh thank you!
I think we should add a linter in the github actions. I am lacking time right now, but in two weeks it should be better so I'll look into it :)

javiereguiluz:
Great. You can take a look at Symfony Docs, because we have a RST Linter run via GitHub actions.

PR #1860, opened 2021-08-12
Always provide a context to swagger-php

DerManoMann:
Upcoming changes in swagger-php will make `context` a required parameter for instantiating `Analysis` and when adding annotations to the created `Analysis` instance.

See https://github.com/zircote/swagger-php/pull/963 for more details.


GuilhemN:
Thank you for the follow up!

PR #1861, opened 2021-08-16
Improve jms testing by checking that collisions are not generated

goetas:
This PR improves the tests checking that no "automatic" names are generated if all the other params (as aliases) are well defined


GuilhemN:
Thank you @goetas!

PR #1862, opened 2021-08-16
Log name collisions

goetas:


Model/ModelRegistry.php
```diff
@@ -115,13 +123,38 @@ private function generateModelName(Model $model): string
         );
         $i = 1;
         while (\in_array($name, $names, true)) {
+            if (isset($this->registeredModelNames[$name])) {
+                $this->logger->warning(sprintf('Can not assign a name for the model, the name "%s" has already been taken.', $name), [
```
GuilhemN, line 127: Maybe this should be an info rather than a warning, what do you think? This is happening quite often so I'm not sure we should warn too hardly the user.

GuilhemN (commented):
Nice! This could indeed be quite useful for debugging purposes :)

Model/ModelRegistry.php
goetas: In my apps i aim to have zero collisions, but i understand that other devs might not be interested. suggestion applied

GuilhemN: I see, that's interesting for your case indeed!

GuilhemN:
Thank you @goetas!

Issue #1856, opened 2021-07-28
Fix Symfony\Component\PropertyInfo\Type::getCollectionValueType() deprecation notice in symfony 5.3


kappa1389:
I get this deprecation notice
Since symfony/property-info 5.3: The "Symfony\Component\PropertyInfo\Type::getCollectionValueType()" method is deprecated, use "getCollectionValueTypes()" instead.

The deprecated method is called in these files:
ObjectPropertyDescriber, ModelRegistry, ArrayPropertyDescriber

FarhadKh:
Also we got this deprecation notice:
Since symfony/property-info 5.3: The "Symfony\Component\PropertyInfo\Type::getCollectionKeyType()" method is deprecated, use "getCollectionKeyTypes()" instead.

GuilhemN:
Closing as it was fixed in https://github.com/nelmio/NelmioApiDocBundle/pull/1863.
This will be released soon.

PR #1863, opened 2021-08-17
Fix Symfony\Component\PropertyInfo\Type::getCollectionValueType() deprecation notice for symfony >=5.3

onatskyy:
fix for https://github.com/nelmio/NelmioApiDocBundle/issues/1856


Model/ModelRegistry.php
```diff
@@ -124,8 +124,8 @@ private function generateModelName(Model $model): string
 
     private function getTypeShortName(Type $type): string
     {
-        if (null !== $type->getCollectionValueType()) {
-            return $this->getTypeShortName($type->getCollectionValueType()).'[]';
+        if (null !== $collectionType = $type->getCollectionValueTypes()[0] ?? null) {
```
GuilhemN, line 127: unfortunately `getCollectionValueTypes` was introduced in Symfony 5.3 and we still support Symfony 4.

So we need to add some backward compatibility layers to support both versions, we could do something like:
```php
method_exists(Type::class, 'getCollectionValueTypes') ? ($type->getCollectionValueTypes()[0] ?? null) :  $type->getCollectionValueType()
```

what do you think? Similar layers should be added in the different places where `getCollectionValueTypes` is needed, and a comment 
```
// BC layer for symfony < 5.3
```
in each place would be great to remember to simplify this code when we remove support for `symfony < 5.3`.

GuilhemN (commented):
Thank you for submitting this PR, it's really great!

Similarly to [what Symfony does](https://symfony.com/doc/current/contributing/code/bc.html), we do our best to keep our code compatible with a wide range of dependency versions when doing changes. In the context of this change, the new method  `getCollectionValueTypes` was introduced in Symfony 5.3 and we still support Symfony 4, so we need to do some more work to support both versions, while fixing these annoying deprecation notices for users :)

Model/ModelRegistry.php
onatskyy: Agree, fixed

PR #1864, opened 2021-08-21
Documentation : Update templates comment path for bundle override

mathieu-ducrot:
## Changelog

Fix the comment path from SwaggerUI to SwaggerUi since case does matter for template override to work.


GuilhemN:
Indeed, thank you @mathieu-ducrot!

Issue #1458, opened 2019-01-02
Only display tags and hide the default section


aguidis:
I really tried many things before creating this issue but I can't figure it out.

I just want to remove the **default** section :

![image](https://user-images.githubusercontent.com/1856652/50606113-ca2f7f00-0ec4-11e9-977e-21904c72e3ce.png)

Here is the nelmio configuration:

```
nelmio_api_doc:
    models: { use_jms: false }
    documentation:
        host: delivery.app/api/internal
        schemes: [https]
        info:
            title: Internal API
            version: 1.0.0
        tags:
            - name: "Authentication"
              description: "Authentication related endpoints"
            - name: "Delivery Note"
              description: "Delivery notes related endpoints"
            - name: "Parcel"
              description: "Parcel related endpoints"
            - name: "User"
              description: "User related endpoints"
    areas: # to filter documented areas
        path_patterns:
             - ^/api(?!/doc/$) # Accepts routes under /api except /api/doc

```

GuilhemN:
Have you tried excluding your internal routes by updating the path patterns?

Bizinovic:
How can you do that ?

GuilhemN:
Check out the last part of the `If you're not using Flex` paragraph of [the docs](https://symfony.com/doc/current/bundles/NelmioApiDocBundle/index.html#installation).

To summarize, you have to update `nelmio_api_doc.areas.path_patterns`.

rendler-denis:
@GuilhemN 
I would like to filter the default path /api/doc to only have my sections without the default section. I've tried with Areas, but then I would have to go to /api/doc/subscriber . Is there any possibility to use the /api/doc to show only my sections? Thanks.

chrisguitarguy:
Related: https://github.com/nelmio/NelmioApiDocBundle/issues/1425

seems like this issue has been resolved via the PR above, though, so closing.

tomazartack:
For me this worked:

nelmio_api_doc:
    documentation:
        info:
            title: Title
            version: 1.0.0
    areas:
        default:
            disable_default_routes: true

PR #1868, opened 2021-08-31
[3.x]: Hide the default section per area with new config parameter

Havrin:
By adding a  new config parameter it is possible to hide the "default" section in your config.

Approach is to not register the RouteDescriber as a tagged service per area.

Fixes: https://github.com/nelmio/NelmioApiDocBundle/issues/1458

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -63,6 +63,7 @@ final class Configuration implements ConfigurationInterface
                                 'with_annotation' => false,
                                 'documentation' => [],
                                 'name_patterns' => [],
+                                'disable_default_routes' => false,
                             ],
                         ]
                     )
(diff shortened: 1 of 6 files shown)


GuilhemN:
Thank you for your proposal!

GuilhemN:
Thanks a lot for your proposal!

However, I'm not sure this is the correct approach, this means we would no longer leverage php annotations, symfony routes, etc. when using this option and that would be quite confusing (as well as suboptimal) in my opinion.

What do you think about the alternative I suggested in https://github.com/nelmio/NelmioApiDocBundle/issues/1458#issuecomment-907269750, about filtering the routes passed to the RouteDescriber depending on the presence of some OA annotations?

Havrin:
@GuilhemN my idea was to prevent generating those routes so it will be faster. But you are right, there can be side effects.  
I will check your comment and change the MR accordingly :)

Havrin:
@GuilhemN in case the notification slipped through: I changed it like discussed in the Bugticket :) let me know if somethings wrong.

roger-codina:
Hi guys, why this feature is not in present in the 4.x versions?

https://github.com/nelmio/NelmioApiDocBundle/blob/v4.6.2/DependencyInjection/Configuration.php#L52

Thanks!

PR #1870, opened 2021-09-06
Add self-describing model describer

tgalopin:
This PR implements self-describing models: models that are able to describe their structure through a static method call.

I already use this technique in one of my projects with league/fractal (https://fractal.thephpleague.com): in Fractal, there is a concept of Transformers, classes that are able to transform a given object into an array. These transformers can be nested and used in collections.

Such transformers carry the output structure of my API: they are dedicated classes that I wish I could use as "models". They do not however match the currently available models describers. Thus I'd like to propose self-describing models: the ability for model classes to implement an interface and describe their structure in a static method:

```php
class SelfDescribingModel implements SelfDescribingModelInterface
{
    public static function describe(Schema $schema)
    {
        $schema->title = 'SelfDescribingTitle';
        $schema->properties = [/* ... */];
        // ...
    }
}
```

### DIFF
--- a/Tests/ModelDescriber/SelfDescribingModelDescriberTest.php
+++ b/Tests/ModelDescriber/SelfDescribingModelDescriberTest.php
@@ -1,0 +1,48 @@
+<?php
+
+/*
+ * This file is part of the NelmioApiDocBundle package.
+ *
+ * (c) Nelmio
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Nelmio\ApiDocBundle\Tests\ModelDescriber;
+
+use Nelmio\ApiDocBundle\Model\Model;
+use Nelmio\ApiDocBundle\ModelDescriber\SelfDescribingModelDescriber;
+use Nelmio\ApiDocBundle\Tests\ModelDescriber\Fixtures\SelfDescribingModel;
+use OpenApi\Annotations\Schema;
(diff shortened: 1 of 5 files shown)


GuilhemN:
I like the idea 👍🏻
It's simple and easy to use, and can be useful for people preferring pure PHP over annotations.

And I guess this method could just return the value from a call to your transformer and everything would be fine, right?

I think I would just pass the `@Model` described as well to this new static method to take into account serialization context, and notably groups.

Tests/ModelDescriber/SelfDescribingModelDescriberTest.php
```diff
@@ -0,0 +1,33 @@
...
+    public function testDescribe()
+    {
+        $describer = new SelfDescribingModelDescriber();
+
+        $model = new Model(new Type('object', false, SelfDescribingModel::class));
+        $schema = new Schema([]);
+
+        $describer->describe($model, $schema);
+        $this->assertSame('SelfDescribingTitle', $schema->title);
+    }
```
chrisguitarguy, line 32: This need a tests for the `supports` method as well -- both tests that pass supports and don't.

chrisguitarguy (changes requested):

tgalopin:
PR rebased and updated

chrisguitarguy (approved):

GuilhemN (approved):

GuilhemN:
All good for me, thank you @tgalopin!

Issue #1880, opened 2021-09-27
RenderOpenApi argument is null


Goazil:
Hello,

starting version 4.6.1.

I have a error message that I get on the `/api/doc.json`

```
Type error: Nelmio\ApiDocBundle\Render\RenderOpenApi::__construct(): Argument #4 must be of type Nelmio\ApiDocBundle\Render\OpenApiRenderer, null given, called in var/cache/test/ContainerR5XqECk/getNelmioApiDoc_RenderDocsService.php on line 25 (Behat\Testwork\Call\Exception\FatalThrowableError)
```

PR #1881, opened 2021-09-28
Fix OpenApiRenderer list could be null

Goazil:
This PR resolve https://github.com/nelmio/NelmioApiDocBundle/issues/1880.

The bug was introduced in this commit: https://github.com/nelmio/NelmioApiDocBundle/pull/1878/commits/118213a4c03436ee7907cc6dcef34fdebe5de17e


umpirsky:
Thanks!

PR #1889, opened 2021-10-19
fix multiple authentication keys processing

hxss:
Allows multiple authentication keys to be stored in a session.


DjordyKoert:

PR #1896, opened 2021-11-03
Proposal: Add ``header_block`` Twig Block

copiri-six:
In attempting to remove the default green header from the API docs (I am integrating the docs into a site that already has a header), I have found that the only possible way to do so is to overwrite the entire ``index.html.twig`` file.  This is because the existing ``{% header %}`` twig tags are placed *inside* the ``<header>`` HTML tags.  So, if one overrides the ``{% header %}`` block, it only adjusts the header content, as opposed to allowing control over the entire header object.

To resolve this problem I considered two methods:

- Move the existing ``{% header %}`` tags _*outside*_ of the ``<header>`` tags.  This would resolve the problem in the simplest fashion, but would be a breaking change for those currently overriding the block.

- Add a new twig block, placing its tags *_outside_* the existing ``<header>`` tags.  This would leave the existing functionality as-is, and would not break any current implementations.

I chose the latter for the reasons specified above, and suggested the name ``{% header_block %}``.  Thanks for considering this proposal.

### DIFF
--- a/Resources/doc/customization.rst
+++ b/Resources/doc/customization.rst
@@ -50,4 +50,4 @@ Just create a file ``templates/bundles/NelmioApiDocBundle/SwaggerUi/index.html.t
         <script type="text/javascript" src="{{ asset('js/custom-request-signer.js') }}"></script>
     {% endblock javascripts %}
 
-You can have a look at the original template to see which blocks can be overridden.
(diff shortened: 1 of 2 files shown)


GuilhemN:
Thank you! That should not be too complicated to maintain and if that can help you, that's welcome :)

PR #1897, opened 2021-11-03
Add Location of Original Template

copiri-six:
This doc page advises that one can `have a look at the original template to see which blocks can be overridden`, but does not say where the original template can be found.  This edit adds that information for clarity.

Note that I haven't run any tests as this is a docs-only change.


GuilhemN:
That's indeed clearer, thank you @copiri-six!

PR #1914, opened 2021-11-30
Upgrade Symfony 6.0 dependencies

### LINKED ISSUE #1913: no support for symphony 6
no support for symphony 6 in composer.json

fabianoroberto:
On this PR:
-----------

* Fix #1913 
* Restrict to only supported Symfony versions https://symfony.com/releases


fabianoroberto:
Check failed because api-platform has not released a version supporting Symfony 6 yet (they are working on this). 
There is an issue for that https://github.com/api-platform/core/issues/4543

dougetovski:
We should also keep in mind that Symfony 6 will define return types. Only updating composer.json to accepts 6.0 without defining types on Symfony methods that NelmioApiDocBundle override may cause problems.
Correct me if i'm wrong.

fabianoroberto:
@dougetovski you're absolutely right, my fault. 
Maybe we could use Rector (https://github.com/rectorphp/rector) to plan the migration. 
I'll try ASAP

composer.json
```diff
@@ -15,41 +15,41 @@
         }
     ],
     "require": {
-        "php": ">=7.1.3",
+        "php": ">=8.0",
```
GuilhemN, line 18: I don't think we should bump the php version yet, PHP 7 is still widely used (see https://blog.packagist.com/php-versions-stats-2021-1-edition/).
Moreover unlike Symfony, I don't have enough resources to keep providing bug fixes to older versions for people still using php 7.x, so I'd rather stick with a constraint allowing to provide bug fixes to most of the users of this bundle with minimum maintenance load.

composer.json
```diff
@@ -15,41 +15,41 @@
...
         "ext-json": "*",
         "doctrine/annotations": "^1.11",
         "psr/cache": "^1.0|^2.0|^3.0",
         "psr/container": "^1.0|^2.0",
         "psr/log": "^1.0|^2.0|^3.0",
-        "symfony/config": "^4.4|^5.0",
```
GuilhemN, line 24: Is it making things harder to keep supporting Symfony 4.4?
If not, I'd rather keep supporting it, the argument is the same as for PHP 7.x, I think we should keep our constraints as loose as possible, both to simplify the maintenance work, and provide bug fixes to more users :)

composer.json
fabianoroberto: Make sense. Maybe could be drop support for php <7.4 (https://www.php.net/supported-versions.php) and symfony <5.3 (https://symfony.com/releases)

composer.json
fabianoroberto: As above, I can reintroduce support for `^4.4|^5.3|^6.0`, but I need revert some useful changes like `str_*` or union types available as of PHP 8.0.0.

Util/ControllerReflector.php
```diff
@@ -20,46 +20,37 @@
  */
 class ControllerReflector
 {
-    private $container;
+    private mixed $controllerNameParser;
```
b1rdex, line 23: This should have a default value (null) or the line 67 will lead to a fatal error (`if ($this->controllerNameParser`)

Annotation/Model.php
```diff
@@ -32,18 +32,15 @@ final class Model extends AbstractAnnotation
...
-     * @var string
-     */
-    public $type;
+    public string $type;
 
     /**
      * @var string[]
      */
-    public $groups;
+    public array $groups;
```
b1rdex, line 40: This will lead to fatal errors too: only $type is required.

b1rdex:
There are tons of hidden fatal errors. I'd recommend splitting the MR to a) allowing sf 6.0 and then b) upgrading to PHP 8 if necessary.

Also, introducing some static code analysis (PHPStan or Psalm) would ease PHP 8.0 upgrade, helping to find the errors.

michaljusiega:
I also thinking about spliting this PR.

For me now, this bundle is blocker to bump my projects with Symfony 6.0.

GuilhemN:
I merged https://github.com/nelmio/NelmioApiDocBundle/pull/1933 with your changes limited to the upgrade to symfony 6.0.
We should be all good, all that remains is to make a release of it :)

For migrating to PHP 8 and Symfony > 5.3, I'm still not in favor of it, this is not necessary and wouldn't really ease the maintenance so for me it should be delayed.

GuilhemN:
I'm closing this as its main goal is fulfilled, feel free to reopen this PR or a new one if you feel the need.

PR #1916, opened 2021-12-01
Add support for zircore/swagger-php 4.0

alshenetsky:
Which is introducing php 8.1 attributes support


GuilhemN:
Thanks for submitting this PR!

It looks like they introduced some more changes, notably `@OA\Undefined` was moved to `Generator::UNDEFINED` (see https://github.com/zircote/swagger-php/blob/master/src/Generator.php#L50), and a migration guide is available at https://github.com/zircote/swagger-php/blob/master/docs/Migrating-to-v4.md.

alshenetsky:
@GuilhemN I'm sorry for being flippant at first. Fixed, and all the tests are passing with v4. But the code is backwards incompatible with the v3 version of zircotte/swagger-php. Do I still have to fix it to get it back compatible with older versions? Or will nelmio/api-doc-bundle also have a major version?

GuilhemN:
No worry, it is very nice of you to start this pull request, I appreciate it very much :)

I believe bumping our requirement on zircote/swagger-php to 3.2.0 should be enough, isn't it? (changes in the API were normally done in this version, see https://github.com/zircote/swagger-php/commit/c5f0819764f9bd8a246732d662b95ce533c0479f)

alshenetsky:
@GuilhemN you are right. But I also had to add compatibility layer with 3.2 because of $context property which was moved to another class in 4.0. You can review it now :)

GuilhemN:
Thank you! It's a very promising change :)

PR #1919, opened 2021-12-03
disable html render if `symfony/asset` not installed

### LINKED ISSUE #1876: The service "nelmio_api_doc.render_docs.html.asset" has a dependency on a non-existent service "twig.extension.assets"
https://github.com/nelmio/NelmioApiDocBundle/pull/1842 require a dependency on `twig.extension.assets` but it's not a dependency of nelmio. Could we make this dependency optional ?

Error message:
```
Uncaught Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: The service "nelmio_api_doc.render_docs.html.asset" has a dependency on a non-existent service "twig.extension.assets".
```

Gemorroj:
fix #1876

it will work at least from symfony 4.4 https://github.com/symfony/twig-bundle/blob/4.4/DependencyInjection/Compiler/ExtensionPass.php#L28 to symfony 6.0 https://github.com/symfony/twig-bundle/blob/6.0/DependencyInjection/Compiler/ExtensionPass.php#L30

### DIFF
--- a/DependencyInjection/NelmioApiDocExtension.php
+++ b/DependencyInjection/NelmioApiDocExtension.php
@@ -152,14 +152,15 @@ final class NelmioApiDocExtension extends Extension implements PrependExtensionI
                 ->setArgument(1, $config['media_types']);
         }
 
-        // ApiPlatform support
         $bundles = $container->getParameter('kernel.bundles');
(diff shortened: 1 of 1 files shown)


GuilhemN:
That's great, thanks a lot for submitting this!

PR #1920, opened 2021-12-06
single controller for different doc formats

paramonov:
removed yaml controller - moved creating response object into renderer
removed manually passed renderer instances by autowire mechanism


Resources/config/services.xml
```diff
@@ -15,22 +19,14 @@
             <argument type="service" id="nelmio_api_doc.render_docs" />
         </service>
 
-        <service id="nelmio_api_doc.controller.swagger" alias="nelmio_api_doc.controller.swagger_json" public="true" />
-
-        <service id="nelmio_api_doc.controller.swagger_json" class="Nelmio\ApiDocBundle\Controller\DocumentationController" public="true">
```
GuilhemN, line 20: This looks like an unnecessary breaking change

Resources/config/services.xml
```diff
@@ -15,22 +19,14 @@
...
-            <argument type="service" id="nelmio_api_doc.render_docs" />
-        </service>
-
-        <service id="nelmio_api_doc.controller.swagger_yaml" class="Nelmio\ApiDocBundle\Controller\YamlDocumentationController" public="true">
```
GuilhemN, line 24: should be avoided as well

GuilhemN (commented):
Thank you for your pull request!
While this looks indeed better to merge the logic inside one controller, this code is already released and we need to be more careful and should add layers to keep backward compatibility with apps relying on previous versions.

I left mostly comments about backward compatibility concerns :)

DjordyKoert:

PR #1922, opened 2021-12-07
fixed : add doc blocks to silence Symfony deprecations

### LINKED ISSUE #1921: Twig deprecations
Symfony 5.4 returning deprecations about 
```
User Deprecated: Method "Twig\Extension\ExtensionInterface::getFunctions()" might add "array" as a native return type declaration in the future.
Do the same in implementation "Nelmio\ApiDocBundle\Render\Html\GetNelmioAsset" now to avoid errors or add an explicit @return annotation to suppress this message.
```

encreinformatique:
Symfony 5.4 returning deprecations about 

```
User Deprecated: Method "Twig\Extension\ExtensionInterface::getFunctions()" might add "array" as a native return type declaration in the future.
Do the same in implementation "Nelmio\ApiDocBundle\Render\Html\GetNelmioAsset" now to avoid errors or add an explicit @return annotation to suppress this message.
```

Closes #1921

### DIFF
--- a/Render/Html/GetNelmioAsset.php
+++ b/Render/Html/GetNelmioAsset.php
@@ -31,10 +31,7 @@ class GetNelmioAsset extends AbstractExtension
         $this->resourcesDir = __DIR__.'/../../Resources/public';
     }
 
-    /**
-     * @return array
-     */
-    public function getFunctions()
+    public function getFunctions(): array
     {
         return [
             new TwigFunction('nelmioAsset', $this, ['is_safe' => ['html']]),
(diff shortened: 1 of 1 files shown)


haidukua:
@encreinformatique 
mb use native PHP types)

 /**
     * {@inheritdoc}
     */
    public function getFunctions(): array
    {}

encreinformatique:
> @encreinformatique mb use native PHP types)

It was my first choice but I have doubts about child inheritance and PHP 7.1 (min package of this repo) as the parent function do not have type hinting.
I had in mind it was introduced in PHP 7.3. I'll check.

GuilhemN:
Nice, thank you @encreinformatique!

PR #1932, opened 2021-12-20
Add support for php attributes

### LINKED ISSUE #1929: Missing attributes support for @Model, @Area, @Operation and @Security
Because of these [annotations](https://github.com/nelmio/NelmioApiDocBundle/tree/master/Annotation), we can't use attributes supported by zircote/swagger-php v4.

akalineskou:
Closed the other PR https://github.com/nelmio/NelmioApiDocBundle/pull/1931
Since I noticed the functional tests used the OA\Property in the entities, so I cherry picked it into this one.

Closes #1929


akalineskou:
Well, this is annoying
```php
#[OA\Schema([
    'required' => ['id', 'user'],
    'value' => new OA\Property(property: 'virtual', properties: ['ref' => new Model(type: JMSUser::class)]),
])]
```
Fails even if you check `if (\PHP_VERSION_ID >= 80100) {`, `new` inside the attribute

akalineskou:
I managed to fix the tests by creating different folders for the controllers and entities which have 8.1 features.
`ControllerPhp81` and `EntityPhp81` accordingly, if there are better solutions do let me know.

akalineskou:
@GuilhemN Are you going to re-add those tests from the php8.1 controller/entity?
Without those tests I wouldnt have been able to find and fix the rest of the things that were broken.

The different folders were needed because the framework would load those files, and it would break for php version < 8.1 (when using new inside attributes), and < 8 (when using named parameters)

GuilhemN:
@akalineskou I think migrating all tests to PHP 8.1 attributes would make the maintenance too complicated (we would need to duplicate all tests for php < 8.1 and for php 8.1).

Here I instead tried to limit testing to just a few methods, I duplicated the `FOSRestController` and the `ApiController::fetchArticle` method with php 8.1 attributes. I believe that should be enough, what do you think?

GuilhemN:
In my opinion, that should be good now.
Sorry for pushing without notice, I just merged Symfony 6.x support and I wanted to merge your PR before publishing a new release, I had some comments but didn't want to pressure you to fix them before I could release :)

If you think it is necessary, we could have some more tests using php 8.1 attributes, I don't think we should migrate all of them though.

akalineskou:
Sure, whatever you think is best.

I appreciate the speedy release :P

GuilhemN:
Thanks a lot @akalineskou!

TheFox:
Where do I find a documentation for this?

GuilhemN:
There is some information on https://zircote.github.io/swagger-php/guide/attributes.html#nesting, you can also have a look at this example: https://github.com/zircote/swagger-php/tree/master/Examples/using-links-php81.

PR #1935, opened 2021-12-21
Add support for JMS v4

GuilhemN:


GuilhemN:
Oh yes, I wanted to push a quick fix through the github UI but for some reason the "commit" button was (and still is) disabled, and then I forgot ^^

Should be better now :)

PR #1938, opened 2021-12-23
Fixes symfony serializer being mandatory

### LINKED ISSUE #1937: Latest version crashes if symfony/serializer is not installed
`Error: Class "Symfony\Component\Serializer\Annotation\DiscriminatorMap" not found` after updating to 4.8

Because the new `getAnnotation` do not check if a class exists.

https://github.com/nelmio/NelmioApiDocBundle/blob/16608de67f01deee829e6ce50463fc8665ae448b/ModelDescriber/ObjectModelDescriber.php#L192

and it's called with a class of the symfony/serializer package

https://github.com/nelmio/NelmioApiDocBundle/blob/16608de67f01deee829e6ce50463fc8665ae448b/ModelDescriber/ObjectModelDescriber.php#L75

My guess is that the doctrine annotation reader was ignoring unknown class, but not the new one.

Not sure of what the fix should be

magnetik:
Fixes #1937 

I've tested locally but maybe we should have a functional test with symfony serializer not installed?

### DIFF
--- a/ModelDescriber/ObjectModelDescriber.php
+++ b/ModelDescriber/ObjectModelDescriber.php
@@ -188,6 +188,9 @@ class ObjectModelDescriber implements ModelDescriberInterface, ModelRegistryAwar
      */
     private function getAnnotation(\ReflectionClass $reflection, string $className)
     {
+        if (false === class_exists($className)) {
+            return null;
+        }
         if (\PHP_VERSION_ID >= 80000) {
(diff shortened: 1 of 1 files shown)


GuilhemN:
Thanks!

> I've tested locally but maybe we should have a functional test with symfony serializer not installed?

I guess the only way to test that would be through github actions, by adding more cases to the matrix, this would need some rethinking of our tests though to skip all tests were the Symfony serializer is required...

PR #1954, opened 2022-01-26
Allow global security policies to be removed

jackcutting:
When using the bundle and including global bundle security information using `package/nelmio_api_doc.yaml` as per the [Symfony documentation](https://symfony.com/bundles/NelmioApiDocBundle/current/index.html#using-the-bundle), it becomes impossible to override this security policy.

Instead of having to remove this global security and configure each route individually using the `@Security` annotation. It is much less repetitive to be able to remove all existing security policies for a specific route using this annotation.

Changes:
If the Security annotation is used on a route and the name is set to `null`, all security policies for that route will be removed.
Created documentation for using the `@Security` annotation

### DIFF
--- a/Describer/OpenApiPhpDescriber.php
+++ b/Describer/OpenApiPhpDescriber.php
@@ -109,6 +109,13 @@ final class OpenApiPhpDescriber
 
                 if ($annotation instanceof Security) {
                     $annotation->validate();
+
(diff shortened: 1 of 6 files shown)


GuilhemN:
Awesome, thank you!
I think your documentation is also a great addition, we had lot of questions about the management of security :)

Issue #1963, opened 2022-02-23
Support PHP 8.1 enums


magnetik:
Hi,

[PHP 8.1 enums](https://www.php.net/manual/en/language.enumerations.php) seems to be "badly" currently supported right now.

In request body, it shows them as an objet that requires a "name" parameter. 
It should be shows as an enum (swagger). 

I've done a basic pull request that works well for me.

GuilhemN:
Closing as https://github.com/nelmio/NelmioApiDocBundle/pull/1965 is merged.

PR #1965, opened 2022-02-24
Create an enum model describer

magnetik:
I have not written any test, but it works fine for my usage : a simple enum used as a choice list in a property of an object.

ref #1963

### DIFF
--- a/ModelDescriber/EnumModelDescriber.php
+++ b/ModelDescriber/EnumModelDescriber.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Nelmio\ApiDocBundle\ModelDescriber;
+
+use Nelmio\ApiDocBundle\Model\Model;
+use OpenApi\Annotations\Schema;
+use Symfony\Component\PropertyInfo\Type;
+
+class EnumModelDescriber implements ModelDescriberInterface
+{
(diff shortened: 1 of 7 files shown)


GuilhemN:
Thanks for working on this!

It seems there exists an integration in the symfony serializer (https://github.com/symfony/symfony/pull/40830) but only for backed enumarations.
So I am not sure we should try to document other enumerations as we would differ from the Symfony serializer behavior which we tried to mimic as much as possible until now.

What do you think? How are you dealing with normal enums serialization in your apps?

magnetik:
Hi!

It was super early in our use of enums in the codebase, and we switched to only backed enums. 

I'll update the PR to support only them for now.

GuilhemN:
I did not look into all the details but I believe the failure was due to an old bug in Api-Platform which has been fixed since.
Bumping Api-Platform's version seems to fix the tests so I guess we're good :)

filkaris:
Hello! Thank you for your contribution since this is an issue we faced as well

@GuilhemN is there a planned release in the near future that would include this feature?

Thank you!

GuilhemN:
@filkaris Sure! I just released 4.9.0: https://github.com/nelmio/NelmioApiDocBundle/releases/tag/v4.9.0.

Issue #1957, opened 2022-02-02
Defining Schema on class results in "User Warning: Multiple definitions for @OA\Schema()->type"


gempir:
I have a class like this

```
/**
 * @OA\Schema (
 *     type="array",
 *     @OA\Items(ref=@Model(type=TagData::class))
 * )
 */
class TagDataCollection extends AbstractCollection
```

With the goal of this collection being interpreted as array of TagData Models in NelmioApiDocBundle.
So I don't have to make the @OA\Property type hint every time in the parent object.

This results in a multiple definition error though:

```
User Warning: Multiple definitions for @OA\Schema()->type
Using: \Nelmio\ApiDocBundle\OpenApiPhp\MyNamespace\TagDataCollection::getSchema()
```

Is this a supported usecase?

swagger-php equivalent would be refs https://zircote.github.io/swagger-php/Getting-started.html#reusing-annotations-ref

chrisguitarguy:
> Is this a supported usecase?

It's probably _should_ be, and it seems like it should work.

Looks like this is coming from the open api library itself: https://github.com/zircote/swagger-php/blob/3f3b551db6eb934e8cb8c132ec2430e2e8c2ca0f/src/Annotations/AbstractAnnotation.php#L238

But the reference to `TagDataCollection::getSchema()` has me wondering if that's a method you've got on `TagDataCollection`?

I kinda suspect that the `ObjectModelDescriber` does it's thing on the class: https://github.com/nelmio/NelmioApiDocBundle/blob/9cd417efb4cac669ba517f66e9dfd484a4d93001/ModelDescriber/ObjectModelDescriber.php#L71-L73 and then continues on its way to try to define the properties for the `TagDataCollection` object. When... it shouldn't do that. Unless the `@OA\Schema` annotation explicitly had no type or an `object` type.

Any chance you could provide a minimal project that reproduces your error?

gempir:
Here is a fresh project based on a boilerplate. https://github.com/gempir/nelmio-docs-example

Just run `docker-compose up` and call localhost/api/doc.json

it will result in the same multiple definition error and it trying to do `getSchema` on my example collection.
Relevant Models in here https://github.com/gempir/nelmio-docs-example/tree/main/src/Model

![image](https://user-images.githubusercontent.com/9765622/160234377-b9ec8980-e43f-450b-98e2-656143328468.png)

PR #1978, opened 2022-04-12
Stop Model Property Description When a Schema Type or Ref is Already Defined

chrisguitarguy:
This basicaly let's user do this:

```php
/**
 * @OA\Schema(type="array", https://github.com/OA\Items(ref=@model(type=SomeEntity::class)))
 */
class SomeCollection implements \IteratorAggregate { }
```

on a plain PHP object or form and the model describers (object, JMS, and form) will stop and _not_ try to describe any further properties.

This should close https://github.com/nelmio/NelmioApiDocBundle/issues/1957 as well as quite a few other issues since this has come up a lot.

## Should We Skip Property Description on _any_ Type Defined?

Right now if an `object` type in a schema will still describe properties. Should we completely short circuit if _any_ type is defined? Assume the user has manually defined everything?

### DIFF
--- a/ModelDescriber/Annotations/UpdateClassDefinitionResult.php
+++ b/ModelDescriber/Annotations/UpdateClassDefinitionResult.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * This file is part of the NelmioApiDocBundle package.
+ *
+ * (c) Nelmio
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Nelmio\ApiDocBundle\ModelDescriber\Annotations;
+
+/**
+ * result object returned from `AnnotationReader::udpateDefinition` as a way
+ * to pass back information about manually defined schema elements.
+ */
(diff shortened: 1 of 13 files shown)


ModelDescriber/Annotations/UpdateClassDefinitionResult.php
```diff
@@ -0,0 +1,39 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Nelmio\ApiDocBundle\ModelDescriber\Annotations;
+
+/**
+ * result object returned from `AnnotationReader::udpateDefinition` as a way
+ * to pass back information about manually defined schema elements.
+ */
+final class UpdateClassDefinitionResult
```
GuilhemN, line 18: I think this should be `@internal`

GuilhemN (approved):
👍🏻 for me, I was afraid at first that this would be too magic, but actually this should have an effect only in cases where that just wouldn't make sense to document the properties anyway.

GuilhemN:
Thank you @chrisguitarguy :)

PR #1983, opened 2022-04-20
Update areas.rst

kira0269:
'with_annotations' option is not mentioned in actual documentation


GuilhemN:
Good catch, thank you @kira0269!

PR #1986, opened 2022-04-27
Add previous exception to BadRequestHttpException on SwaggerUiController

WedgeSama:
My case, while I dev, I got an error on Swagger UI without explicite details about why it bugged.

With previous exception set on line 51, it improve debug by providing the original exception.

### DIFF
--- a/Controller/SwaggerUiController.php
+++ b/Controller/SwaggerUiController.php
@@ -48,7 +48,7 @@ final class SwaggerUiController
                 $advice = ' Since the area provided contains `.json`, the issue is likely caused by route priorities. Try switching the Swagger UI / the json documentation routes order.';
             }
 
-            throw new BadRequestHttpException(sprintf('Area "%s" is not supported as it isn\'t defined in config.%s', $area, $advice));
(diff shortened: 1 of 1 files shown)


GuilhemN:
Did you encounter a case where you got an error that was not due to the misconfiguration explained in the message?
Could you give us some details about what original exception it originated from?

WedgeSama:
Yes, I encounter an error that is not due to mis configuration of NelmioApiDocBundle.

I currently working (not only me) on a bundle at my company that make ours APIs (and use this bundle to generate Doc). And the exception was from our bundle and not propagate.

We go the error at 2 points:
- On the action where the bug is (our code)
- On the load of Swagger UI in NelmioApiDocBundle

Hope it will be sufficient for you, but if you need more info, do not hesitate to ask ;)

PS: Our bundle will be open sourced in the year.

chrisguitarguy (approved):

GuilhemN:
Thinking about this again I believe we should instead catch a more precise exception to avoid catching userland exceptions.
I believe this large catch was introduced by mistake, I proposed a fix in https://github.com/nelmio/NelmioApiDocBundle/pull/2005.

Issue #1984, opened 2022-04-20
Version is only available reliably for validation and serialization


igordeveloper:
I'm receiving the message below to generate a api doc after update in bundle zircote/swagger-php

**{"message":"Version is only available reliably for validation and serialization","code":0,"status":500}**

Below current configuration and version bundles:

```
nelmio/api-doc-bundle               v4.8.2
zircote/swagger-php                  v4.3.0
```

nelmio_api_doc.yaml
```
nelmio_api_doc:
    areas:
        path_patterns: # an array of regexps
        - ^/[a-z]
    documentation:
        openapi: 3.1.0
        components:
        ....
```
Any idea what configuration should be done to tell the openapi version?
Tks.

DerManoMann:
Related to https://github.com/zircote/swagger-php/issues/1213.

The underlying change was made to allow to set the version in the `@OA\OpenApi` annotations. This means there is no reliable way to know the actual version until analysis is complete.

So, in order to avoid changing the version halfway through processing, attempts to use the version via the `Context` are rejected.

This is not an issue when using `swagger-php` standalone as the `Generator` class handles this internally.

That means setting the version via `Generator::setVersion()` still works, but access via the context is only guaranteed for processors, validation and serialization.

I would suspect the issue here is that annotation instances are created that do not share the context with the root context created by the generator (if that is used at all).

I am not familiar with many details of this project, but if you are happy to only support setting the version via your bundle config then setting the version in all `Context` instances created should solve this issue.

Ideally, all contexts should be child instances of a single root context. That allows to share things like version, logger and a couple other things across all code that has access.

igordeveloper:
> Related to [zircote/swagger-php#1213](https://github.com/zircote/swagger-php/issues/1213).
> 
> The underlying change was made to allow to set the version in the `@OA\OpenApi` annotations. This means there is no reliable way to know the actual version until analysis is complete.
> 
> So, in order to avoid changing the version halfway through processing, attempts to use the version via the `Context` are rejected.
> 
> This is not an issue when using `swagger-php` standalone as the `Generator` class handles this internally.
> 
> That means setting the version via `Generator::setVersion()` still works, but access via the context is only guaranteed for processors, validation and serialization.
> 
> I would suspect the issue here is that annotation instances are created that do not share the context with the root context created by the generator (if that is used at all).
> 
[...]

PR #1988, opened 2022-04-30
Use the same root context everywhere

GuilhemN:
Fixes https://github.com/nelmio/NelmioApiDocBundle/issues/1984

Currently, in some places new Swagger-php annotations do not have a parent context, and thus do not inherit from the root context.
This causes issues with the new version field which uses the context extensively.

This PR fixes this and uses the same root context everywhere.


chrisguitarguy:
Seems like #1984 is closed now, but this still seems like a good idea, right? Same root context should always be used?

GuilhemN:
Yes definitely!

PR #1989, opened 2022-05-02
docs(commands.rst): fix command name

gnumoksha:


GuilhemN:
Indeed, good catch!
Thank you @gnumoksha :)

Issue #1770, opened 2021-01-07
Strict CSP settings - inline css is not allwoed


fliespl:
Problem is due to inline css on svg image.

```
style="position:absolute;width:0;height:0"
```

when unsafe-inline is disabled on style-src directive.

GuilhemN:
Fixed in #1999

PR #1999, opened 2022-05-16
Remove Inline Styles in Favor of the Stylesheet

chrisguitarguy:
Should fix https://github.com/nelmio/NelmioApiDocBundle/issues/1770


GuilhemN:
Nice, thank you!

PR #2008, opened 2022-07-15
Fix typo "expect" to "except" in documentation.

cyrilverloop:
I suppose it means that routes under `/api/doc` will not be parsed for api documentation.


GuilhemN:
Oh Indeed, good catch!
Thank you @cyrilverloop .

PR #2011, opened 2022-07-18
The ObjectModelDescriber can support interfaces

magnetik:
I've encountered a class hierarchy where there was an interface.

The ObjectModelDescriber already behave well with interface but the `class_exists` prevent it from being used with interface.

Not sure if I'm missing something.


GuilhemN:
Thank you!

PR #2013, opened 2022-07-18
Allow symfony/flex composer plugin in CI

magnetik:
Recent composer version requires to allow plugin before execution.


maxhelias:
FYI no need to install symfony/flex globaly since composer v2

PR #2016, opened 2022-08-03
Adding support for Symofny Assert\Sequentially

cretiv-dev:
Currently ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php not display constraints when use Assert\Sequentially, it should works like Assert\Compound


GuilhemN:
Indeed, thank you @cretiv-dev!

PR #2024, opened 2022-09-03
Fix  Disable Default Section exempale

rmed19:
The disable_default_routes is under areas and not under default. 
I'm using the version 4.9 of the bundle

### DIFF
--- a/Resources/doc/faq.rst
+++ b/Resources/doc/faq.rst
@@ -206,8 +206,7 @@ A: Use ``disable_default_routes`` config in your area.
 
     nelmio_api_doc:
         areas:
-            default:
-                disable_default_routes: true
+            disable_default_routes: true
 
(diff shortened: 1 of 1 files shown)


GuilhemN:
Just checked and this example should work fine.

If you define `path_patterns`, `host_patterns` or `documentation` under `nelmio_api_doc.areas`, it will be aliased to `nelmio_api_doc.areas.default` which may be what you've done in your app.

rmed19:
@GuilhemN 
In my case, I just installed the bundle, and I added the option us mention on the documentation, but i got I error telling that no option "disable_default_routes" under "areas.default"
To resolve the problem I made this : 
`nelmio_api_doc:
    areas: # to filter documented areas
        path_patterns:
            - ^/api(?!/doc$) # Accepts routes under /api except /api/doc
        disable_default_routes: true`

PR #2028, opened 2022-09-09
Fix methods in route annotation in documentation

johnrazeur:
Fix the bad typo.


GuilhemN:
Indeed, thank you @johnrazeur!

PR #2034, opened 2022-09-19
Read discriminator mapping from file configuration

EmilMassey:
Currently, the bundle can extract discriminator mappings only from the annotations. This MR uses Serializer's metadata factory to support all of configurations: annotations, yaml files, etc.


GuilhemN:
Good idea! Thank you very much @EmilMassey :)

PR #2042, opened 2022-10-10
Fix OA\XmlContent class reference

nosnickid:
Simple fix to reference class correctly, fixing XML API descriptions.


GuilhemN:
Thank you @nosnickid :)

Issue #2050, opened 2022-11-05
Indirect deprecations due to zircote/swagger-php 4.5.0


schodemeiss:
Masses of indirect deprecations about Context::detect() from zircote/swagger-php.

This has been raised in an issue upstream (https://github.com/zircote/swagger-php/issues/1339) but as they point out, perhaps the issue is with this bundle.

DerManoMann:
I suspect this is the result of rolling back #2043 . 
I am not convinced the issue was the number of objects created, since swagger-php will fall back to creating a new Context via the deprecated detect method.
Surly something to investigate. If this becomes too much of an issue we could perhaps turn off the deprecated message, but then I wonder how other projects do this

GuilhemN:
Thank you for the report!

I tried to fix that issue in https://github.com/nelmio/NelmioApiDocBundle/pull/2053 (and also sort of reintroduce the changes reverted in #2043).

But I think it is problematic to throw a deprecation in this method as doctrine annotations instanciate all annotations they find on a method/class/property. So any code retrieving annotations where there is a Swagger-PHP annotation will trigger the deprecation. I think most apps don't see them only because the annotations loader is cached and the deprecations are thus triggered only on the first run.
This should be fixed with PHP attributes though as third party apps should not instanciate Swagger-PHP attributes out of context.

If the goal is ultimately to throw an exception when the context is not explicitly given, this will likely break a lot of apps (because of how doctrine annotations are retrieved).

PR #2053, opened 2022-11-13
Fix deprecation of Context::detect() and copy root context properties everywhere

GuilhemN:
This PR is an attempt to set an inherited context everywhere again (reverted in https://github.com/nelmio/NelmioApiDocBundle/pull/2043).

The memory leak seems to be caused by references to parent/nested annotations, which likely makes the mission of the garbage collector more difficult.
So instead of actually referencing the parent annotation everywhere, we copy its most important properties.

Should fix https://github.com/nelmio/NelmioApiDocBundle/issues/2050#issuecomment-1304632460 and https://github.com/nelmio/NelmioApiDocBundle/issues/1984.


GuilhemN:
@BafS @grimgit would you be able to confirm this does not reintroduce a memory leak?

From my testing the memory usage only increases slightly with these changes, but I'd better be safe this is not a very deterministic thing to test :)

PR #2055, opened 2022-11-17
Fix `phpdocumentor/reflection-docblock` usage for `dev`

aweelex:
I found that `property_info.php_doc_extractor` won't be available at Symfony ^5.0, if NelmioApiDocBundle is required for dev only.

https://github.com/symfony/symfony/blob/6.1/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L1889

That's why I added custom pass for this class. Since `reflection-docblock` is mandatory, there is no need to check for `ContainerBuilder::willBeAvailable`.

### DIFF
--- a/DependencyInjection/Compiler/PhpDocExtractorPass.php
+++ b/DependencyInjection/Compiler/PhpDocExtractorPass.php
@@ -1,0 +1,34 @@
+<?php
+
+/*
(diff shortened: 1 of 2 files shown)


GuilhemN:
Good catch, thank you @aweelex!

PR #2057, opened 2022-11-22
Fix documentation

ColonelMoutarde:
add square close braket  in sample


GuilhemN:
Good catch, thank you @ColonelMoutarde!

PR #2059, opened 2022-11-30
Check nullable property has nullable set

akalineskou:
My use case is I have an entity which has a nullable string `?string` property (so the deserialization will work, since `null` can be passed) but the validation will check that it is not null, in which case I know that it will always be a `string`

Without this change, even tough I have `@OA\Property(nullable=false)` set in the entity, it gets ignored... this change fixes this and it will show up correctly as `nullable: false` in swagger docs

Made a small change after the fact, it checks if `nullable` is not undefined, and if it's false it will mark it undefined, so they don't show up as `nullable: false` for all the properties in the doc which is not necessary


GuilhemN:
Thank you @akalineskou!

PR #2061, opened 2022-12-08
Reorganize server urls | fixes #2060

### LINKED ISSUE #2060: Unable to specify server_url and servers urls
This commit (https://github.com/nelmio/NelmioApiDocBundle/commit/181f4b276356f34ea307b5d44ec7ff6cc189ea27) broke functionality of defining servers urls to area, as well you can't specify now custom server_url, its only taken from request parameters and can't be rewrited.

1. First issue is that if you specify servers for area it is rewrited by: https://github.com/arturssmirnovs/NelmioApiDocBundle/blob/master/Render/RenderOpenApi.php#L77

2. Second issue is that you can't specify custom server_url for default area. As its taken from: https://github.com/arturssmirnovs/NelmioApiDocBundle/blob/master/Render/RenderOpenApi.php#L54

Real word scenario for second issue is when you have HTTPS connection managed by Cloudflare or some loadbalancer & end system gets 80 port request, doc generates links as http, but it would be nice to force overwrite them with server_url to https.

arturssmirnovs:


GuilhemN:
I updated the tests to go a bit deeper and fixed some issues that were uncovered.
All is good now.

Thank you!

PR #2062, opened 2022-12-09
Add enum description to route parameter if possible.

lhausammann:
Adds an "enum" schema to route parameters if the routing requirements consists only of a simple OR regex epression.
It skips complex expression.
The schema "pattern" field still contains the original regex.
This is mainly useful to display enum values in the Swagger UI gui as select boxes rather than text boxes.


GuilhemN:
Nice addition @lhausammann, thank you!

PR #2067, opened 2023-01-13
Adds enum support for Choice Constraint in QueryParam

Apipa169:
This adds support for Choice Constraints used in the FOSRest QueryParam.

Currently it is required to describe the Parameter separately if you want to show the valid choices as enum in the documentation. But with this change it will do that automatically for the user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,5 +52,3 @@ Then run the test suite:
 ## License
 
 This bundle is released under the MIT license.
-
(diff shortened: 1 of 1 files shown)


Apipa169:
All tests work properly now for PHP 8 and lower. Tests for PHP 8.1 are failing for an unrelated issue, as this is also failing in the master branch.

PR #2068, opened 2023-01-13
Fixes assertion in test to make it compatible with zircote/swagger-php >=4.5.5

Apipa169:
Since zircote/swagger-php [4.5.5](https://github.com/zircote/swagger-php/releases/tag/4.5.5) the Attribute `OpenApi\Attributes\PathParameter` no longer extends `OpenApi\Annotations\PathParameter`, which broke the testcase that is being modified in this PR.

It is also possible to just assert if it is an instance of `OpenApi\Generator\Parameter`, but I personally I prefer to keep it as specific as it is right now.

I'm not a big fan of the aliases I made, so I'm happy if there are better suggestions. On the other hand, this is pretty clear now.

### DIFF
--- a/Tests/Functional/FunctionalTest.php
+++ b/Tests/Functional/FunctionalTest.php
@@ -13,7 +13,8 @@ namespace Nelmio\ApiDocBundle\Tests\Functional;
 
 use Nelmio\ApiDocBundle\OpenApiPhp\Util;
 use Nelmio\ApiDocBundle\Tests\Helper;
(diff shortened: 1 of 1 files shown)


Apipa169:
There are a few new Attributes that do not extend their Annotation variant, which are also newly added to 4.5.5:

- CookieParameter
- HeaderParameter
- QueryParameter

As these are completely new I don't see any issues with that.

GuilhemN:
Thank you @Apipa169!

PR #2075, opened 2023-02-03
Fixed minor error in the default security policy documentation.

mickverm:
A missing ``-`` prevents the default security policy from being set.
Added attributes/annotations switcher to the configuration block.


GuilhemN:
thank you @mickverm!

Issue #1873, opened 2021-09-14
Should Route Requirements be Surround with `^...$` in Path Schema Patterns?


chrisguitarguy:
Given a route like this:

```
    <route
        id="users.view"
        methods="GET"
        path="/users/{userId}"
        controller="App\Controller\UsersController::view">
            <requirement key="userId">[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}</requirement>
    </route>
```

an open API schema gets generated where the path parameter looks like this:

```yaml
parameters:
  -
    name: userId
    in: path
    required: true
    schema:
      type: string
      pattern: '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}'
 ```

But the [pattern docs](https://swagger.io/docs/specification/data-models/data-types/#pattern) say that without `^` and `$` the pattern is a partial match, not a full string.

For a route, where the generated regex would require the specific, full string it might make sense to change the [route describer](https://github.com/nelmio/NelmioApiDocBundle/blob/b53edda10ce4a263991d6f10032a905f783b3d29/RouteDescriber/RouteMetadataDescriber.php#L60) to surround the requirement with `^...$`.

Here's the regex that symfony generates on that route, for context:

![Screen Shot 2021-09-14 at 10 58 10 AM](https://user-images.githubusercontent.com/1010392/133292541-487df4ee-1738-4b63-b350-89c4ff9707ba.png)

Thoughts?

**Comment by vodevel** [comment]
Thank you, @chrisguitarguy! Absolutely make sense!

The problem cannot even be solved by manual adding `^...$` like

```php
#[Route('route/{id}', requirements: ['id' =>  '^\d+$])]
// or
#[Route('route/{id<^\d+$>}')]
```

Due to Symfony Router converts all these variants to `{^/route/(?P<id>\d+)$}sDu`

Therefore, it is worth adding boundary conditions for correct UI validation. Is there any reason not to do this, other than not having a pull request?)

PR #2083, opened 2023-03-02
fix: validation of requirements in route description

vodevel:
See https://github.com/nelmio/NelmioApiDocBundle/issues/1873

Example
```php
#[Route('route/{id}', requirements: ['id' =>  '\d+'])]
```
Now Swagger UI allows any {id} that contains at least one digit (eg. `x1` or `"1"`). Whereas in Symfony Routing allows {id} that consists only of digits. To be sure, run
```
bin/console route:match /route/x1
# Error: almost matches but requirement for "id" does not match (\d+)

bin/console route:match /route/11
# Success: Path Regex   | {^/route/(?P<id>\d+)$}sDu  
```
See also https://github.com/symfony/routing/blob/6.2/RouteCompiler.php#L301

Therefore, a wrapper has been added to fully comply with the requirements.

### DIFF
--- a/RouteDescriber/RouteMetadataDescriber.php
+++ b/RouteDescriber/RouteMetadataDescriber.php
@@ -68,7 +68,7 @@ final class RouteMetadataDescriber implements RouteDescriberInterface
                     }
                     // add the pattern anyway
                     if (Generator::UNDEFINED === $parameter->schema->pattern) {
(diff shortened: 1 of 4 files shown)


shakaran:
@vodevel could you rebase with last updates? Thanks

PR #2088, opened 2023-03-18
Add support Pure Enum and remove IntBackedEnum interface

pvlg:
Removed the IntBackedEnum interface, since it's missing from PHP, the check doesn't make sense.

Added support for Pure Enum.

Added tests for EnumModelDescriber.

### DIFF
--- a/ModelDescriber/EnumModelDescriber.php
+++ b/ModelDescriber/EnumModelDescriber.php
@@ -2,22 +2,33 @@
 
 namespace Nelmio\ApiDocBundle\ModelDescriber;
 
+use BackedEnum;
 use Nelmio\ApiDocBundle\Model\Model;
 use OpenApi\Annotations\Schema;
 use Symfony\Component\PropertyInfo\Type;
+use UnitEnum;
 
 class EnumModelDescriber implements ModelDescriberInterface
 {
-    public function describe(Model $model, Schema $schema)
+    public function describe(Model $model, Schema $schema): void
     {
+        /** @var UnitEnum $enumClass */
         $enumClass = $model->getType()->getClassName();
 
         $enums = [];
-        foreach ($enumClass::cases() as $enumCase) {
-            $enums[] = $enumCase->value;
+        if (is_subclass_of($enumClass, BackedEnum::class)) {
+            foreach ($enumClass::cases() as $enumCase) {
+                $enums[] = $enumCase->value;
+            }
+            $type = isset($enums[0]) && is_int($enums[0]) ? 'integer' : 'string';
+        } else {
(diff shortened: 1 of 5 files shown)


GuilhemN:
Thanks for this PR!

As mentioned in https://github.com/nelmio/NelmioApiDocBundle/pull/1965#issuecomment-1054124176, I'm still not sure we should support enums that are not `BackedEnum` though.
Those are not supported by Symfony Serializer as far as I can see, so supporting them would make us have a different behavior.

Unless I'm wrong and pure enums are correctly serialized? What would be your case for pure enums support?

webspec2012:
@GuilhemN 

> Unless I'm wrong and pure enums are correctly serialized? What would be your case for pure enums support?

At the moment I am using DTO (Assert) PHP Enum for validation according to the following recommendation:

[`https://github.com/symfony/symfony/pull/43047#issuecomment-1161704955`](https://github.com/symfony/symfony/pull/43047#issuecomment-1161704955)

**ChannelStatusEnum::class**

`
enum ChannelStatusEnum: string
{
    case PUBLISHED = 'published';
    case UNPUBLISHED = 'unpublished';
    case BLOCKED = 'blocked';

    public static function values(): array
    {
        return \array_column(self::cases(), 'value');
    }
}
`

**DTO**

`
#[Assert\NotBlank]
#[Assert\Choice(callback: [ChannelStatusEnum::class, 'values'] , multiple: false)]
public readonly string $status,
`

The scheme is working, there are no problems. But I can't specify it in the NelmioApiDocBundle.
It doesn't work that way:

**Request DTO**

`
#[OA\Property(type: 'string', enum: ChannelStatusEnum::class, example: 'published', nullable: false)]
public readonly string $status,
`

A simple solution would be to specify so:

`
#[OA\Property(type: 'string', enum: ChannelStatusEnum::values(), nullable: false)]
public readonly string $status,
`

But you can't do that-(

GuilhemN:
We automatically extract info from Assert/Choice, see https://github.com/nelmio/NelmioApiDocBundle/blob/master/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php#L107-L109, so just having `#[Assert\NotBlank] #[Assert\Choice(callback: [ChannelStatusEnum::class, 'values'] , multiple: false)] public readonly string $status` and then `#[OA\Property(type: 'string', example: 'published', nullable: false)] public readonly string $status` should work

webspec2012:
@GuilhemN, Thanks! This solution works.

DjordyKoert:
I will go ahead and close this one because of https://github.com/nelmio/NelmioApiDocBundle/pull/2121

PR #2096, opened 2023-04-27
feat: add arrays of enums and enum callbacks for query params

indexxd:
This adds support for the `callback` option in Choice constraint in QueryParams to allow providing enum choices at runtime. Also adds support for arrays of enums using the `multiple` QueryParam option.


shakaran:
@indexxd  could you rebase with last updates? Thanks

indexxd:
@shakaran done, should be good now.

shakaran:
@DjordyKoert could you check this for quick merge if suitable? It is already rebased

PR #2097, opened 2023-04-28
Improved extendability of the ObjectModelDescriber (private to protected)

stollr:
I'd like to add a specialized model describer and therefore want to extend the `ObjectModelDescriber` and override the `describeProperty` method. Currently this is not possible because it is private - like some other items of that class, too.

It would be nice if this PR could be merged, so that I can extend the bundle.


stollr:
I realized that the same could be achieved by service decoration. So I'll close this PR.

PR #2098, opened 2023-04-28
feat: added symfony/uuid property describer

stollr:
This describer adds support for UUIDs from the following libraries: symfony/uid and ramsey/uuid

This makes types of these classes configured as 
```
"type":  "string",
"format": "uuid",
"pattern": "^[0-9a-f]{8}-?[0-9a-f]{4}-?[0-9a-f]{4}-?[0-9a-f]{4}-?[0-9a-f]{12}$"
```

This is based on the [spec](https://swagger.io/docs/specification/data-models/data-types/#string).


DependencyInjection/NelmioApiDocExtension.php
```diff
@@ -161,6 +161,11 @@ public function load(array $configs, ContainerBuilder $container): void
         $container->registerForAutoconfiguration(ModelDescriberInterface::class)
             ->addTag('nelmio_api_doc.model_describer');
 
+        // Remove UUID describer if none of the supported libraries exists
+        if (!class_exists(\Symfony\Component\Uid\Uuid::class) && !class_exists(\Ramsey\Uuid\UuidInterface::class)) {
+            $container->removeDefinition('nelmio_api_doc.object_model.property_describers.uuid');
+        }
+
```
DjordyKoert, line 168: I think it makes more sense to only create the service if they do exists, instead of removing it.

PropertyDescriber/UuidPropertyDescriber.php
```diff
@@ -0,0 +1,38 @@
...
+        $property->format = 'uuid';
+        $property->pattern = '^[0-9a-f]{8}-?[0-9a-f]{4}-?[0-9a-f]{4}-?[0-9a-f]{4}-?[0-9a-f]{12}$';
+        $this->setNullableProperty($types[0], $property);
+    }
+
+    public function supports(array $types): bool
+    {
+        return 1 === count($types)
+            && Type::BUILTIN_TYPE_OBJECT === $types[0]->getBuiltinType()
+            && (is_a($types[0]->getClassName(), Uuid::class, true)
```
DjordyKoert, line 35: Use the `AbstractUid` class instead just like Symfony's normalizer does https://github.com/symfony/symfony/blob/7.0/src/Symfony/Component/Serializer/Normalizer/UidNormalizer.php

PropertyDescriber/UuidPropertyDescriber.php
```diff
@@ -0,0 +1,38 @@
...
+                || is_a($types[0]->getClassName(), UuidInterface::class, true));
```
DjordyKoert, line 36: It doesn't look like Symfony's serializer supports this out of the box.

I think users of this bundle should implement their own  `PropertyDescriber` which describes the behaviour for describing it, because there is no way for us to know what `UuidInterface::toString()` will output.

PropertyDescriber/UuidPropertyDescriber.php
```diff
@@ -0,0 +1,38 @@
...
+
+class UuidPropertyDescriber implements PropertyDescriberInterface
+{
+    use NullablePropertyTrait;
+
+    public function describe(array $types, OA\Schema $property, array $groups = null)
+    {
+        $property->type = 'string';
+        $property->format = 'uuid';
+        $property->pattern = '^[0-9a-f]{8}-?[0-9a-f]{4}-?[0-9a-f]{4}-?[0-9a-f]{4}-?[0-9a-f]{12}$';
```
DjordyKoert, line 27: This pattern might not necessarily be valid. 
https://github.com/symfony/symfony/blob/7bbc9be7ea7d607fa901edb2068c2cea13d24978/src/Symfony/Component/Serializer/Normalizer/UidNormalizer.php#L53-L62

DjordyKoert (changes requested):

DjordyKoert:
I would also like to see some test for this

DependencyInjection/NelmioApiDocExtension.php
stollr: Yeah, that makes sense ;-)

PropertyDescriber/UuidPropertyDescriber.php
stollr: Done.

PropertyDescriber/UuidPropertyDescriber.php
stollr: > It doesn't look like Symfony's serializer supports this out of the box.

If the serializer is configured by the framework bundle it is the case (see [here](https://github.com/symfony/symfony/blob/7bbc9be7ea7d607fa901edb2068c2cea13d24978/src/Symfony/Bundle/FrameworkBundle/Resources/config/serializer.php#L113)).

> I think users of this bundle should implement their own PropertyDescriber which describes the behaviour for describing it, because there is no way for us to know what UuidInterface::toString() will output.

Usually it should output a UUID. If this is not the case the user can override it with the property attribute. And if there is the demand one could extend the bundle's configuration to make if configurable.

I see an advantage in providing this describer out of the box, as it makes live easier for devs.

DjordyKoert: > If the serializer is configured by the framework bundle it is the case (see [here](https://github.com/symfony/symfony/blob/7bbc9be7ea7d607fa901edb2068c2cea13d24978/src/Symfony/Bundle/FrameworkBundle/Resources/config/serializer.php#L113)).

I am specifically talking about `Ramsey\Uuid\UuidInterface`. Serialization of this class is not explicitly supported by Symfony. 

It does work through a different normalizer ([JsonSerializableNormalizer](https://github.com/symfony/symfony/blob/7.0/src/Symfony/Component/Serializer/Normalizer/JsonSerializableNormalizer.php)) because it implements `JsonSerializable` but there is no real easy way (at the moment) for us to figure out what it would output. 

I don't think this bundle is the right place for describing Ramsey uuid at the moment.

stollr: Okay, I understand and I am fine with that. I have removed the support of ramsey/uuid.

PropertyDescriber/UuidPropertyDescriber.php
stollr: Okay, I have removed the pattern

stollr:
I'll try to add some.

DependencyInjection/NelmioApiDocExtension.php
```diff
@@ -161,6 +162,13 @@ public function load(array $configs, ContainerBuilder $container): void
         $container->registerForAutoconfiguration(ModelDescriberInterface::class)
             ->addTag('nelmio_api_doc.model_describer');
 
+        // Remove UUID describer if none of the supported libraries exists
```
DominicLuidold, line 165: The comment is outdated

PropertyDescriber/UuidPropertyDescriber.php
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Nelmio\ApiDocBundle\PropertyDescriber;
+
+use OpenApi\Annotations as OA;
+use Symfony\Component\PropertyInfo\Type;
+use Symfony\Component\Uid\AbstractUid;
+
+class UuidPropertyDescriber implements PropertyDescriberInterface
+{
+    use NullablePropertyTrait;
```
DominicLuidold, line 20: The trait has been deprecated with #2184 and can be removed in conjunction with calling `setNullableProperty` in line 26. 🙂

DominicLuidold (changes requested):
@stollr Any help needed with implementing test cases for this? 😃

DependencyInjection/NelmioApiDocExtension.php
stollr: Thanks. Fixed now

PropertyDescriber/UuidPropertyDescriber.php
stollr: Fixed

stollr:
@DominicLuidold I have rebased the PR to current master. But running the tests, I get

> 1) Nelmio\ApiDocBundle\Tests\Command\DumpCommandTest::testJson with data set "pretty print" (array(), 192)
> Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: The service "sensio_framework_extra.controller.listener" has a dependency on a non-existent service "annotation_reader".

Any idea how to fix this?

DominicLuidold:
@stollr Let me know if I can help in any other way 😄

stollr:
@DominicLuidold I have added a functional and some unit tests and hope it is correct the way I did it.
Please let me know, if anything is missing or wrong.

composer.json
```diff
@@ -46,6 +46,7 @@
         "symfony/validator": "^5.4|^6.0|^7.0",
         "symfony/expression-language": "^5.4|^6.0|^7.0",
         "symfony/security-csrf": "^5.4|^6.0|^7.0",
+        "symfony/uid": "^5.1|^6.0|^7.0",
```
DjordyKoert, line 49:
```suggestion
        "symfony/uid": "^5.4|^6.0|^7.0",
```

src/DependencyInjection/NelmioApiDocExtension.php
```diff
@@ -177,6 +178,12 @@ public function load(array $configs, ContainerBuilder $container): void
         $container->registerForAutoconfiguration(ModelDescriberInterface::class)
             ->addTag('nelmio_api_doc.model_describer');
 
+        if (class_exists(\Symfony\Component\Uid\Uuid::class)) {
```
DjordyKoert, line 181: This check could be removed to `UuidPropertyDescriber::supports()` and always register the describer in the `.xml`.

tests/Functional/Controller/ApiController81.php
```diff
@@ -345,6 +346,19 @@ public function entityWithObjectType()
     {
     }
 
+    #[Route('/entity-with-uuid', methods: ['GET', 'POST'])]
+    #[OA\Get(operationId: 'entity-with-uuid')]
```
DjordyKoert, line 350: Should not be needed for tests to function
```suggestion
```

tests/Functional/Controller/ApiController80.php
```diff
@@ -414,6 +415,19 @@ public function entityWithObjectType()
     {
     }
 
+    /**
+     * @Route("/entity-with-uuid", methods={"GET", "POST"})
+     *
+     * @OA\Get(operationId="entity-with-uuid")
```
DjordyKoert, line 421: Should not be needed for tests to function
```suggestion
```

tests/Functional/FunctionalTest.php
```diff
@@ -893,6 +893,17 @@ public function testEntitiesWithOverriddenSchemaTypeDoNotReadOtherProperties()
...
 
+    public function testEntityWithUuid()
+    {
+        $model = $this->getModel('EntityWithUuid');
+
+        $this->assertSame('object', $model->type);
+
+        $this->assertSame('id', $model->properties[0]->property);
+        $this->assertSame('string', $model->properties[0]->type);
+        $this->assertSame('uuid', $model->properties[0]->format);
```
DjordyKoert, line 904: I prefer testing the entire model instead of individual keys to prevent any unknown/unwanted changes from appearing in the future (extra keys for example).

You can see an example of this here 😄 https://github.com/nelmio/NelmioApiDocBundle/blob/4ff744fcaa21bdd6722e6c1772154eedc5982871/tests/Functional/FunctionalTest.php#L856

tests/PropertyDescriber/UuidPropertyDescriberTest.php
```diff
@@ -0,0 +1,81 @@
...
+
+use Nelmio\ApiDocBundle\PropertyDescriber\UuidPropertyDescriber;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\PropertyInfo\Type;
+use Symfony\Component\Uid\Uuid;
+
+/**
+ * @covers \Nelmio\ApiDocBundle\PropertyDescriber\UuidPropertyDescriber::supports
+ * @covers \Nelmio\ApiDocBundle\PropertyDescriber\UuidPropertyDescriber::describe
+ */
```
DjordyKoert, line 22: I don't think this provides any benefits

```suggestion
```

DjordyKoert (changes requested):

src/DependencyInjection/NelmioApiDocExtension.php
DominicLuidold: Wouldn't it also make sense to register the `PropertyDescriberInterface` for autoconfiguration (just like with the `ModelDescriberInterface` a few lines above)?

With the approach of using the `supports()` method, all describers are registered in the `.xml` anyway and this would reduce the configuration effort both for the bundle and for project-specific/userland describers.

composer.json
stollr: Fixed

src/DependencyInjection/NelmioApiDocExtension.php
stollr: @DjordyKoert There's already a check in the `supports` method. I am intentionally registering the service only if symfony/uid is installed, to prevent performance impact on projects not having installed symfony/uid. But I have moved the service declaration to a separate xml config file, like it was done with fos_rest.

@DominicLuidold You are right. Registering PropertyDescriberInterface for autoconfiguration makes sense. But this is out of scope of this PR.

tests/Functional/Controller/ApiController81.php
stollr: I have removed it, now.

tests/Functional/Controller/ApiController80.php
stollr: I have removed it, now.

tests/Functional/FunctionalTest.php
stollr: This makes sense. I have adjusted the assertions.

tests/PropertyDescriber/UuidPropertyDescriberTest.php
stollr: I have removed it, now.

DjordyKoert:
> @DjordyKoert I have addressed all comments and rebased the PR on master. I hope it is ready for merge, now :-)

Thanks I will try and have a look next week :)

DjordyKoert (approved):

DjordyKoert:
It took a bit, but it's now merged 😅 

Thank you @stollr!

stollr:
You're welcome. Thanks for merging

PR #2105, opened 2023-05-29
feat(#2104): add support for Backed Enum as model in Symfony Forms

### LINKED ISSUE #2104: Add support for Backed Enum as model in Synfony Forms
Currently, there is support for Backed Enum in Entity that is creating component and referencing it in in open api specification.

But when adding [EnumType](https://symfony.com/doc/current/reference/forms/types/enum.html) in symfony form, it's specified as simple enum and not a reusable component, same as [ChoiceType](https://symfony.com/doc/current/reference/forms/types/choice.html), and this is creating a lot of duplicate enum types when generating api clients, one per each form where it's used.

I guess the update should be happening [here](https://github.com/nelmio/NelmioApiDocBundle/blob/ebe6f5917c883c3b3591e8d04df0d7247bcb1b03/ModelDescriber/FormModelDescriber.php#LL202C22-L202C22) (but i'm not sure).

Current State:

Enum:
```
enum ExampleEnum: string
{
    case ONE = 'one';
    case TWO = 'two';
    case THREE = 'three';
}
```

Form:
```
->add('formField', EnumType::class, ['class' => ExampleEnum::class])
```

generated specs:
```
formField:
    type: string
    enum: ['one', 'two', 'three']
```

Requested state and wanted changes:
```
components:
  schemas:
    ExampleEnum:
      type: string
      enum:
        - one
        - two
        - three
```

```
formField:
    $ref: '#/components/schemas/ExampleEnum'
```

or if form field is array of enums (`multiple = true`):
```
formField:
    type: array
    items:
        $ref: '#/components/schemas/ExampleEnum'
```

nikola-jovanovic-php:
* resolves #2104

### DIFF
--- a/ModelDescriber/FormModelDescriber.php
+++ b/ModelDescriber/FormModelDescriber.php
@@ -199,7 +199,7 @@ final class FormModelDescriber implements ModelDescriberInterface, ModelRegistry
                 break;
             }
 
-            if ('enum' === $blockPrefix || $config->getType()->getInnerType()->getBlockPrefix() === 'enum') {
+            if ('enum' === $blockPrefix || 'enum' === $config->getType()->getInnerType()->getBlockPrefix()) {
                 $model = new Model(
                     new Type(Type::BUILTIN_TYPE_OBJECT, !$config->getOption('required'), $config->getOption('class')),
                     null,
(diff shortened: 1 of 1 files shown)


shakaran:
@nikola-jovanovic-php could you rebase with last updates? Thanks

PR #2112, opened 2023-07-07
Add Symfony attribute describers

DjordyKoert:
[Symfony 6.3](https://symfony.com/blog/new-in-symfony-6-3-mapping-request-data-to-typed-objects) added the possibility for developers to map request data to typed objects.  This release also included the addition of [mapping query parameters](https://symfony.com/blog/new-in-symfony-6-3-query-parameters-mapper).

This made me wonder why NelmioApiDocBundle wouldn't be able to automatically create basic documentation when these attributes are present.


phpunit.xml.dist
```diff
@@ -14,7 +14,7 @@
         <env name="SHELL_VERBOSITY" value="-1" />
         <ini name="error_reporting" value="-1" />
         <ini name="memory_limit" value="512M" />
-        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=0" />
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=5" />
```
DjordyKoert, line 17:
```suggestion
        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=0" />
```

This was added to prevent deprecation warnings from being thrown during development.

DjordyKoert:
The `CI / PHPUnit (8.1, 6.3.*)` job also seems to be failing for some unknown reason. I think an update of a package broke a lot of stuff.

DjordyKoert:
Some package update broke `CI / PHPUnit (8.1, 6.3.*)` this is also currently broken on main branch. This should ideally be fixed before merging.

kylekatarnls (approved):

RouteDescriber/SymfonyDescriber.php
```diff
@@ -0,0 +1,161 @@
...
+                    /** @var OA\RequestBody $requestBody */
+                    $requestBody = Util::getChild($operation, OA\RequestBody::class);
+
+                    if (!is_array($attribute->acceptFormat)) {
+                        $this->describeRequestBody($requestBody, $parameter, $attribute->acceptFormat ?? 'json');
+                    } else {
+                        foreach ($attribute->acceptFormat as $format) {
+                            $this->describeRequestBody($requestBody, $parameter, $format);
+                        }
+                    }
```
kylekatarnls, line 40:
```suggestion
                    foreach (((array) ($attribute->acceptFormat ?? 'json')) as $format) {
                        $this->describeRequestBody($requestBody, $parameter, $format);
                    }
```

DjordyKoert:
@Wulfheart I agree that it makes sense that this PR should include `#[MapQueryString]` but I couldn't think of a nice way to implement this at the time of making this PR. I will have a look when I have time to see if I can manage to do it.

Wulfheart:
@DjordyKoert you could either try to handle it like a response and create a `$ref` or add all query parameters based on the DTO I guess.

DjordyKoert:
@Wulfheart I have added the `#[MapQueryString]` into this PR. Could you check if everything is working now as you expected? 😄

DjordyKoert:
@GuilhemN this PR should be fully done now for review. 

I would also appreciate it if you or someone else could take a look at https://github.com/nelmio/NelmioApiDocBundle/pull/2126 so all phpunit checks can atleast be ran through github actions instead of permanently failing.

BafS:
Thank you for this PR!

GuilhemN:
I'm very sorry for the delay, I don't really have the time to review this right now as I'm on a rather long vacation.

But with a very quick look, it feels you put a lot of effort into the code quality and documentation, and others approved it. 
I merged the PR fixing the tests, and if that's ok for you I can just trust you on this one and merge it. Let me know 😉

Otherwise, I think I'll only be able to review it in at least a month.

DjordyKoert:
@GuilhemN You can kinda forget my above comment. As far as I have seen it isn't possible to use groups together with `MapRequestPayload` because Symfony will just throw a validation failure response since the data isn't complete. So everything should be working as intended. 

Still an FYI that users who use `MapRequestPayload` together with `OA\RequestBody`, as written in my above comment, will result in an error being thrown while generating the documentation. 

I guess this would require a major version since this would be a breaking change for some users? But I will let you decide whatever suits best for this situation.

So in my opinion this could be merged as I myself have also tested this PR on my own (fairly big) project + unit tests.

RouteDescriber/SymfonyAnnotationDescriber/SymfonyMapRequestPayloadDescriber.php
```diff
@@ -0,0 +1,92 @@
...
+
+    public function describe(OA\OpenApi $api, OA\Operation $operation, ReflectionParameter $parameter): void
+    {
+        $attribute = SymfonyAnnotationHelper::getAttribute($parameter, MapRequestPayload::class);
+
+        /** @var OA\RequestBody $requestBody */
+        $requestBody = Util::getChild($operation, OA\RequestBody::class);
+        SymfonyAnnotationHelper::modifyAnnotationValue($requestBody, 'required', !($parameter->isDefaultValueAvailable() || $parameter->allowsNull()));
+
+        $formats = $attribute->acceptFormat;
```
junaidbinfarooq, line 33: It may be that `describe` is only called if `supports` returns `true` but wouldn't this lead to a _null-pointer exception_, at least at the compile time/static analysis level, considering `$attribute` could either be an instance of the required attribute or `null`??

RouteDescriber/SymfonyAnnotationDescriber/SymfonyMapRequestPayloadDescriber.php
```diff
@@ -0,0 +1,92 @@
...
+        SymfonyAnnotationHelper::modifyAnnotationValue($contentSchema, 'type', 'object');
+
+        $schema = Util::getProperty($contentSchema, $parameter->getName());
+
+        SymfonyAnnotationHelper::describeCommonSchemaFromParameter($schema, $parameter);
+    }
+
+    private function getContentSchemaForType(OA\RequestBody $requestBody, string $type): OA\Schema
+    {
+        SymfonyAnnotationHelper::modifyAnnotationValue($requestBody, 'content', []);
```
junaidbinfarooq, line 56: A general remark. Private methods, especially those that aren't too small, can be implicit dependencies and hence demand their dedicated classes/services. This may apply to these private methods too.

akankov:
Hey guys! Do you have any plans to merge this PR?

DjordyKoert:
@GuilhemN can you please take a look? 😄

SBNTT:
Hello.
Thanks for your work.

Was on [752b834](https://github.com/nelmio/NelmioApiDocBundle/pull/2112/commits/752b83469cc1d1997c08842dc240680f22ea8d7d) for for weeks and it had been working quite well until now.

I updated to your latest commits, but unfortunately, it brokes something.
BackedEnum are not correctly represented in the generated specs.

This was introduced while merging master into this branch ([f513fe2](https://github.com/nelmio/NelmioApiDocBundle/pull/2112/commits/f513fe2ade70703c8fe732d12d4470013a6d07c1)).

---

The controller:
```php
public function handler(
    #[MapQueryString]
    ?MyModel $input,
): JsonResponse { /* ... */ }
```

The Model class:
```php
readonly class MyModel
{
    public function __construct(
        public ?MyEnum $myEnum,
    ) {}
}
```

---

I computed the diff between working and broken specs:
```diff
{
    "in": "query",
    "required": false,
    "schema": {
        "nullable": true,
+       "oneOf": [
-       "allOf": [
            {
+               "$ref": "#/components/schemas/MyEnum"
-               "$ref": "#\/components\/schemas\/MyEnum"
            }
        ],
        "property": "myEnum"
    }
    }
},
```
- Models now include a `required` array (presumably nullsafety purposes). Not related imo
- All `/` were previously escaped (`"$ref": "#\/components\/schemas\/MyModel"` => `"$ref": "#/components/schemas/MyModel"`). I've no idea about the potential impact ..
- Some `allOf` are now replaced with `oneOf` (not only for BackedEnum !).

DjordyKoert:
What exactly isn't working here @SBNTT? Your generated spec seems fine.
> Models now include a required array (presumably nullsafety purposes). Not related imo
Introduced in https://github.com/nelmio/NelmioApiDocBundle/pull/2141

>All / were previously escaped ("$ref": "#\/components\/schemas\/MyModel" => "$ref": "#/components/schemas/MyModel"). I've no idea about the potential impact ..
This was fixed in https://github.com/nelmio/NelmioApiDocBundle/pull/2157

> Some allOf are now replaced with oneOf (not only for BackedEnum !).
Introduced in https://github.com/nelmio/NelmioApiDocBundle/pull/2156

SBNTT:
Neither swagger editor nor https://github.com/OpenAPITools/openapi-generator behave correctly

With `oneOf` instead of `allOf`:
- Swagger editor interprets it as a basic text field, instead of a select input with all enum values
- Openapi generator outputs broken client code

Isn't this representation more suitable than using `oneOf` or `allOf` ?
I mean, `oneOf`, `allOf`etc aims to combine multiple schemas, but in this case there is only one
```diff
        "name": "myEnum",
        "in": "query",
        "required": false,
        "schema": {
            "nullable": true,
+           "$ref": "#/components/schemas/TypologyKindModel",
-            "oneOf": [
-                {
-                    "$ref": "#/components/schemas/TypologyKindModel"
-                }
-            ]
        }
```

DjordyKoert:
Ah yup that would make more sense, that is also the way it is described to do it here https://swagger.io/docs/specification/data-models/enums/. 

But I also think this is just a problem with the way enums are currently described as can be seen here: https://github.com/nelmio/NelmioApiDocBundle/blob/ed742a7650bb3b8c70bda102876a184678fa2447/PropertyDescriber/ObjectPropertyDescriber.php#L40-L46
Enum's should not be handled here, because they are quite simple schema's. I will go ahead and create a different issue for this.

SBNTT:
`allOf` representation is back and fixed the issue. Is it temporary until swapper-php issue is solved ?

You also fixed another issue (weird duplicated Models). That's cool !

There is only one remaining issue:
The `schema` object is not supposed to have a `property` property as indicated in [https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#schemaObject](). This value is already set by the `parameter` object's `name` property (https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#parameterObject)

DjordyKoert:
> `allOf` representation is back and fixed the issue. Is it temporary until swapper-php issue is solved ?

Yup this is hopefully temporary, this was the fix for now. https://github.com/zircote/swagger-php/issues/1528

> There is only one remaining issue: The `schema` object is not supposed to have a `property` property as indicated in https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#schemaObject. This value is already set by the `parameter` object's `name` property ([OAI/OpenAPI-Specification@`main`/versions/3.0.3.md#parameterObject](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md?rgh-link-date=2024-01-05T16%3A03%3A20Z#parameterObject))

Yup thats one of the things still on my 'todo' list. The other one is finding a way to prevent breaking the docs generation whenever `#[OA\RequestBody(content: new Model(...)]` is used in combination with `MapRequestPayload`.

DjordyKoert:
@SBNTT Managed to fix everything + tests for all situations 😄

SBNTT:
Yaay !
I've just tested it, and it's working perfectly now.

For months, we had a script that was postprocessing the generated OA specs using some ugly regexps, but you've fixed everything :)

Thanks for your hard work @DjordyKoert !

BafS:
Thanks a lot @DjordyKoert for this amazing PR, I hope to see it merged soon

DjordyKoert:
@akankov thank you, fixed it in https://github.com/nelmio/NelmioApiDocBundle/pull/2189

PR #2121, opened 2023-07-27
Fix enum backing type always 'string'

DidierLmn:
Every enum will be described as type "string" because the current logic to determine the backing type does not work
```
is_subclass_of($enumClass, \StringBackedEnum::class));
is_subclass_of($enumClass, \IntBackedEnum::class));
```
These are always false, regardless of backing type as you can see at https://3v4l.org/fButY

Moreover the type would be set as "int", which is not valid in OpenApi-Spec.

This MR ensures proper backing detection of enums and sets int-backed enums to type "integer".


palgalik (approved):

ModelDescriber/EnumModelDescriber.php
```diff
@@ -17,6 +17,12 @@ public function describe(Model $model, Schema $schema)
             $enums[] = $enumCase->value;
         }
 
+        $reflectionEnum = new \ReflectionEnum($enumClass);
+        if ($reflectionEnum->isBacked() && 'int' === $reflectionEnum->getBackingType()->getName()) {
+            $schema->type = 'integer';
+        } else {
+            $schema->type = 'string';
+        }
         $schema->type = is_subclass_of($enumClass, \IntBackedEnum::class) ? 'int' : 'string';
```
DjordyKoert, line 26: I'm not really a fan of the if-else combined with the multi places where `$schema->type` gets set.

```suggestion
        $schema->type = is_subclass_of($enumClass, \IntBackedEnum::class) || ($reflectionEnum->isBacked() && 'int' === $reflectionEnum->getBackingType()->getName())
            ? 'integer' 
            : 'string';
```

DjordyKoert:
Also please rebase on master for CI

ModelDescriber/EnumModelDescriber.php
Book56: Adil Ishaq

Tests/Functional/Entity/ArticleType81NotBacked.php
Book56, line 1: Adil Ishaq

DjordyKoert:
Thank you @DidierLmn!

DidierLmn:
I kinda lost track of this one. Thanks a lot for finishing it off and merging @DjordyKoert!!  :pray:

PR #2129, opened 2023-08-16
add support for versioning

goetas:
This pull request allows to define a versioning strategy per model.

im not 100% sure it is the best approach. a different strategy is to have different versions of the API defined using "areas" (that would allow also to not clutter the latest api doc with "old" version infos.

The question is: one document per version or a single document with versions handled by media types?

reason i prefer this approach is (instead of areas):
-  you do not need to send to the user multiple urls for different api versions
- the user still sees the old data models (eventually described as deprecated)


shakaran:
@goetas could you rebase with last updates? Thanks

lacpandore:
maybe a line for the doc too, to be used ;)

PR #2131, opened 2023-08-19
connect the OpenAPI Context with Loggers from Symfony

### LINKED ISSUE #2077: Connect the OpenAPI Context with Loggers from Symfony
Seems like the OpenAPI library's `Context` has a logger: https://github.com/zircote/swagger-php/blob/a060ea2494ae6336eb4481f5c276d7c7ef038eda/src/Context.php#L66

that is uses a few places: https://github.com/zircote/swagger-php/blob/a060ea2494ae6336eb4481f5c276d7c7ef038eda/src/Analysers/AttributeAnnotationFactory.php#L47

This bundle should connect the logger from Symfony to the context. Name here: https://github.com/nelmio/NelmioApiDocBundle/blob/b4df34cdf78cf42c02afb85f9d10181bbf835da5/Describer/OpenApiPhpDescriber.php which already has the logger.

I can pick this up later, but it'd be a create "first issue" for anyone looking to contribute. I just had an error when rendering API docs that was silent and hard to debug becuase the default logger seemed to not work as expected.

magrigry:
Fix #2077. Very simple but useful change because there is no other way achieve this e.g by decorating `ApiDocGenerator`. 

The patch sound quiet simple and was already suggested in the issue. Did I miss something ?


GuilhemN:
Sounds good! I'm wondering however if this doesn't create a lot of warning logs and noise, it would be nice if several people could give it a try and give a feedback

magrigry:
My goal is actually to prevent noise and logs spamming. 

When `null` is passed (the actual behavior), PHP Swagger `DefaultLogger` is used. This logger calls php `trigger_error()` function https://github.com/zircote/swagger-php/blob/434cd0a555b6672b6d609c6f3f6437739df4469d/src/Loggers/DefaultLogger.php#L31, which create a lot of logs and noise and there is no way to have any control about those error are logger, because it would impact others warning / errors triggered with `trigger_error()`.

With this MR, you can just pass e.g. `Psr\NullLogger` https://github.com/php-fig/log/blob/master/src/NullLogger.php and everything is silenced, or use any logger which is PSR compliant to control where and how warning emitted are logged.

GuilhemN:
Thank you @magrigry!

Issue #2120, opened 2023-07-26
Show CSRF token on Symfony forms


stollr:
The detection and documentation of forms is a great feature!

One thing I am missing is the documentation of the CRSF token if it is enabled in the form.

PR #411, opened 2014-06-17
Csrf support

jonmchan:
Adds CSRF Token to ApiDoc so that we can submit forms that have CSRF protection. This PR checks if CSRF is enabled, and if it, generates the proper CSRF token. 

Requires https://github.com/nelmio/NelmioApiDocBundle/pull/408 to be fixed and merged first.

PR #2132, opened 2023-09-12
Support CSRF protection token of Symfony's form component

stollr:
This PR adds support of the CSRF protection token which is created by the Symfony form component if the `csrf_protection` option is enabled.

As soon as this option is enabled a corresponding property will be added to the schema. 

This PR addresses #2120.

There already was another PR (#411) some years ago, but it wasn't merged. I'd be happy if this could be accepted.


stollr:
@GuilhemN I'd be really happy if you could have a look at this addition.

DjordyKoert:
I would also like to see a functional test for this https://github.com/nelmio/NelmioApiDocBundle/tree/master/Tests/Functional/Form  😄

DjordyKoert:
Thank you, could you please also write a test which shows the behaviour when using `configureOptions()` https://symfony.com/doc/current/security/csrf.html#csrf-protection-in-symfony-forms 😄

stollr:
@DjordyKoert I have added some edge case handling and more tests. What do you think now?

DjordyKoert:
Thank you @stollr!

PR #2144, opened 2023-11-07
Add CI jobs for PHP 8.2 and 8.3

derrabus:
This PR adds test coverage for PHP 8.2 and 8.3 and bumps several GitHub actions.

I believe that the bundle should be tested against the current and the next PHP minor. However, I'm not sure if we really need the full test matrix here. Please advice if I should remove any of the jobs.


derrabus:
The test failure seems to be unrelated to my changes. I can reproduce the issue locally as well.

GuilhemN:
Thank you @derrabus! I just modified the matrix to test on PHP 7.3 and 7.4 and removed PHP 7.2.
Best wishes for the new year.

PR #2148, opened 2023-11-15
Fix wrong detection of OA min/max for symfony contraints with datetime strings

Tobion:
symfony range and comparison constraints can compare datetime strings, see https://github.com/symfony/validator/blob/0dbfc4b9b87a0e8245ff37d85cbe6b6ef677999f/Constraints/AbstractComparisonValidator.php#L68

If the constraints contain a comparison value that is not int (e.g. datetime string), we cannot infer a OpenAPI value for it. Previously it falsely casted those to int resulting in nonesense like minimum=0 and maximum=0 for string values.


DjordyKoert:
Looks good. could you also write a test for the situation where a string is passed to any of these attributes?

DjordyKoert:
Thanks @Tobion

Issue #1721, opened 2020-09-23
[feature] @OA\RequestBody, but for "query"


arjanfrans:
**Version** 
I am using  version `4.0@dev`.

**Problem**
I define all my request parameters in a "Model" class. By using the `@OA\RequestBody` annotation I get my parameters documented automatically. This works great for all POST/PATCH/PUT requests that have a body. For GET requests I also like to define my parameters in a Model class.

As a "workaround" to document the endpoints I put "query" in the description. However, I can then not use the "Try it out" feature. Here is my implementation:

![image](https://user-images.githubusercontent.com/3588470/94018362-d2260580-fdb0-11ea-9af9-2f53f11511d5.png)
![image](https://user-images.githubusercontent.com/3588470/94022109-0bf90b00-fdb5-11ea-8ee3-f8e202eacd87.png)

**Feature**
What I'd like to see is an `@OA\RequestBody` annotation for query parameters (`@OA\RequestQuery`?) so that I can define my query parameters in my Model class.

My Model classes contains the types and validations, one place where everything is defined. I don't want to define `OA\Parameter` annotations on my route too.

GuilhemN:
This is actually supported using openapi 3 features, please see https://github.com/nelmio/NelmioApiDocBundle/issues/1596#issuecomment-650008083.

In summary, you should use an usual `@Parameter` with `@Model` and the `explode` field.

GuilhemN:
Closing as fixed, feel free to reopen if this is not the case.

Luehrsen:
Does anybody have a working implementation for this? 

Currently doing 
```
     * @OA\Parameter(
     *   name="query",
     *   in="query",
     *   style="form",
     *   explode=false,
     *   @OA\Schema(
     *     ref=@Model(type=CharacterFilter::class, groups={"read"})
     *   )
     * )
```

and getting 

<img width="453" alt="Bildschirmfoto 2022-01-06 um 21 29 08" src="https://user-images.githubusercontent.com/1105871/148447355-274e8ef5-e79c-4d5d-ac91-4002eb2bea18.png">

Side question: Does the bundle pick up parameters that are in traits?

PetrNovotnyGS:
> Does anybody have a working implementation for this?
> 
> Currently doing
> 
> ```
>      * @OA\Parameter(
>      *   name="query",
>      *   in="query",
>      *   style="form",
>      *   explode=false,
>      *   @OA\Schema(
[...]

PR #2149, opened 2023-11-16
Allow `custom processors` through tagged services

### LINKED ISSUE #2146: Allow `custom Processors` for `swagger-php`
There are several issues that require somewhat custom behavior, mostly for convenience in managing huge codebases. A lot of those would be solved with the option of a custom processor. (see #1871, #1721)

Currently, there is no way to register custom processors in the `NelmioApiDocBundle`. It would be nice to be able to hook into the `setProcessors` function somewhere here: https://github.com/nelmio/NelmioApiDocBundle/blob/bd10c342bfe7aae3efe864be2708cd93b01da641/ApiDocGenerator.php#L87-L92

Ideally we could register own processors through the `nelmio_api_doc.yaml` config file:
```yaml
# config/packages/nelmio_api_doc.yaml
nelmio_api_doc:
    areas:
    documentation:
    processors:
      - /App/OpenApi/Processors/CustomProcessor1
      - /App/OpenApi/Processors/CustomProcessor2
```

For code quality and clarity I would suggest enforcing these processors to implement `ProcessorInterface`.

Luehrsen:
This pull request aims to provide a solution to register custom [processors](https://zircote.github.io/swagger-php/reference/processors.html) for `swagger-php` through services with the tag `swagger.processor`.

fixes #2146


DependencyInjection/Compiler/CustomProcessorPass.php
```diff
@@ -0,0 +1,60 @@
...
+        $definitions = [];
+        // Find all registered generator services.
+        foreach($container->getDefinitions() as $id => $definition) {
+            // Check if the service-id contains the tag 'nelmio_api_doc.generator'
+            if (strpos($id, 'nelmio_api_doc.generator.') === false) {
+                continue;
+            }
+
+            $definitions[$id] = $definition;
+        }
```
Nyholm, line 43: I dont see that this code it used.

Nyholm (changes requested):
Thank you. 

I like it. It works. But what if we want to add a processor in the middle of the list? 

Could we register `Generator` as a service and use constructor injection to add that service to `ApiDocGenerator`? If so, we would be able to re-use `Generator::addProcessor($$processor, ?string $before = null)`.

Luehrsen:
Thank you for your feedback!

The 6.3.* test keep failing on me, but I think that's a general issue at the moment.

I think as of now we have a generally working solution for custom processors in `swagger-php`. 

One thing I noticed though, is that a lot of what the API Docs Bundle does is abstracted away from the processors, particularly anything relating the `ModelRegistry` and `ModelRegister`. There is a lot of logic happening there, that is inaccessible for custom processors at this point. (At least not with my skills.)

That means, that while this is a nice addition for the Bundle in general, it would not solve https://github.com/nelmio/NelmioApiDocBundle/issues/1721 for example, because you need access to the '@Model' and refs to do that.

I'm currently thinking about if we could find a way to pipe the ModelRegistry into a processor if needed.

DependencyInjection/Compiler/CustomProcessorPass.php
```diff
@@ -0,0 +1,53 @@
...
+     * Process services tagged as 'swagger.processor'.
+     *
+     * @param ContainerBuilder $container The container builder
+     */
+    public function process(ContainerBuilder $container): void
+    {
+        // Find the OpenAPI generator service.
+        $definition = $container->findDefinition('nelmio_api_doc.open_api.generator');
+
+        foreach ($container->findTaggedServiceIds('swagger.processor') as $id => $tags) {
```
Nyholm, line 35: Maybe prefix the tag with nelmio_api

Nyholm (approved):
This is awesome. 

I think this is great. Thank you for working on this.

DependencyInjection/Compiler/CustomProcessorPass.php
Luehrsen: done

Luehrsen:
I think it should be in the scope of this PR to extend the `ModelRegister` a bit to attach some more useful information to the transformed annotations, so that the processors have some "meat" to work on. 

Does that make sens?

Luehrsen:
> I think it should be in the scope of this PR to extend the `ModelRegister` a bit to attach some more useful information to the transformed annotations, so that the processors have some "meat" to work on.
> 
> Does that make sense?

To answer my own question: Nope! I got it without touching the ModelRegister.

The secret sauce is to use `Nelmio\ApiDocBundle\OpenApiPhp\Util::getSchema` to get all the properties you desire.

So this would be an example of the `__invoke` function in a custom processor registered for `swagger-php`. (Dropping keywords here for search.)

```php

    const X_QUERY_AGS_REF = 'query-args-explode';

    public function __invoke(Analysis $analysis)
    {
	$this->analysis = $analysis;

        /** @var OA\Parameter[] $schemas */
        $parameters = $analysis->getAnnotationsOfType(OA\Parameter::class);

        foreach ($parameters as $parameter) {
            if ($parameter->x !== GENERATOR::UNDEFINED && array_key_exists(self::X_QUERY_AGS_REF, $parameter->x)) {
		// Generate the name from the parameter schema.
		$name = str_replace(OA\Components::SCHEMA_REF, '', $parameter->schema->ref);
		$schema = Util::getSchema($this->analysis->openapi, $name);
            }
        }
    }
```

This would result into `$schema` being a proper `OpenApi\Annotations\Schema` class with all the relevant properties attached. (So essentially the object the `ref` is pointing at. 

It was registered like this in a controller:
```php
	#[OA\Parameter(
		in: 'query',
		name: 'query',
		explode: true,
		style: 'form',
		schema: new OA\Schema(
			type: 'object',
			ref: new Model(
				type: TeamFilter::class,
				groups: ['read']
			)
		),
		x: [
			'query-args-explode' => true
		]
	)]
```

So to reiterate: If you want to access `Nelmio\ApiDocBundle\Annotation\Model` data inside a custom processor do this:

1) Find the property where the model was used. 
2) You'll see, that the model was replaced with an almost empty `Schema` object that now has a `ref` that points to your model, that has been processed by the `ModelRegistry`. 
3) To get the 'name' of your model as registered in the system, just remove the `OA\Components::SCHEMA_REF` from your `ref`. 
4) The base OpenAPI object can be found at `$analysis->openapi`. 
5) Use these two parameters inside `Nelmio\ApiDocBundle\OpenApiPhp\Util::getSchema` to retrieve the glory of your model.

DjordyKoert (changes requested):

ApiDocGenerator.php
```diff
@@ -48,16 +49,20 @@ final class ApiDocGenerator
...
 
+    /** @var Generator */
+    private $generator;
+
     /**
      * @param DescriberInterface[]|iterable      $describers
      * @param ModelDescriberInterface[]|iterable $modelDescribers
      */
-    public function __construct($describers, $modelDescribers, CacheItemPoolInterface $cacheItemPool = null, string $cacheItemId = null)
+    public function __construct($describers, $modelDescribers, Generator $generator, CacheItemPoolInterface $cacheItemPool = null, string $cacheItemId = null)
```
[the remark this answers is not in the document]
Luehrsen: Then we have to update the [container registration](https://github.com/nelmio/NelmioApiDocBundle/blob/3405e7b3beb5fae865f7dd52a1fce5e8013c95bd/DependencyInjection/NelmioApiDocExtension.php#L87) as well. Just changing this line leads to an error. 

I'm not quite sure how one would do that. Would it just be:

```php
                ->setArguments([
                    null,
                    null,
                    new Reference('nelmio_api_doc.open_api.generator'),
                ]);
```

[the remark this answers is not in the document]
DjordyKoert: Ye that seems fine.

ApiDocGenerator.php
```diff
@@ -48,16 +49,20 @@ final class ApiDocGenerator
...
      */
-    public function __construct($describers, $modelDescribers, CacheItemPoolInterface $cacheItemPool = null, string $cacheItemId = null)
+    public function __construct($describers, $modelDescribers, CacheItemPoolInterface $cacheItemPool = null, string $cacheItemId = null, Generator $generator = new Generator())
     {
         $this->describers = $describers;
         $this->modelDescribers = $modelDescribers;
         $this->cacheItemPool = $cacheItemPool;
         $this->cacheItemId = $cacheItemId;
+        $this->generator = $generator;
     }
```
DjordyKoert, line 66: I made a small mistake with my previous suggestion, instantiating a new class like this is only supported since PHP 8.1.0.

```suggestion
    public function __construct($describers, $modelDescribers, CacheItemPoolInterface $cacheItemPool = null, string $cacheItemId = null, Generator $generator = null)
    {
        $this->describers = $describers;
        $this->modelDescribers = $modelDescribers;
        $this->cacheItemPool = $cacheItemPool;
        $this->cacheItemId = $cacheItemId;
        $this->generator = $generator ?? new Generator();
    }
```

DjordyKoert (changes requested):

DjordyKoert (approved):

Nyholm:
I've tested this in a SF 6 application. It works as expected. 

Thank you!

DjordyKoert:
Thank you @Luehrsen 😄

PR #2156, opened 2023-11-23
fix: use oneOf instead of allOf

DjordyKoert:
Fix when generating a nullable property (an enum for example) in OpenApi 3.1.0.

swagger-php makes sure that the `nullable` property is unset when working with OpenApi 3.1.0 and transform it to the 3.1.0 equivelant (see: https://github.com/zircote/swagger-php/blob/595d8b63a589e6ac0835c83a45304c7397dc37d1/src/Annotations/AbstractAnnotation.php#L390-L402), but this doesn't work when using `allOf`.

This PR aims to fix that by replacing `allOf` with `oneOf`. Additionally `oneOf` is more logical to use in this context than `allOf`  since `allOf` means that every schema could be valid where `oneOf` means that only one schema can be valid https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/.


GuilhemN:
Nice, thank you!

PR #2162, opened 2023-12-05
fix CI

smoench:
This PR adjusted following:
- uses ubuntu-22.04 as latest runner image
- uses actions/checkout@v4 to prevent deprecations
- uses actions/cache@v3 to prevent deprecations
- replaces [deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) `set-ouput` with `$GITHUB_OUTPUT`
- adds `flex` via  `shivammathur/setup-php` tools
- uses direct dev dependency for `phpunit` as it faster, cached by actions composer cache and fixes an error on PHP 7.2
- adds Symfony 6.4 to the test matrix and updates deprecation baseline


DjordyKoert:
Nice!

.github/workflows/continuous-integration.yml
```diff
@@ -32,28 +33,28 @@ jobs:
             symfony-require: "5.4.*"
           - php-version: 8.1
             symfony-require: "5.4.*"
-          - php-version: 8.1
-            symfony-require: "6.3.*"
+          - php-version: 8.2
+            symfony-require: "6.4.*"
```
DjordyKoert, line 37: Php 8.1 is still supported by this version, we could also add a test for this

DjordyKoert (approved):

DjordyKoert:
@GuilhemN ci fails because a deprecation warning is not present in the phpunit baseline. Generating a new baseline should fix it.

> 1x: The "Sensio\Bundle\FrameworkExtraBundle\Routing\AnnotatedRouteControllerLoader" class is deprecated since version 5.2. Use "Symfony\Bundle\FrameworkBundle\Routing\AnnotatedRouteControllerLoader" instead.

GuilhemN:
Thanks a lot @smoench! Best wishes for the new year.

DjordyKoert:
Also small question from my end. Are you planning to make a new release for the Symfony 7 support after this is merged?

Issue #2136, opened 2023-09-15
Migrate to PHP 8 - PHP 7 unmaintained


DjordyKoert:
Support for PHP 7.4 has stopped 9 months ago. NelmioApiDocBundle still supports PHP 7.2 and higher so it's probably a good idea to migrate to PHP 8 and [Symfony 6](https://symfony.com/releases/6.0) since Symfony 6 only supports PHP 8.0.2 or higher.

This would also improve maintainability of this project since backwards compatibility with older PHP versions (read PHP 7) wouldn't need to be kept in mind.

Thoughts?

GuilhemN:
I'm not sure, as long as we don't have a specific new feature or fix where using PHP 8 would help significantly I prefer not to change the constraint.
It's not a very big deal on our side I think, and if it can help some developers maintain NelmioApiDocBundle up to date I think it's worth it.

DjordyKoert:
Well yeah for now it doesn't look like PHP 8 is needed to be able to bring specific feature/fix to NelmioApiDocBundle, the main thing it would do is ease maintenance of this package as far as I can tell.

It could for example remove some boilerplate code because of the introduction of promoted constructor properties.

DjordyKoert:
Done in https://github.com/nelmio/NelmioApiDocBundle/issues/2215

PR #2171, opened 2024-01-02
bump minimum version to php 7.4

DjordyKoert:
Related to https://github.com/nelmio/NelmioApiDocBundle/issues/2136.

According to Packagist less than 1% use PHP 7.2 & less than 1% use PHP 7.3. https://packagist.org/packages/nelmio/api-doc-bundle/php-stats#4

Around 20% of installs still happen on PHP 7.4 which is the reason why I still would like to support this version for now.


Describer/ModelRegistryAwareTrait.php
```diff
@@ -15,10 +15,7 @@
 
 trait ModelRegistryAwareTrait
 {
-    /**
-     * @var ModelRegistry
-     */
-    private $modelRegistry;
+    private ModelRegistry $modelRegistry;
```
kevinpapst, line 18: I would probably initialize this with null to be BC safe, as it is not initialized in the constructor.

DjordyKoert: Good find, will do!

kevinpapst:
What is the release policy for this bundle? 
Does bumping the required PHP version need a new major version of the bundle?

DjordyKoert:
> What is the release policy for this bundle? Does bumping the required PHP version need a new major version of the bundle?

This would result in a major version bump

Describer/ModelRegistryAwareTrait.php
DjordyKoert: Hmm but then again, it was assumed that `$modelRegistry` that it could not be `null` according to the docblock. It this would be the case then existing code should already be failing to work is what I would assume. 

Since this PR would require a major bump anyways then I think we could be safe keeping it like it is.

kevinpapst: You could access uninitialized properties without type declaration before ... and I think they were treated as null(?). Now with the type you cannot read the property without the setter being called first, or it will raise an Exception.
Anyway, this is a minor issue and if the framework code makes sure that the setter is always called first, there shouldn't be a problem.

DjordyKoert: Yup NelmioApiDocBundle already call the setter whenever needed, so this should never happen

chrisguitarguy (approved):

composer.json
```diff
@@ -25,7 +25,8 @@
         "symfony/options-resolver": "^5.4|^6.0|^7.0",
         "symfony/property-info": "^5.4|^6.0|^7.0",
         "symfony/routing": "^5.4|^6.0|^7.0",
-        "zircote/swagger-php": "^4.2.15",
+        "zircote/swagger-php": "^4.6.1",
+        "phpstan/phpdoc-parser": "^1.0",
```
kevinpapst, line 29: Why in require? Is it used somewhere?

DjordyKoert: This is related to failing https://github.com/nelmio/NelmioApiDocBundle/actions/runs/7767859362/job/21185134733?pr=2171

Might make more sense to pin this only in dev dependency

DjordyKoert: Tested it, this does not work because the service gets removed

kevinpapst: I don't get it, are you adding a new dependency for every using project, because on 7.4 some tests are failing? Sounds wrong to me. If you plan a new major version anyway, kick 7.4.

DjordyKoert: I agree that this is wrong. Still need to decide about what to do with this. Ideally I would go for PHP 8.1 & Symfony >6

kevinpapst: I'd always vote for moving forward, but hard to decide for a library, if 20% of all installations are still using PHP 7.4.

DjordyKoert: I had another look at https://packagist.org/packages/nelmio/api-doc-bundle/php-stats#4 and now it report 10% of installs happen on PHP 7.4

I am considering that PHP 8.1 & Symfony >6 isn't that bad of an idea after all. Also together with the fact that we can still keep releasing fixes for `v4.*` if needed. This would also really help with cleaning up the codebase to new standards and start using [phpstan](https://phpstan.org/).

DjordyKoert: Anyways thanks a lot for your feedback @kevinpapst 😄

kevinpapst: Oh yes, there was a steep drop in January. You have my go 😁 but I have no project on 7, so I am a terrible advisor here. IMO these old EOL versions of PHP can still use old releases. 

Other libs might have commercial interests and aim for maximum reach. But this is OSS and maintaining a free library should be fun! PHPStan is a huge bonus and if it keeps your spirits up, then it is worth doing so.

Issue #2172, opened 2024-01-03
v4.15.1 break if symfony/serializer not installed


magnetik:
Hey,

I just updated and it seems that symfony/serializer is now required. 

```
Error: Class "Symfony\Component\Serializer\Normalizer\AbstractNormalizer" not found
/app/vendor/nelmio/api-doc-bundle/Model/Model.php:50
/app/vendor/nelmio/api-doc-bundle/Model/Model.php:63
/app/vendor/nelmio/api-doc-bundle/Model/ModelRegistry.php:60
/app/vendor/nelmio/api-doc-bundle/OpenApiPhp/ModelRegister.php:48
/app/vendor/nelmio/api-doc-bundle/ApiDocGenerator.php:126
/app/vendor/nelmio/api-doc-bundle/Render/RenderOpenApi.php:76
/app/vendor/nelmio/api-doc-bundle/Render/RenderOpenApi.php:61
/app/vendor/nelmio/api-doc-bundle/Controller/SwaggerUiController.php:37
/app/vendor/symfony/http-kernel/HttpKernel.php:181
/app/vendor/symfony/http-kernel/HttpKernel.php:76
/app/vendor/symfony/http-kernel/Kernel.php:197
/app/vendor/symfony/http-kernel/HttpKernelBrowser.php:65
/app/vendor/symfony/framework-bundle/KernelBrowser.php:171
/app/vendor/symfony/browser-kit/AbstractBrowser.php:403
/app/tests/Controller/DocControllerTest.php:18
```

@DjordyKoert might be related to your latest changes (thanks for them <3)

DjordyKoert:
woops, thanks for mentioning it!

DjordyKoert:
Should be fixed in [`v4.15.2`](https://github.com/nelmio/NelmioApiDocBundle/releases/tag/v4.15.2)

magnetik:
Awesome, thanks for the fast fix !

PR #2174, opened 2024-01-03
use string instead of constant

DjordyKoert:
fixes https://github.com/nelmio/NelmioApiDocBundle/issues/2172


magnetik:
I don't know the in and out of symfony serializer, maybe we can use class_exist to use the constant if it exists and fallback to string?

PR #2141, opened 2023-10-27
Generate non-nullable properties on model as required

DjordyKoert:
Fix for issue: https://github.com/nelmio/NelmioApiDocBundle/issues/2111

PR #2181, opened 2024-01-06
fix: do not mark properties with default as required

### LINKED ISSUE #2179: Properties with default values are required
```php
namespace App;

class FooBar
{
    /**
     * @OA\Property( default="emtpy string ('')",
     *     description="text that should be displayed in monitoring for the status code (when it will be detected as error)")
     * @Assert\Type(type="string")
     */
    private string $foo = '';

    /**
     * @Assert\Valid()
     *
     * @var string[]
     */
    private array $bar;

    public function __construct()
    {
        $this->bar = [];
    }
}
```
Before #2141 with a class definition like above the properties `$foo` and `$bar` where not marked as required. Now they are.

DjordyKoert:
fixes #2179

### DIFF
--- a/PropertyDescriber/NullablePropertyTrait.php
+++ b/PropertyDescriber/NullablePropertyTrait.php
@@ -39,6 +39,11 @@ trait NullablePropertyTrait
                 return;
             }
 
+            // Do not describe default values as required
+            if (Generator::UNDEFINED !== $property->default) {
+                return;
+            }
+
             $existingRequiredFields = Generator::UNDEFINED !== $schema->required ? $schema->required : [];
             $existingRequiredFields[] = $propertyName;
 
--- a/Tests/Functional/Entity/User.php
+++ b/Tests/Functional/Entity/User.php
@@ -65,7 +65,7 @@ if (TestKernel::isAnnotationsAvailable()) {
          *
          * @OA\Property(default = 0.0)
          */
-        private $money;
+        private $money = 0.0;
 
         /**
          * @var \DateTime
@@ -209,7 +209,7 @@ if (TestKernel::isAnnotationsAvailable()) {
          * @var float
          */
         #[\OpenApi\Attributes\Property(default: 0.0)]
-        private $money;
+        private $money = 0.0;
 
         /**
          * @var \DateTime
--- a/Tests/Functional/FunctionalTest.php
+++ b/Tests/Functional/FunctionalTest.php
@@ -247,9 +247,6 @@ class FunctionalTest extends WebTestCase
                 ],
                 'schema' => 'User',
                 'required' => [
-                    'id',
-                    'roles',
-                    'money',
                     'creationDate',
                     'users',
                     'status',


PropertyDescriber/NullablePropertyTrait.php
```diff
@@ -39,6 +39,11 @@ protected function setNullableProperty(Type $type, OA\Schema $property, ?OA\Sche
                 return;
             }
 
+            // Do not describe default values as required
+            if (Generator::UNDEFINED !== $property->default) {
```
heiglandreas, line 43: This will only work for properties where the default has been set via an annotation or attribute. Not for assigned defaults. Or am I missing something?

DjordyKoert: that's correct. I am attempting to look for a better solution

heiglandreas: Feel free to use the tests that I have already created in #2186

DjordyKoert: Thank you ❤️

heiglandreas: From my tests so far it looks like the default value from the property assignement is nowhere actually added to the properties schema...

Should that not be happening? 

It didn't seem to happen in 4.14.0 either, so the change towards 4.15.0 – marking things that are not nullable as required – seems to be perfectly valid. The properties should never have been marked as nullable in the first place as they **do** contain a default value and are therefore not nullable even when no parameter is passed in. The default parameter seems to just never have been recognized - unless separately declared in the Property-definition, which should not be necessary when it is retrievable from the code.

DjordyKoert: That is also what I noticed, I have a local branch experimenting with a `ReflectionPropertyDescriber` which allows us to actually check if a default is assigned (and set it in the schema if not yet manually defined) to a property. 

This should also prevent any unwanted `required` from being set. 

I am waiting for https://github.com/nelmio/NelmioApiDocBundle/pull/2184 to actually make a somewhat better setup for this new describer.

heiglandreas: Would it be possible to separate those two things from one another? So that it would be possible to no longer have the attributes with defaults assigned marked as required? 

As that is kinda a bug that is probably currently breaking some API-Docs already I fix for that would be great.

In a project where I encountered the issue I have currently set the requirement hard to `4.14.0` as that keeps the expected behaviour, but that should probably not be the default.

DjordyKoert: It think it makes sense to revert https://github.com/nelmio/NelmioApiDocBundle/pull/2141 for now and wait for a proper solution for documenting the `required` property in schemas.

heiglandreas: You might want to have a look at the previously mentioned PR. I think I found a not too hacky solution.

At least my new tests as well as previous tests run - though with slight modifications where the previous expectations were broken in the light of the PRs scope

PR #2182, opened 2024-01-06
Check if variable array key exists before execution method

Vlado-web:


Vlado-web:
Fix 'Undefined array key "owl:maxCardinality"' bug.

DjordyKoert:
This looks like an issue with an older version of api-platform https://github.com/api-platform/core/issues/5913. An update to https://github.com/api-platform/core/releases/tag/v3.2.2 should fix this.

Vlado-web:
I am on the latest v3.2.10 and the problem persists.

OpenApiPhp/Util.php
```diff
@@ -424,7 +424,7 @@ private static function mergeFromArray(OA\AbstractAnnotation $annotation, array
             if ('$ref' === $propertyName) {
                 $propertyName = 'ref';
             }
-            if (!\in_array($propertyName, $done, true)) {
+			if (($defaults[$propertyName] ?? null) && !\in_array($propertyName, $done, true)) {
```
DjordyKoert, line 427:
```suggestion
			if (!in_array($propertyName, $done, true) && array_key_exists($propertyName, $defaults)) {
```

DjordyKoert (changes requested):

OpenApiPhp/Util.php
Vlado-web: OK, applied, but I swapped conditions due to " This condition execution costs less than the previous one.".
I don't understand what a problem is with style? Formatting seems to be fine.

DjordyKoert: I think you are using tabs instead of spaces

Vlado-web: Exactly, thanks.

DjordyKoert:
Thank you @Vlado-web 😄

Vlado-web:
@DjordyKoert Thank you too, I like the bundle, your work is appreciated.

PR #2184, opened 2024-01-09
refactor: PropertyDescribers

DjordyKoert:
Refactor to the behaviour of the property describers:
- [x] Automatic tagging for `PropertyDescriberInterface` with `nelmio_api_doc.swagger.processor`. This makes it easier for users of the bundle the create their own property describers.
- [x] Deprecate `NullablePropertyTrait` in favor of using a property describer service.
	- [x] Implement a nullable property describer as replacement
	- [x] Implement `PropertyDescriberAwareTrait` & `PropertyDescriberAwareInterface` to allow 'chaining' of describers
- [x] Refactor `PropertyDescriberInterface` dependency injection by no longer requiring to pass an array of describers
	- [x] Potential BC: `ArrayPropertyDesciber` & `CompoundPropertyDesciber` constructor type changed from `PropertyDescriberInterface[]`to `PropertyDescriberInterface`

`ObjectModelDescriber` changes:
- [x] Deprecate injection of `PropertyDescriberInterface[]` in favor of a single `PropertyDescriberInterface`
- [x] Require `symfony/deprecation-contracts` to ease deprecations

Additional change:
- [x] add `$context` (context from Model) parameter to methods of `PropertyDescriberInterface` 
	- Prevents potential recursion when chaining property describers
	- Allows better property generation using model context. Potential use cases:
		- https://github.com/nelmio/NelmioApiDocBundle/pull/2098 Better doc generation for uuid https://github.com/symfony/symfony/blob/7bbc9be7ea7d607fa901edb2068c2cea13d24978/src/Symfony/Component/Serializer/Normalizer/UidNormalizer.php#L53-L62

### DIFF
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -76,7 +76,7 @@
         <service id="nelmio_api_doc.model_describers.object" class="Nelmio\ApiDocBundle\ModelDescriber\ObjectModelDescriber" public="false">
             <argument type="service" id="property_info" />
             <argument type="service" id="annotations.reader" on-invalid="null"/>
-            <argument type="tagged" tag="nelmio_api_doc.object_model.property_describer" />
+            <argument type="service" id="nelmio_api_doc.object_model.property_describer" />
             <argument />
             <argument type="service" id="serializer.name_converter.metadata_aware" on-invalid="ignore" />
             <argument>%nelmio_api_doc.use_validation_groups%</argument>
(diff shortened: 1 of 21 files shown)


Resources/config/services.xml
```diff
@@ -94,29 +94,49 @@
         </service>
 
         <!-- Property Describers -->
+        <instanceof id="Nelmio\ApiDocBundle\PropertyDescriber\PropertyDescriberInterface">
+            <tag name="nelmio_api_doc.object_model.property_describer"/>
+        </instanceof>
+
+        <service id="Nelmio\ApiDocBundle\PropertyDescriber\PropertyDescriberInterface" alias="nelmio_api_doc.object_model.property_describer"/>
```
DjordyKoert, line 101: Remove this to prevent users from unintentionally injecting this in unrelated services. Use the `PropertyDescriberAware` instead

PR #2186, opened 2024-01-10
Fix marking falsy default values as required

### LINKED ISSUE #2179: Properties with default values are required

heiglandreas:
Properties declared like this used to not be marked as required in v 4.14.0 and then resulted in those properties being marked as required in v 4.15.0

After some testing we found out that the issue seems not to be related to whether properties are marked as nullable or not but whether default values were set for properties or not. This seemed to be unrelated to whether those default values were falsy or not.

Properties that had a default value set **in code** were not considered as having a default value set and would therefore be marked as required fields

So a property declared like this

```php
class MyClass {
    public $property = 'default';
}
```

would not have been considered as having a default value and therefore is marked as required in the API docs.

```php
class MyClass {
	#[Property(default: "default"]
	public $property = 'default';
}
```

on the other hand would have been considered to have a default value but despite that would still have been flagged as required.

So the second part was to modify the changes that were introduced in v4.15.0 to check whether a default value was assigned to the property and if so to not add that to the required fields.

Fixes #2179


heiglandreas:
The failing tests seem not to be related to the code-changes.

DjordyKoert:
@heiglandreas I updated with master & simplified the method to retrieve default values to make it work on every version instead of only php 8 or higher. 😄

DjordyKoert:
Thank you @heiglandreas !

heiglandreas:
You're welcome! Though I'd say you did the most work in cleaning up my mess 😁

Issue #2099, opened 2023-05-04
Reduce the quantity of logs generated by name collisions


emhovis:
I've just undertook the upgrade from v3 to v4 and have been overwhelmed by a nonsensical amount of logged name collisions from the logging statement added in #1862.  When an offending model is used, the name generator appends/increments an integer on the end of the model's name. Visually, seeing RedactedEntity9 in the documentation can be a bit annoying, but its still functionally fine. The problem I have with the way things currently work is that once a name collision occurs, any further usages of that model will collide with the original, as well as the first generated name, then with the second generated name etc. Along the way, every single collision gets logged. Where n is the number of times the model is used, n*((n-1) / 2) describes the amount of times the name generator will attempt to name the model and log everything about the collision. 

Did I mention that every one of those collisions and secondary collisions gets logged?

One of the offending models is referenced 258 times and therefore has 33,153 attempts at naming (and logging the attempt) for A SINGLE MODEL. Chrome crashes when attempting to view profiler logs. Out of curiosity, I downloaded firefox and attempted to view the profiler log page and firefox ate 10gb of memory! I can't even filter out the api_doc/info specific logs until they finish loading, which never happens given memory limitations.

Example of a single log from one of the hundreds of thousands of entries from a single /doc query:

nelmio_api_doc.INFO: Can not assign a name for the model, the name "RedactedEntityName257" has already been taken. {"model":{"type":{"class":"App\\Persistence\\Entity\\RedactedEntityName\\RedactedEntityName","built_in_type":"object","nullable":false,"collection":false,"collection_key_types":null,"collection_value_types":null},"options":null,"groups": [REDACTEDGROUPS]
},"taken_by":{"type":{"class":"App\\Persistence\\Entity\\RedactedEntityName\\RedactedEntityName","built_in_type":"object","nullable":false,"collection":false,"collection_key_types":null,"collection_value_types":null},"options":null,"groups":[OTHERREDACTEDGROUPS"]
}} []

Should we consider re-evaluating how the model name-check is performed or at least (for the time being) look into ways to prevent such a ridiculous amount of logging? I've tested with commenting out that single log statement- the logging itself adds 50% to the loadtime and takes nearly 2x the memory for the /doc route and makes my logfile/profiler completely unusable.

PR #2214, opened 2024-02-06
Remove model name sub-collision logging

emhovis:
With all of the love this bundle has recently been getting, I got tired of manually removing the logging functionality every update to be able to actually use my local environment's OA docs.

This should alleviate a good chunk of the problems referenced in #2099
Rather than logging ~n^2/2 average times per shared name, this change cuts it down to at most once per shared name. There's still quite a bit of logging, but this cuts down the number of logs by an order of magnitude... so I'd say its a good start. If you have 1000 models named "image", and load the entire doc reference, the current package would generate ~500k logs but with the changes in this PR would only generate 999.

I'd prefer if it were an option we could opt out of logging it in its entirety via a config var but I've spent far too long trying to makeshift that to work.

### DIFF
--- a/src/Model/ModelRegistry.php
+++ b/src/Model/ModelRegistry.php
@@ -152,7 +152,7 @@ final class ModelRegistry
 
         if ($i > 1) {
             if (isset($this->registeredModelNames[$base])) {
-                $this->logger->info(sprintf('Can not assign a name for the model, the name "%s" is now being used %s times', $base, $i), [
+                $this->logger->info(sprintf('Can not assign a name for the model, the name "%s" has already been taken.', $base), [
                     'model' => $this->modelToArray($model),
                     'taken_by' => $this->modelToArray($this->registeredModelNames[$base]),
                 ]);
--- a/src/Model/ModelRegistry.php
+++ b/src/Model/ModelRegistry.php
@@ -146,16 +146,16 @@ final class ModelRegistry
         );
         $i = 1;
         while (\in_array($name, $names, true)) {
-            if (isset($this->registeredModelNames[$name])) {
-                $this->logger->info(sprintf('Can not assign a name for the model, the name "%s" has already been taken.', $name), [
-                    'model' => $this->modelToArray($model),
-                    'taken_by' => $this->modelToArray($this->registeredModelNames[$name]),
-                ]);
-            }
-            ++$i;
+            $i++;
             $name = $base.$i;
         }
 
+        if($i > 1) {
+            $this->logger->info(sprintf('Can not assign a name for the model, the name "%s" is now being used %s times', $base, $i), [
+                'model' => $this->modelToArray($model),
+                'taken_by' => $this->modelToArray($this->registeredModelNames[$base]),
+            ]);
+        }
         return $name;
     }
 
--- a/src/Model/ModelRegistry.php
+++ b/src/Model/ModelRegistry.php
@@ -158,6 +158,7 @@ final class ModelRegistry
                 ]);
             }
         }
+
         return $name;
     }
 
--- a/src/Model/ModelRegistry.php
+++ b/src/Model/ModelRegistry.php
@@ -151,10 +151,12 @@ final class ModelRegistry
         }
 
         if ($i > 1) {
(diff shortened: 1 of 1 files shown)


emhovis:
Ideally I'd have liked to log the number of times its being reused, but the current testing is a bit shallow on capabilities. I could just hardcode the "is being used 2 times." into the test and it'd pass phpunit but that is a bit misleading since we're only testing 1 collision and no sub-collisions... Then again, current testing doesn't take subcollisions into account so I guess both ways are a tad misleading...

Sorry for the mention, but as the person leading the recent surge of maintenance, I figured your input would be valuable- Do you have any thoughts @DjordyKoert

DjordyKoert:
> Ideally I'd have liked to log the number of times its being reused, but the current testing is a bit shallow on capabilities. I could just hardcode the "is being used 2 times." into the test and it'd pass phpunit but that is a bit misleading since we're only testing a collision depth of 1... Then again, current testing doesn't take secondary collisions into account so I guess both ways are a tad misleading...

I don't think it hurts to rewrite the current test to ensure this doesn't happen again. Maybe using a dataProvider

DjordyKoert:
> I'd prefer if it were an option we could opt out of logging it in its entirety via a config var but I've spent far too long trying to makeshift that to work.

Since https://github.com/nelmio/NelmioApiDocBundle/pull/2131 it's possible to pass a `NullLogger`. to the `ApiDocGenerator` (or manually call `setLogger`). Would that not be a solution for you?

emhovis:
> > I'd prefer if it were an option we could opt out of logging it in its entirety via a config var but I've spent far too long trying to makeshift that to work.
> 
> Since #2131 it's possible to pass a `NullLogger`. to the `ApiDocGenerator` (or manually call `setLogger`). Would that not be a solution for you?

That actually mostly fixes my issue. Can't believe I missed that getting added!
On a side note, I still think it'd be ideal to pass this PR. I don't see the need for logging an entire loop of previous collision instances for each new collision as default behavior.

emhovis:
> > > I'd prefer if it were an option we could opt out of logging it in its entirety via a config var but I've spent far too long trying to makeshift that to work.
> > 
> > 
> > Since #2131 it's possible to pass a `NullLogger`. to the `ApiDocGenerator` (or manually call `setLogger`). Would that not be a solution for you?
> 
> That actually mostly fixes my issue. Can't believe I missed that getting added! On a side note, I still think it'd be ideal to pass this PR. I still don't see the need for logging an entire loop of previous collision instances for each new collision as default behavior.

I stand corrected- it appears that the Symfony Profiler captures all logged items regardless of whether a nullhandler is used to "blackhole" them or not. My Symfony Profiler still OOMs the browser when trying to view the logs for the /doc page.

As a bit of an experiment, I set monolog to separate the Nelmio logs into their own file- it generates a ~290mb logfile of name collisions for 1 /doc request.

With the changes in this PR, the logfile generated for the /doc page is only ~4mb and my browser doesn't crash. I'd call that a good start.

DjordyKoert:
I am happy to merge this if you could write some tests for this (probably by rewriting the current test).

Model/ModelRegistry.php
```diff
@@ -127,14 +127,17 @@ private function generateModelName(Model $model): string
...
         while (\in_array($name, $names, true)) {
-            if (isset($this->registeredModelNames[$name])) {
-                $this->logger->info(sprintf('Can not assign a name for the model, the name "%s" has already been taken.', $name), [
+            ++$i;
+            $name = $base.$i;
+        }
+
+        if ($i > 1) {
+            if (isset($this->registeredModelNames[$base])) {
+                $this->logger->info(sprintf('Can not assign a name for the model, the name "%s" has already been taken.', $base), [
```
DjordyKoert, line 136:
```suggestion
                $this->logger->info(sprintf('Can not assign a name for the model, the name "%s" has been taken %d times.', $base, $i), [
```

DjordyKoert (changes requested):

shakaran:
@emhovis could you rebase with last updates? Thanks

emhovis:
> @emhovis could you rebase with last updates? Thanks
Done. 

Things are slowing down a bit for me, I'll see if I can set some time aside soon for writing those tests so we can finally get this merged

PR #2225, opened 2024-02-21
Allow `cache_pool` option in configuration

Brajk19:
ApiDocGenerator has CachePoolInterface as property but it is always null and can not be set to anything else.

This PR enables setting cache_pool so if its defined ApiDocGenerator will use it.
Value under cache_pool should be service ID

Example:
in my services.yaml i will have something like this
```yaml
services:
    app.cache:
        class: Symfony\Component\Cache\Adapter\FilesystemAdapter
```

and in nelmio_api_doc.yaml:
```yaml
nelmio_api_doc:
    cache:
        pool: app.cache
        item_id: docs
    ...
```


DependencyInjection/NelmioApiDocExtension.php
```diff
@@ -96,7 +98,7 @@ public function load(array $configs, ContainerBuilder $container): void
                 ->setArguments([
                     new TaggedIteratorArgument(sprintf('nelmio_api_doc.describer.%s', $area)),
                     new TaggedIteratorArgument('nelmio_api_doc.model_describer'),
-                    null,
+                    $cachePool,
                     null,
```
DjordyKoert, line 102: I think it would also make sense to allow configuring the `$cacheItemId` with this PR. 

https://github.com/nelmio/NelmioApiDocBundle/blob/61c08f139e104db21b12832173b08fa37c22e619/ApiDocGenerator.php#L68

Would you be able to include this in this PR? 😄

Brajk19: @DjordyKoert yes, of course
take a look 😄 
https://github.com/nelmio/NelmioApiDocBundle/pull/2225/commits/b30390eba7c8fec268f41ed889f10b0e4317143d

DjordyKoert: Yeah that would make sense to me

Brajk19: done!
https://github.com/nelmio/NelmioApiDocBundle/pull/2225/commits/c2d076b216d6af8ec3c7bf089ca5af3d2085d665

DjordyKoert: Thank you! 😄

DjordyKoert:
Thank you @Brajk19! ❤️

PR #2228, opened 2024-03-02
Update bundle directory structure

DjordyKoert:
Update directory structure to the new recommended structure https://symfony.com/doc/current/bundles/best_practices.html#directory-structure


j0k3r:
Isn't a BC?

We defined a route like:

```yml
doc-api:
    resource: "@NelmioApiDocBundle/config/routing/swaggerui.xml"
    prefix: /api/doc
```

But upgrading to 4.22.0 broke the build because that file no longer exist.
Maybe a better note in the release note about that would have been great. This isn't only related to assets.

DjordyKoert:
@j0k3r My apologies, this is something that has not crossed my mind at the time. I will update the release notes about this

PR #2232, opened 2024-03-04
Fix for using cache pools with multiple areas

Brajk19:
https://github.com/nelmio/NelmioApiDocBundle/pull/2225 enabled usage of cache pools.
I noticed a problem when using this with multiple areas.

If i have configuration with `area1` and `area2`, depending on which ApiDocGenerator is called first, its specification will be cached and all following generators (for any area) will return that cached response because same item id is used.

Defining global cache (for all areas) is deprecated and all cache config should be area specific.

```yaml
nelmio_api_doc:
    areas:
        default:
            cache:
                pool: app.cache
                item_id: docs
        area2:
            cache:
                pool: app.cache
                item_id: docs2
    ...
```


src/DependencyInjection/NelmioApiDocExtension.php
```diff
@@ -100,7 +100,7 @@ public function load(array $configs, ContainerBuilder $container): void
                     new TaggedIteratorArgument(sprintf('nelmio_api_doc.describer.%s', $area)),
                     new TaggedIteratorArgument('nelmio_api_doc.model_describer'),
                     $cachePool,
-                    $cacheItemId,
+                    $cacheItemId.'.'.$area,
```
Brajk19, line 103: This would set cache key even if cache pool is null. Maybe it would be better to only set it if cache pool is set?

DjordyKoert:
> I'm not sure what would be best fix for this. I've thought about moving cache.pool and cache.item_id config under areas but that would break BC.

Let's deprecate the ` cache.pool` and `cache.item_id` and move the cache config to areas.

> Other idea is that area is automatically appended to cache key so if we have config like this:

I don't think we should do this.

Brajk19:
> > I'm not sure what would be best fix for this. I've thought about moving cache.pool and cache.item_id config under areas but that would break BC.
> 
> Let's deprecate the ` cache.pool` and `cache.item_id` and move the cache config to areas.
> 
> > Other idea is that area is automatically appended to cache key so if we have config like this:
> 
> I don't think we should do this.

Alright. I'll move it to areas 👍

src/DependencyInjection/Configuration.php
```diff
@@ -122,6 +128,22 @@ public function getConfigTreeBuilder(): TreeBuilder
                                 ->example(['info' => ['title' => 'My App']])
                                 ->prototype('variable')->end()
                             ->end()
+                            ->arrayNode('cache')
+                                ->validate()
+                                    ->ifTrue(function ($v) { return null !== $v['pool'] && null === $v['item_id']; })
+                                    ->thenInvalid('Must set cache.item_id if cache.pool is set')
```
Brajk19, line 134: @DjordyKoert i have changed this validation so that item_id must be specified...
this will prevent accidents if someone does not specify cache key so potentially wrong specification will be returned.
is that ok?

DjordyKoert: a different pool could be used per area, if that happens then it doesn't necessarily need to be specified

DjordyKoert:
> > I'm not sure what would be best fix for this. I've thought about moving cache.pool and cache.item_id config under areas but that would break BC.
> 
> Let's deprecate the ` cache.pool` and `cache.item_id` and move the cache config to areas.

Looking at the per area implementation that you made I think it won't be necessary to deprecate the old cache configuration.  
It doesn't hurt to automatically append the area to the `cache.item_id` if this happens

> > Other idea is that area is automatically appended to cache key so if we have config like this:
> I don't think we should do this.

I also fully take this back (my bad 😅) the old behaviour (before the cache option was implemented https://github.com/nelmio/NelmioApiDocBundle/pull/2225) was already broken if someone if someone did the following:
- Manually define the cache args for the ApiDocGenerator service(s)
- Forget to set an arg for the `cacheItemId`
- Use multiple areas in the same cache pool

Doing above these steps would also result in the cache to be shared between areas.

src/DependencyInjection/Configuration.php
Brajk19: true, that makes sense. i'll remove it

Brajk19:
How about this:
 - we leave old configuration as is, do not deprecate it
 - if cache is defined globally it will be used for every area
 - if it's also defined for area, that that config will be used
 
 so basically global config is used as fallback if area-specific config is not defined
 
 i could have something like this
 ```yaml
nelmio_api_doc:
    cache:
        pool: app.cache
    areas:
        default:
            item_id: docs
        area2:
            item_id: docs2
    ...
```
same pool, different item ids

**Comment by Brajk19** [comment]
i'm just not sure how to fit auto appending area to cache key into this logic
should we event do that?

**Comment by Brajk19** [comment]
> > i'm just not sure how to fit auto appending area to cache key into this logic should we even do that?
> 
> I think we should. I can also create something for this really quick

https://github.com/nelmio/NelmioApiDocBundle/pull/2232/commits/c8e7c8fe191ff8010863faa022d67781b94d9e95
i added change which removes deprecation and uses global config as fallback if area-specific config is not defined
feel free to add that auto appending of area how you envisioned it 😄

File: src/DependencyInjection/NelmioApiDocExtension.php [hunk 3]
```diff
@@ -88,6 +88,9 @@ public function load(array $configs, ContainerBuilder $container): void
         $cacheItemId = $config['cache']['item_id'] ?? null;
 
         foreach ($config['areas'] as $area => $areaConfig) {
+            $areaCachePool = $areaConfig['cache']['pool'] ?? $cachePool;
```
HypeMC, line 91: @Brajk19 Shouldn't `$areaConfig['cache']['pool']`be wrapped in `new Reference()`?

DjordyKoert: Fixed it in [38711df](https://github.com/nelmio/NelmioApiDocBundle/pull/2232/commits/38711df494bf497bc40303be56def281f931bc27), thank you 😄

Brajk19: My bad, thanks 😄

DjordyKoert (approved):

DjordyKoert:
Thank you @Brajk19!

PR #2234, opened 2024-03-05
Update faq.rst; wrong nesting of disable_default_routes option

texnicii:
The option disable_default_routes located into areas (not into nested default option)
image

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -205,8 +205,7 @@ A: Use ``disable_default_routes`` config in your area.
 
     nelmio_api_doc:
         areas:
-            default:
-                disable_default_routes: true
+            disable_default_routes: true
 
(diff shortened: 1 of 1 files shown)


DjordyKoert:
The docs are correct here. It looks like you are incorrectly using the areas configuration.

From the exception I assume that your configuration looks like this:
```yaml
nelmio_api_docs:
  areas:
    default:
      default:
...
```

This should be:
```yaml
nelmio_api_docs:
  areas:
    default:
...
```

texnicii:
My configuration is and it is working fine
```yaml
nelmio_api_doc:
    areas:
        disable_default_routes: true
```

but config says and that config throws this exception
```yaml
nelmio_api_doc:
    areas:
        default:
            disable_default_routes: true
```

Issue #2212, opened 2024-02-06
[Symfony 7.1] support for TypeInfo component


DjordyKoert:
https://github.com/symfony/symfony/pull/52510

Symfony 7.1 will introduce a new `TypeInfo` component in favor of the current `PropertyInfo` component. This component includes much better type information compared to the current `PropertyInfo` component so Nelmio should support this.

PR #2236, opened 2024-03-06
Support for range integers + drop support for PHP 7.2 & 7.3

bnowak:
It adds support for some basic https://phpstan.org/writing-php-code/phpdoc-types#integer-ranges.


bnowak:
@DjordyKoert That feature would work on any versions except the lowest ones (the single CI job with `--prefer-lowest` flag). I've checked that it works from the very lowest dependencies (comparing with the lowest installed by default):
```
SYMFONY_REQUIRE=5.4.* composer update \
phpdocumentor/reflection-docblock=4.3.4 \
symfony/property-info=5.4.10 \
phpstan/phpdoc-parser=1.13.0 \ # dependency of phpdocumentor/type-resolver=1.8.2 so not needed to specify in composer
phpdocumentor/type-resolver=1.8.2
```

 but `phpdocumentor/type-resolver=1.8.2` required PHP 7.3 at least so my question is: could we bump dependencies and php to that version as minimum?

I know bumping PHP to next major version could be to too radical, so my alternative would be bumping only `phpdocumentor/reflection-docblock` & `symfony/property-info` versions and make the feature (tests actually) PHP & phpdocumentor/type-resolver versions specific (that it will work from these the lowest versions).

Would do you think?

DjordyKoert:
> Would you mean just only bumping it up in composer or also any other related adjustments in code (like getting rid of some backward-compatible workarounds etc.)?

I don't think we have any real backward-compatible workarounds, only in a few tests I think it looks like. 
- https://github.com/search?q=repo%3Anelmio%2FNelmioApiDocBundle%207.2&type=code
- https://github.com/search?q=repo%3Anelmio%2FNelmioApiDocBundle%207.3&type=code

Feel free to remove this, otherwise I will remove them myself 😃

bnowak:
:+1: thanks for replying

I'm going to handle that tomorrow :wink:

DjordyKoert:
> However, I noticed failed pipelines (on SF >= 6.3) which are not related to my changes. It seems that the change which broke them was introduced in [nelmio/NelmioApiDocBundle/actions/runs/8231364006/job/22506509816](https://github.com/nelmio/NelmioApiDocBundle/actions/runs/8231364006/job/22506509816) but I'm not sure how it should be fixed.

I already have an issue open at swagger-php about this :) https://github.com/zircote/swagger-php/issues/1556

DjordyKoert:
> @DjordyKoert in general PR is ready for review. Finally, I decided to bump only to PHP 7.3 (plus needed composer dependencies) which is enough for this feature.

LGTM, I would still like to drop support for PHP 7.3 in this and make PHP 7.4 the minimum supported version would you like & be able to do this? :)
 
Additionally an update inside the changelog.md would be appreciated.

bnowak:
:+1: will handle that

bnowak:
@DjordyKoert adjusted to PHP 7.4 as minimum version, but I'm struggling with 2 errors in `BazingaFunctionalTest` (https://github.com/nelmio/NelmioApiDocBundle/actions/runs/8279847555/job/22655096369?pr=2236) which I cannot reproduce locally in docker (using PHP 7.4 image). I double checked that have all composer installed dependencies exactly the same as in CI job, but different results.

Do you have any idea what else I could check?

DjordyKoert:
> Do you have any idea what else I could check?

It looks like there is a caching issue. On my local machine the first test (`testModelComplexDocumentationBazinga`) passes whenever no cache is created yet. The tests after fail (because they use the cache created in `testModelComplexDocumentationBazinga`).

Rerunning the full `BazingaFunctionalTest` without deleting that '/var` directory (aka without removing cache) also causes `testModelComplexDocumentationBazinga` to fail.....

DjordyKoert:
@bnowak [514239c](https://github.com/nelmio/NelmioApiDocBundle/pull/2236/commits/514239ce9945495b38aec09d8fd41f90a61e1abd) fixed it 🥳

bnowak:
cool, thanks for help with that :)

CHANGELOG.md
```diff
@@ -1,6 +1,24 @@
...
+     * @var int<min, 11>
+     * @var int<6, max>
+     * @var positive-int
+     * @var negative-int
+     */
+  ```
+  will be interpreted as appropriate `minimum` and `maximum` properties in the generated OpenAPI specification.
+
+### Breaking change
+Dropped support for PHP 7.2 and PHP 7.3, so 7.4 is the minimum required version now.
```
DjordyKoert, line 20:
```suggestion
Dropped support for PHP 7.2 and PHP 7.3. PHP 7.4 is the minimum required version now.
```

CHANGELOG.md
```diff
@@ -1,6 +1,24 @@
 CHANGELOG
 =========
 
+4.24.0
+-----
+* Added support for some integer ranges (https://phpstan.org/writing-php-code/phpdoc-types#integer-ranges).
```
DjordyKoert, line 6:
```suggestion
* Added support for integer ranges (https://phpstan.org/writing-php-code/phpdoc-types#integer-ranges).  
```

Is it possible to support all of these documented types?

- [x] positive-int
- [x] negative-int
- [ ] non-positive-int
- [ ] non-negative-int
- [ ] non-zero-int
- [x] int<0, 100>
- [x] int<min, 100>
- [x] int<50, max>

tests/Functional/FunctionalTest.php
```diff
@@ -1359,4 +1228,50 @@ public function testEntityWithFalsyDefaults()
...
+                    'type' => 'integer',
+                    'nullable' => true,
+                    'minimum' => 1,
+                    'maximum' => 99,
+                ],
+            ],
+            'type' => 'object',
+        ];
+
+        if (version_compare(Kernel::VERSION, '6.1', '>=')) {
```
DjordyKoert, line 1261: Should this not be available on every version now that both `phpdocumentor/reflection-docblock` & `phpdocumentor/type-resolver` received a version bump?

DjordyKoert (changes requested):

CHANGELOG.md
bnowak: `phpdocumentor/type-resolver` doesn't support the rest (`non-positive-int`, `non-negative-int`, `non-zero-int`) as pseudo types ([here are the only supported types](https://github.com/phpDocumentor/TypeResolver/tree/1.x/src/PseudoTypes)).

I could parse and handle them more directly, however there is another blocker in `symfony/property-info` ([here](https://github.com/symfony/property-info/blob/7.0/Util/PhpStanTypeHelper.php#L173-L197) even in the newest version) which doesn't support them as well. Because of that we're getting errors like:
```
1) Nelmio\ApiDocBundle\Tests\Functional\FunctionalTest::testRangeIntegers
LogicException: Schema of type "\Nelmio\ApiDocBundle\Tests\Functional\Entity\non-positive-int" can't be generated, no describer supports it. Class "\Nelmio\ApiDocBundle\Tests\Functional\Entity\non-positive-int" does not exist, did you forget a use statement, or typed it wrong?

/app/src/Model/ModelRegistry.php:107
/app/src/ApiDocGenerator.php:129
/app/tests/Functional/WebTestCase.php:28
/app/tests/Functional/WebTestCase.php:41
/app/tests/Functional/FunctionalTest.php:1275
```

As Nelmio bundle is highly coupled and based on property info component, to make it work - it would probable a lot of further refactors and changes.

tests/Functional/FunctionalTest.php
bnowak: This is not related to any `phpdocumentor` lib, but to `symfony/property-info` supports which I described in https://github.com/nelmio/NelmioApiDocBundle/pull/2236#discussion_r1525012814.

`positive-int` and `negative-int` pseudo types are supported there, but only from 6.1 version ([here](https://github.com/symfony/property-info/blob/v6.1.0/Util/PhpStanTypeHelper.php)). Without this condition, tests would throw similar error like mentioned in https://github.com/nelmio/NelmioApiDocBundle/pull/2236#discussion_r1525012814.

CHANGELOG.md
DjordyKoert: Ah okay, too bad then! Maybe the new TypeInfo component is able to do this in the future https://github.com/nelmio/NelmioApiDocBundle/issues/2212

DjordyKoert (approved):

bnowak:
@DjordyKoert thanks for approving :heart: 

will https://github.com/nelmio/NelmioApiDocBundle/pull/2236#discussion_r1525012814 be a blocker for merging it or not?

DjordyKoert:
> will [#2236 (comment)](https://github.com/nelmio/NelmioApiDocBundle/pull/2236#discussion_r1525012814) be a blocker for merging it or not?

Nope! 😄

bnowak:
sorry I meant swagger-php issue (comment updated) :sweat_smile:

DjordyKoert:
> sorry, I meant swagger-php issue (comment updated) 😅

Still a nope hahaha. That is an issue on its own.

DjordyKoert:
Thanks a lot @bnowak for this new feature + the other things that you were willing to pick up ❤️

PR #2241, opened 2024-03-08
Update .symfony.bundle.yaml

javiereguiluz:
We saw some errors in the logs of the Symfony server related to the building of the docs of this bundle for the 3.x branch. We need to update the Symfony bundle config because the doc location is different in 3.x and master.

Thanks!


DjordyKoert:
Thank you! @javiereguiluz

PR #2248, opened 2024-03-29
fix(#2222): Fix properties with default values getting marked as required

### LINKED ISSUE #2222: Properties with default values still get marked as required
Since v4.15.0, we've encountered a problem regarding non-required properties
A property gets initialized in the constructor with a default value:
```php
    public function __construct(
        #[Groups('docs')]
        #[Valid]
        private readonly CompanyQueryFilters $filters = new CompanyQueryFilters()
    ) {
    }
```
```php
final readonly class CompanyQueryFilters {
    public function __construct(
        #[Groups('docs')]
        private ?string $organisation = null
    ) {
    }

    public function getOrganisation(): ?string {
        return $this->organisation;
    }
}
```

The schema output is as follows:
```json
    "GetAllCompaniesQuery": {
        "required": [
            "filters"
        ],
        "properties": {
            "filters": {
                "$ref": "#/components/schemas/CompanyQueryFilters"
            }
        },
        "type": "object"
    },
```

We expect the `filters`-property to be non-required.

Follow-up on #2185 and #2179

DominicLuidold:
| Issues        | Fix #2222  |

This PR fixes the following two bugs:
* Promoted properties with a value initialized by the constructor are not considered to have a default value and are therefore not returned by `ReflectionClass::getDefaultProperties()`.[^1]
* The `SymfonyConstraintAnnotationReader::processPropertyAnnotations()` method is called before a property's default value is set and additionally doesn't consider a possible set default value.

[^1]: https://bugs.php.net/bug.php?id=81386

### DIFF
--- a/src/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
+++ b/src/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
@@ -84,7 +84,7 @@ class SymfonyConstraintAnnotationReader
                     return;
                 }
 
-                if (true === $property->nullable || !Generator::isDefault($property->default)) {
+                if (!Generator::isDefault($property->default)) {
                     return;
                 }
 
--- a/src/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
+++ b/src/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
@@ -84,6 +84,10 @@ class SymfonyConstraintAnnotationReader
                     return;
                 }
 
+                if (true === $property->nullable || !Generator::isDefault($property->default)) {
+                    return;
+                }
+
                 $existingRequiredFields = Generator::UNDEFINED !== $this->schema->required ? $this->schema->required : [];
                 $existingRequiredFields[] = $propertyName;
 
--- a/src/ModelDescriber/ObjectModelDescriber.php
+++ b/src/ModelDescriber/ObjectModelDescriber.php
@@ -136,7 +136,7 @@ class ObjectModelDescriber implements ModelDescriberInterface, ModelRegistryAwar
         $reflClassConstructor = $reflClass->getConstructor();
         $reflClassConstructorParameters = null !== $reflClassConstructor ? $reflClassConstructor->getParameters() : [];
         foreach ($reflClassConstructorParameters as $parameter) {
-            if (!$parameter->isDefaultValueAvailable() || null === $parameter->getDefaultValue()) {
+            if (!$parameter->isDefaultValueAvailable()) {
                 continue;
             }
 
--- a/src/ModelDescriber/ObjectModelDescriber.php
+++ b/src/ModelDescriber/ObjectModelDescriber.php
@@ -130,6 +130,19 @@ class ObjectModelDescriber implements ModelDescriberInterface, ModelRegistryAwar
             return null !== $value;
         });
 
+        // Fix for https://github.com/nelmio/NelmioApiDocBundle/issues/2222
+        // Promoted properties with a value initialized by the constructor are not considered to have a default value
+        // and are therefore not returned by ReflectionClass::getDefaultProperties(); see https://bugs.php.net/bug.php?id=81386
+        $reflClassConstructor = $reflClass->getConstructor();
+        $reflClassConstructorParameters = null !== $reflClassConstructor ? $reflClassConstructor->getParameters() : [];
+        foreach ($reflClassConstructorParameters as $parameter) {
+            if (!$parameter->isDefaultValueAvailable() || null === $parameter->getDefaultValue()) {
+                continue;
+            }
+
+            $defaultValues[$parameter->name] = $parameter->getDefaultValue();
+        }
+
         foreach ($propertyInfoProperties as $propertyName) {
             $serializedName = null !== $this->nameConverter ? $this->nameConverter->normalize($propertyName, $class, null, $model->getSerializationContext()) : $propertyName;
 
@@ -141,6 +154,13 @@ class ObjectModelDescriber implements ModelDescriberInterface, ModelRegistryAwar
             }
 
             $property = Util::getProperty($schema, $serializedName);
+
+            // Fix for https://github.com/nelmio/NelmioApiDocBundle/issues/2222
+            // Property default value has to be set before SymfonyConstraintAnnotationReader::processPropertyAnnotations()
+            // is called to prevent wrongly detected required properties
+            if (Generator::UNDEFINED === $property->default && array_key_exists($propertyName, $defaultValues)) {
+                $property->default = $defaultValues[$propertyName];
+            }
 
             // Interpret additional options
             $groups = $model->getGroups();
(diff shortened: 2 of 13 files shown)


DominicLuidold:
@DjordyKoert tests are failing with PHP 7.4 because of the promoted constructor properties syntax in the two newly added test entities. Do you have an approach in mind how to prevent those failures?

I've thought of wrapping the entity classes and methods within the `ApiController`s with `if (PHP_VERSION_ID < 80000) {` but I'm not sure if that's the correct approach here.. :thinking:

src/ModelDescriber/ObjectModelDescriber.php
```diff
@@ -130,6 +130,19 @@ public function describe(Model $model, OA\Schema $schema)
             return null !== $value;
         });
 
+        // Fix for https://github.com/nelmio/NelmioApiDocBundle/issues/2222
+        // Promoted properties with a value initialized by the constructor are not considered to have a default value
+        // and are therefore not returned by ReflectionClass::getDefaultProperties(); see https://bugs.php.net/bug.php?id=81386
+        $reflClassConstructor = $reflClass->getConstructor();
+        $reflClassConstructorParameters = null !== $reflClassConstructor ? $reflClassConstructor->getParameters() : [];
+        foreach ($reflClassConstructorParameters as $parameter) {
+            if (!$parameter->isDefaultValueAvailable() || null === $parameter->getDefaultValue()) {
```
DjordyKoert, line 139: `null` could be a valid default value, this check seems unnecessary.

```suggestion
            if (!$parameter->isDefaultValueAvailable()) {
```

src/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
```diff
@@ -84,6 +84,10 @@ private function processPropertyAnnotations($reflection, OA\Property $property,
                     return;
                 }
 
+                if (true === $property->nullable || !Generator::isDefault($property->default)) {
```
DjordyKoert, line 87: It does not seem to make sense to check `true === $property->nullable` on an annotation of type `Assert\NotNull`

tests/Functional/Entity/EntityWithPromotedPropertiesWithDefaults81.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    #[Assert\NotBlank]
+    public readonly string $order;
+
+    public function __construct(
+        string $order = 'asc',
+
+        public readonly int $page = 30,
+        public readonly string $sort = 'id',
+    ) {
```
DjordyKoert, line 26: Preferably add some more parameters options here for tests:
- `string $...,` - No default value specified for a non-promoted property
- `?string $...,` - No default value specified for a nullable non-promoted property
- `?string $... = null,` - `null` default value specified for a nullable non-promoted property
- `public ?string $... = null,` - `null` default value specified for a nullable promoted property

tests/Functional/Controller/ApiController80.php
```diff
@@ -497,7 +498,7 @@ public function entityWithNullableSchemaSet()
     }
 
     /**
-     * @Route("/entity-with-falsy-defaults", methods={"POST"})
+     * @Route("/entity-with-falsy-defaults", methods={"GET"})
```
DjordyKoert, line 501: This change seems accidental and unrelated

```suggestion
     * @Route("/entity-with-falsy-defaults", methods={"POST"})
```

DjordyKoert (changes requested):

DjordyKoert:
> @DjordyKoert tests are failing with PHP 7.4 because of the promoted constructor properties syntax in the two newly added test entities. Do you have an approach in mind how to prevent those failures?
> 
> I've thought of wrapping the entity classes and the methods within the `ApiController`s with `if (PHP_VERSION_ID >= 80000) {` but I'm not sure if that's the correct approach here.. 🤔

I think it's okay to simply wrap the entity class (`EntityWithPromotedPropertiesWithDefaults80`) with an if-else statement and create an empty class for PHP 7.4, the class should and cannot be tested on 7.4 anyway

tests/Functional/Controller/ApiController80.php
DominicLuidold: Good catch! :sweat_smile:

DominicLuidold:
@DjordyKoert Thanks for your fast response and the code review/input! I've addressed all the open points and was also able to get the tests running with PHP 7.4. :blush:

Unfortunately, the test now fails on PHP 8.0 only and I really can't get my head around why. [According to the job log](https://github.com/nelmio/NelmioApiDocBundle/actions/runs/8490044202/job/23260744858), the `EntityWithPromotedPropertiesWithDefaults80` is declared more than once:
```
PHP Fatal error:  Cannot declare class EntityWithPromotedPropertiesWithDefaults80, because the name is already in use in /home/runner/work/NelmioApiDocBundle/NelmioApiDocBundle/tests/Functional/Entity/EntityWithPromotedPropertiesWithDefaults80.php(17) : eval()'d code on line 4
```

But this isn't the case. I also manually checked the code beforehand and as far as I'm concerned, it should work. See https://3v4l.org/d0p5L#v8.0.30 with no error output. Am I missing something obvious here? :thinking: :sweat_smile:

DjordyKoert:
> @DjordyKoert Thanks for your fast response and the code review/input! I've addressed all the open points and was also able to get the tests running with PHP 7.4. 😊

Thank you!

> Unfortunately, the test now fails on PHP 8.0 only and I really can't get my head around why. [According to the job log](https://github.com/nelmio/NelmioApiDocBundle/actions/runs/8490044202/job/23260744858), the `EntityWithPromotedPropertiesWithDefaults80` is declared more than once:
> 
> ```
> PHP Fatal error:  Cannot declare class EntityWithPromotedPropertiesWithDefaults80, because the name is already in use in /home/runner/work/NelmioApiDocBundle/NelmioApiDocBundle/tests/Functional/Entity/EntityWithPromotedPropertiesWithDefaults80.php(17) : eval()'d code on line 4
> ```
> 
> But this isn't the case. I've also manually checked the code beforehand and as far as I'm concerned, it should work. See [3v4l.org/d0p5L#v8.0.30](https://3v4l.org/d0p5L#v8.0.30) with no error output. Am I missing something obvious here? 🤔 😅

I got no idea why your setup wouldn't work but I have moved this this to its own controller instead of building more onto the `ApiController..`

DjordyKoert (approved):
@DominicLuidold could you see if I missed anything by any chance? 😄

tests/Functional/Controller/PromotedPropertiesController80.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
```
DominicLuidold, line 2: The license header is missing here

tests/Functional/Controller/PromotedPropertiesController81.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
```
DominicLuidold, line 2: Here as well :slightly_smiling_face:

DominicLuidold (commented):
Just two nit-picky things. Thanks a lot for your support @DjordyKoert! :pray:

tests/Functional/Controller/PromotedPropertiesController80.php
DjordyKoert: Thanks!

DjordyKoert:
Thank you very much @DominicLuidold!

PR #2249, opened 2024-03-30
phpstan installation

DjordyKoert:
| Issues        |  |

Adds PHPStan + fixes for associated PHPStan level

- [x] Level 0
- [x] Level 1
- [x] Level 2
- [x] Level 3
- [x] Level 4
- [x] Level 5

**Will be done in a separate PR because of the amount of changes**
- [ ] ~~Level 6~~
- [ ] ~~Level 7~~
- [ ] ~~Level 8~~
- [ ] ~~Level 9~~

**Extra's**
-  Cleaned up `ControllerReflector` class
    - Removed code related to (unsupported) Symfony < 4
    - Early returns
    - Return types for methods
    - Removed usage of unreachable `ClassUtils`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -62,8 +62,7 @@
 
         "phpstan/phpstan": "^1.10",
         "phpstan/phpstan-symfony": "^1.3",
-        "phpstan/phpstan-phpunit": "^1.3",
-        "phpstan/extension-installer": "^1.3"
+        "phpstan/phpstan-phpunit": "^1.3"
     },
     "conflict": {
         "zircote/swagger-php": "4.8.7"
--- a/composer.json
+++ b/composer.json
@@ -110,10 +110,5 @@
         "phpunit": "phpunit",
         "phpstan": "phpstan --memory-limit=2G",
         "phpstan-baseline": "phpstan --memory-limit=2G --generate-baseline"
-    },
-    "config": {
-        "allow-plugins": {
-            "phpstan/extension-installer": true
-        }
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -101,7 +101,8 @@
         "phpcs-check": "Run PHP-CS-Fixer in dry-run mode",
         "phpcs-fix": "Run PHP-CS-Fixer",
         "phpunit": "Run phpunit",
-        "phpstan": "Run phpstan"
+        "phpstan": "Run phpstan",
+        "phpstan-baseline": "Generate a baseline for phpstan"
     },
     "scripts": {
         "phpcs-check": "vendor/bin/php-cs-fixer check -v --diff",
--- a/composer.json
+++ b/composer.json
@@ -108,7 +108,8 @@
         "phpcs-check": "vendor/bin/php-cs-fixer check -v --diff",
         "phpcs-fix": "vendor/bin/php-cs-fixer fix -v",
         "phpunit": "phpunit",
-        "phpstan": "phpstan --memory-limit=2G"
+        "phpstan": "phpstan --memory-limit=2G",
+        "phpstan-baseline": "phpstan --memory-limit=2G --generate-baseline"
     },
     "config": {
         "allow-plugins": {
--- a/composer.json
+++ b/composer.json
@@ -62,7 +62,8 @@
 
         "phpstan/phpstan": "^1.10",
         "phpstan/phpstan-symfony": "^1.3",
-        "phpstan/phpstan-phpunit": "^1.3"
+        "phpstan/phpstan-phpunit": "^1.3",
+        "phpstan/phpstan-strict-rules": "^1.5"
     },
     "conflict": {
         "zircote/swagger-php": "4.8.7"
--- a/composer.json
+++ b/composer.json
@@ -55,7 +55,13 @@
         "composer/package-versions-deprecated": "1.11.99.1",
 
         "phpunit/phpunit": "^8.5|^9.6",
-        "doctrine/annotations": "^2.0"
+
+        "doctrine/annotations": "^2.0",
+
+        "phpstan/phpstan": "^1.10",
+        "phpstan/phpstan-symfony": "^1.3",
+        "phpstan/phpstan-phpunit": "^1.3",
+        "phpstan/extension-installer": "^1.3"
     },
     "conflict": {
         "zircote/swagger-php": "4.8.7"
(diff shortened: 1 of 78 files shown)


composer.json
```diff
@@ -55,7 +55,13 @@
...
 
         "phpunit/phpunit": "^8.5|^9.6",
-        "doctrine/annotations": "^2.0"
+
+        "doctrine/annotations": "^2.0",
+
+        "phpstan/phpstan": "^1.10",
+        "phpstan/phpstan-symfony": "^1.3",
+        "phpstan/phpstan-phpunit": "^1.3",
+        "phpstan/extension-installer": "^1.3"
```
DominicLuidold, line 64: The extension installer is a really handy tool, but I would consider not adding a dependency if it's not really necessary - mainly to minimize the impact of bugs or unintended side effects, even if it's "just" a dev dependency.

To achieve the same result, only five lines would've to be added to the config file:
```yaml
includes:
    - vendor/phpstan/phpstan-phpunit/extension.neon
    - vendor/phpstan/phpstan-phpunit/rules.neon
    - vendor/phpstan/phpstan-symfony/extension.neon
    - vendor/phpstan/phpstan-symfony/rules.neon
```

composer.json
```diff
@@ -55,7 +55,13 @@
         "composer/package-versions-deprecated": "1.11.99.1",
 
         "phpunit/phpunit": "^8.5|^9.6",
-        "doctrine/annotations": "^2.0"
+
+        "doctrine/annotations": "^2.0",
+
+        "phpstan/phpstan": "^1.10",
```
DominicLuidold, line 61: [`phpstan/phpstan-strict-rules`](https://github.com/phpstan/phpstan-strict-rules) could be a useful extension as well

phpstan.neon
```diff
@@ -0,0 +1,19 @@
+includes:
+    - phar://phpstan.phar/conf/bleedingEdge.neon
+
+parameters:
```
DominicLuidold, line 4: If I'm not mistaken, the `symfony.containerXmlPath` parameter needs to be configured to make full use of the extension (see https://github.com/phpstan/phpstan-symfony?tab=readme-ov-file#configuration)

DominicLuidold (changes requested):
@DjordyKoert I hope you don't mind some input even though your PR is still marked as a draft. Also, it looks more than it is 😅

Thanks for implementing PHPStan! 😄

phpstan.neon
```diff
@@ -0,0 +1,19 @@
...
+    phpVersion: 70400
+    level: 6
```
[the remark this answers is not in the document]
DjordyKoert: I am not against increasing it, but I need to slowly work through the different levels in this PR first hahaha 😅 

Locally I will slowly walk through each level (1 -> 2 -> 3 e.t.c.) because even the lower levels are triggering phpstan for quite some things.

After that I can make a better decision about which level would be good, I would prefer 9.

composer.json
DjordyKoert: 👍

phpstan.neon
DjordyKoert: I have seen that this isn't needed because it was able to correctly give the types for `TreeBuilder` and `NodeDefinition` objects for example. This is most likely only needed for the container related features:
> - Notifies you when you try to get an unregistered service from the container.
> - Notifies you when you try to get a private service from the container.

composer.json
DjordyKoert: it would not hurt to not use it 👍 

manual entries to the phpstan config could even help some people understand what it's doing for people not familiair with phpstan

phpstan.neon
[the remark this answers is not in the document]
DominicLuidold: > I am not against increasing it, but I need to slowly work through the different levels in this PR first hahaha 😅

Makes sense :blush: 

> After that I can make a better decision about which level would be good, I would prefer 

That would be optimal, of course. Let's see how feasible it is to do all in one go or if maybe multiple smaller PRs make sense. 🙂

phpstan.neon
DominicLuidold: Even better :smile:

phpstan.neon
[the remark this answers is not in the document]
DjordyKoert: @DominicLuidold I think it is reasonable to stop at level 5 inside this PR. level 6 is reporting **a lot** of errors because of missing typehints:
- `src/`: [ERROR] Found 264 errors
- `tests/`: [ERROR] Found 325 errors

It's probably best to move to level 6 inside a different PR to make sure no accidental breaking changes get added by adding typehints in places.

src/Describer/OpenApiPhpDescriber.php
```diff
@@ -39,15 +39,17 @@ final class OpenApiPhpDescriber
      */
     private $annotationReader;
     private $logger;
-    private $overwrite;
 
     public function __construct(RouteCollection $routeCollection, ControllerReflector $controllerReflector, ?Reader $annotationReader, LoggerInterface $logger, bool $overwrite = false)
     {
+        if ($overwrite || func_num_args() > 4) {
+            trigger_deprecation('nelmio/api-doc-bundle', '4.25.1', 'The "$overwrite" argument of "%s" is unused and therefore deprecated.', __METHOD__);
```
DominicLuidold, line 46: Wouldn't this have to be 
```suggestion
            trigger_deprecation('nelmio/api-doc-bundle', '4.25.2', 'The "$overwrite" argument of "%s" is unused and therefore deprecated.', __METHOD__);
```
as `4.25.1` has already been released?

DominicLuidold (changes requested):
Great effort - I really think this will help the bundle long term! :smile: 

I only found some very minor things :slightly_smiling_face:

phpstan.neon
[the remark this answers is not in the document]
DominicLuidold: > adding phpstan.neon to the .gitignore file, so that you can use a different config locally, e.g. for testing purposes.

If possible, I'd add this as part of this PR as well :)

phpstan.dist.neon
DominicLuidold, line 1: > This change would also open up the possibility of adding phpstan.neon to the .gitignore file, so that you can use a different config locally, e.g. for testing purposes.

If possible, I'd add this as part of this PR as well :)

DominicLuidold (changes requested):

phpstan.dist.neon
DjordyKoert: missed that one, thank you :)

DjordyKoert:
> Great effort - I really think this will help the bundle long term! 😄
> 
> I only found some very minor things 🙂

Thank you very much for reviewing this 😄

DominicLuidold (approved):

Issue #2215, opened 2024-02-07
NelmioApiDocBundle v5 overview


DjordyKoert:
I think we should start preparing for NelmioApiDocBundle v5 and move forward with this bundle.

# PHP 8.1+
First of all I would like to remove support for PHP 7. PHP 7 has been EOL for quite some time now (PHP 7.4 active support stopped 2 years ago https://endoflife.date/php) and is currently preventing us from using new PHP 8 features (typed properties, native attributes, property promotion, union types).

Additionally I would like to start using PHP 8.1 as a minimum instead of PHP 8.0 because:
1. Enums (could replace https://github.com/nelmio/NelmioApiDocBundle/blob/daadb0b567c637d737f58fd25da0b4db2321560d/Render/Html/AssetsMode.php)
2. `new` in initializers https://stitcher.io/blog/new-in-php-81#new-in-initializers-rfc
	- Simplifies some constructors even more by using `new` as default parameters https://github.com/nelmio/NelmioApiDocBundle/blob/daadb0b567c637d737f58fd25da0b4db2321560d/ApiDocGenerator.php#L69 	
3. `readonly` properties
	- Simplifies some classes by replacing getters with a `public readonly` property https://github.com/nelmio/NelmioApiDocBundle/blob/daadb0b567c637d737f58fd25da0b4db2321560d/Model/Model.php#L39

## Planned changes:
- Rewrite every class for constructor property promotion
- Add proper types for every class property

### Static analysis
PHP 8 is also a perfect time to start using static analysis tools. That's why I would like to introduce ([PHPStan](https://phpstan.org/)) to this project. This would also include some refactoring here and there to remove PHPStan's detected problems.

# ~~Symfony version bump to 6.4+~~ Drop support for Symfony 6.0, 6.1, 6.2 & 6.3
Symfony 6.0, 6.1, 6.2 & 6.3 have already stopped receiving bug fixes & security updates which is why a bump to Symfony 6.4 would be ideal, this version also closely align with our new minimum PHP version support.
Additionaly let's keep supporting Symfony 5.4 for now. Support for it should be dropped in a different minor release. Preferably we wait until after November 2024 (end of bug fixes for Symfony 5.4 https://symfony.com/releases/5.4).

# General refactoring
- Remove current deprecations
[...]

PR #2250, opened 2024-04-01
Install & run PHP-CS-Fixer

DominicLuidold:
| Q             | A  |
|---------------|----|
| Bug fix?      | no |
| New feature?  | no |
| Deprecations? | no |
| Issues        | /  |

As there are currently efforts to introduce PHPStan (see #2249), I thought of also introducing PHP-CS-Fixer as a dev dependency (to additionally improve the code quality) and using/upgrading the existing config file. 

**Already done:**
- Introduce [`PHP-CS-Fixer`](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer) as a dev dependency
- Upgrade the existing config
- Run PHP-CS-Fixer on the project files

**Possible further steps as part of this PR that need to be discussed:**
- Replace StyleCI with a PHP-CS-Fixer job as part of the GitHub Actions `CI`/`continuous-integration` workflow (to have a single source of truth regarding code style)?
  - Run PHP-CS-Fixer on same test matrix as PHPUnit[^1] or just on latest supported PHP version (e.g. 8.3)?

[^1]: This approach will not work on PHP 7.4 and PHP 8.0 as some files already use newer syntax leading to failures in these jobs.

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -20,6 +20,5 @@ For the full copyright and license information, please view the LICENSE
 file that was distributed with this source code.
 HEADER
         ],
-        'nullable_type_declaration_for_default_null_value' => false,
     ])
     ->setFinder($finder);
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -1,16 +1,15 @@
 <?php
 
-$finder = PhpCsFixer\Finder::create()
+declare(strict_types=1);
+
+$finder = (new PhpCsFixer\Finder())
     ->in(__DIR__)
     ->exclude('var')
-    ->exclude('Tests/Functional/cache')
-;
+    ->exclude('tests/Functional/cache');
 
-return PhpCsFixer\Config::create()
+return (new PhpCsFixer\Config())
     ->setRules([
         '@Symfony' => true,
-        'ordered_imports' => true,
-        'phpdoc_order' => true,
         'header_comment' => [
             'header' => <<<HEADER
 This file is part of the NelmioApiDocBundle package.
@@ -21,6 +20,6 @@ For the full copyright and license information, please view the LICENSE
 file that was distributed with this source code.
 HEADER
         ],
+        'nullable_type_declaration_for_default_null_value' => false,
     ])
-    ->setFinder($finder)
-;
+    ->setFinder($finder);
--- a/.github/workflows/common/composer-install/action.yaml
+++ b/.github/workflows/common/composer-install/action.yaml
@@ -1,0 +1,51 @@
+name: Composer setup
+description: Setup, install and cache Composer dependencies
+
+inputs:
+  symfony-version:
+    description: The required Symfony version that will be installed
+    required: false
+    type: string
+
+  install-doctrine-annotations:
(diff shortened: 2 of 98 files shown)


.php-cs-fixer.dist.php
```diff
@@ -1,16 +1,15 @@
...
-    ->exclude('Tests/Functional/cache')
-;
+    ->exclude('tests/Functional/cache');
 
-return PhpCsFixer\Config::create()
+return (new PhpCsFixer\Config())
     ->setRules([
         '@Symfony' => true,
-        'ordered_imports' => true,
-        'phpdoc_order' => true,
```
DominicLuidold, line 13: These two rules are already part of the `@Symfony` rule-set

.php-cs-fixer.dist.php
```diff
@@ -21,6 +20,6 @@
 file that was distributed with this source code.
 HEADER
         ],
+        'nullable_type_declaration_for_default_null_value' => false,
```
DominicLuidold, line 23: This is required to maintain BC - possibly a rule that could be removed (meaning following the `@Symfony` rule-set which has set this rule to `true`) as part of #2215

DominicLuidold: Actually, this wouldn't even provoke a BC break, see https://3v4l.org/UkVXL.

So this could definitely be removed already to align with the `@Symfony` rule-set and to be prepared for the next PHP minor version which will deprecate the implicit nullable parameters (see https://wiki.php.net/rfc/deprecate-implicitly-nullable-types).

DjordyKoert: Yeah I agree that we should remove this 👍

DjordyKoert:
> **Possible further steps as part of this PR that need to be discussed:**
> 
> * Replace StyleCI with a PHP-CS-Fixer job as part of the GitHub Actions `CI`/`continuous-integration` workflow (to have a single source of truth regarding code style)?
>   

I also think that we should remove StyleCI in favor of a PHP-CS-Fixer job. I would have to message/email @GuilhemN though asking if he can disable StyleCI for this project before merging this though. As far as I can see there is no easy way to disable it using a simple config change in `.styleci.yml`

>   * Run PHP-CS-Fixer on same test matrix as PHPUnit[1](#user-content-fn-1-6d0da86d1317f5d8de24e36bde6fba6e) or just on latest supported PHP version (e.g. 8.3)?

I think running it on the latest PHP version (8.3) will suffice.

DominicLuidold:
I've refactored the current CI workflow a bit and introduced a [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) that allows to re-use the steps for setting up and installing the Composer dependencies across multiple jobs.

This change should make it quite easy to add a PHPStan CI job in #2249 as well :smile:

DjordyKoert:
Update: I have sent an email to @GuilhemN asking to disable StyleCI for this repo

DjordyKoert:
StyleCI has now been disabled, I will apply my own suggestion and merge it asap

DjordyKoert (changes requested):

DjordyKoert:
Thank you very much @DominicLuidold for your work ❤️

PR #2259, opened 2024-04-06
PHPStan bump to level 6

DjordyKoert:
| Q             | A                                                                                                                         |
|---------------|---------------------------------------------------------------------------------------------------------------------------|
| Bug fix?      | no                                                                                                                    |
| New feature?  | no                                                                    |
| Deprecations? | no                                                   |
| Issues        | Fix #...  |

Follow-up on https://github.com/nelmio/NelmioApiDocBundle/pull/2249. Bump PHPStan to level 6 with required changes.

PR #2265, opened 2024-04-13
fix: add missing trigger_deprecation calls to deprecated classes, methods

DominicLuidold:
| Q             | A   |
|---------------|-----|
| Bug fix?      | no  |
| New feature?  | no  |
| Deprecations? | yes |
| Issues        | /   |

Adds missing deprecation triggers to classes and methods that have been deprecated with a `@deprecated` annotation.


src/Form/Extension/DocumentationExtension.php
```diff
@@ -34,6 +34,14 @@ public function configureOptions(OptionsResolver $resolver): void
...
     public function getExtendedType()
     {
+        trigger_deprecation(
+            'nelmio/api-doc-bundle',
+            '4.26.0',
+            'Calling %s is deprecated since Symfony 4.2, call %s instead',
+            __METHOD__,
+            'DocumentationExtension::getExtendedTypes()',
+        );
+
```
DominicLuidold, line 44: This deprecation is based on the `@deprecated` annotation which will be added to this method as part of #2259.

TODO:
- Update version based on @DjordyKoert's feedback
- Update CHANGELOG.md

src/Form/Extension/DocumentationExtension.php
```diff
@@ -34,6 +34,14 @@ public function configureOptions(OptionsResolver $resolver): void
 
     public function getExtendedType()
     {
+        trigger_deprecation(
+            'nelmio/api-doc-bundle',
+            '4.26.0',
```
DjordyKoert, line 39:
```suggestion
            '4.25.3',
```

src/PropertyDescriber/ArrayPropertyDescriber.php
```diff
@@ -24,6 +24,14 @@ class ArrayPropertyDescriber implements PropertyDescriberInterface, ModelRegistr
 
     public function describe(array $types, OA\Schema $property, ?array $groups = null, ?OA\Schema $schema = null, array $context = [])
```
DjordyKoert, line 25: We should also deprecate using `null` for the `$schema` parameter

DjordyKoert (changes requested):

src/PropertyDescriber/ArrayPropertyDescriber.php
```diff
@@ -27,6 +27,22 @@ class ArrayPropertyDescriber implements PropertyDescriberInterface, ModelRegistr
...
+            );
+        }
+
+        if (null !== $groups) {
+            trigger_deprecation(
+                'nelmio/api-doc-bundle',
+                '4.17.0',
+                'Using the $groups parameter of "PropertyDescriberInterface::describe()" is deprecated and will be removed in a future version. Pass groups via $context[\'groups\']',
+            );
+        }
```
DjordyKoert, line 44: Would this way of checking the parameter be better for triggering deprecations?

```suggestion
		if (null !== $groups || func_num_args() >= 3) {
            trigger_deprecation(
                'nelmio/api-doc-bundle',
                '4.17.0',
                'Using the $groups parameter of "%s" is deprecated and will be removed in a future version. Pass groups via $context[\'groups\'].', __METHOD__
            );
        }

        if (\func_num_args() < 4 || null === $schema) {
            trigger_deprecation(
                'nelmio/api-doc-bundle',
                '4.15.0',
                'The "%s()" method will have a new "OA\Schema $schema" arument not defining it is deprecated.', __METHOD__
            );

            $schema = null;
        } else {
            $schema = func_get_arg(3);
        }
```

DjordyKoert (changes requested):

src/PropertyDescriber/ArrayPropertyDescriber.php
DominicLuidold: @DjordyKoert I'm not entirely sure that relying on the `func_num_args` count is an approach we should take here. Mainly because we deprecate _not_ defining one argument while also deprecating if the other is defined.

For example, with `func_num_args() >= 3` we wouldn't know if someone calls the method like this, which would be totally valid:
```php
$describer->describe(
    types: $types,
    property: $property,
    schema: $schema,
);
```

While `func_num_args() < 4` would not work if someone calls the method like this, even though `$schema` is not defined/`null`:
```php
    types: $types,
    property: $property,
    groups: $groups,
    context: $context,
```

DominicLuidold:
@DjordyKoert Is there any way we could merge this PR? I unfortunately always have to keep up with the latest version changes (and update the PHPUnit baseline) after each release to keep the right version number (see e.g. b36a2191ae786eb5454c64622963a421afc20dfd). :sweat_smile:

DominicLuidold:
> I unfortunately always have to keep up with the latest version changes (and update the PHPUnit baseline) after each release to keep the right version number (see e.g. [b36a219](https://github.com/nelmio/NelmioApiDocBundle/commit/b36a2191ae786eb5454c64622963a421afc20dfd)). 😅

If #2335 is merged before this, that's not necessary anymore (also, some changes would have to be reverted in this PR).

DjordyKoert (approved):

DjordyKoert:
Sorry this took way too long than it should have to merge.

Thank you for the time you spend on this @DominicLuidold ❤️

PR #2266, opened 2024-04-14
fix: Generate operationids properly

lhausammann:
| Bug fix?      | yes                                                                                                                   |
| New feature?  | no                                                                  |
| Deprecations? | no                                                  |
| Issues        | --- |

*Current behaviour*
 - OperationId is not generated when only using symfony attributes, e.g. #[Route]
 - OperationId is not generated when only using one OA attributes, eg. #[Get] in combination with routes
 - However, some combination results in properly generated OperationIds, e.g. using #[Security] in combination with 
   #[Get] attributes.

*Expected behaviour*
 - In all cases an operationId gets generated from the route information (unless it is explicitly set).

*Implementation*
Test cases have been added for those scenarios. The existing OpenApiPhpDescriber has been updated to always reach the part where the operationId is generated.

### DIFF
--- a/tests/Functional/Controller/OperationIdController.php
+++ b/tests/Functional/Controller/OperationIdController.php
@@ -18,6 +18,14 @@ use OpenApi\Attributes as OA;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\Routing\Annotation\Route;
 
+/*
+ * Not all operationIds were generqted properly. This test case covers the following scenarios:
+ * - routes with unnamed symfony route annotations (fall back to SF naming strategy)
+ * - routes combined with OA\Get annotations
+ * - routes with OA\Get annotations and additional ApiDoc root annotations
+ * - routes with operationId explicitly set
+
+ */
 class OperationIdController
 {
     // a route with only a symfony route annotation (generates GET Operation using available metadata as operationId)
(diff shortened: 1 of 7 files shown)


tests/Functional/Controller/OperationIdController.php
```diff
@@ -0,0 +1,69 @@
...
+
+namespace Nelmio\ApiDocBundle\Tests\Functional\Controller;
+
+use Nelmio\ApiDocBundle\Annotation\Security;
+use OpenApi\Attributes as OA;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\Routing\Annotation\Route;
+
+/*
+ * Not all operationIds were generqted properly. This test case covers the following scenarios:
```
rothfahl, line 22:
```suggestion
 * Not all operationIds were generated properly. This test case covers the following scenarios:
```

tests/Functional/Controller/OperationIdController.php
```diff
@@ -0,0 +1,69 @@
...
+
+    // a route with an OA\GET annotation (extends GET operation with operationId)
+    #[Route(path: '/generate/operation_id_get', name: 'get_annotation', methods: 'GET')]
+    #[OA\Get(summary: 'OperationId must be generated automatically if not provided')]
+    public function getMustGenerateOperationIByGetAnnotation(): JsonResponse
+    {
+        return new JsonResponse();
+    }
+
+    // custom operationId is uesed when set explicitly
```
rothfahl, line 62:
```suggestion
    // custom operationId is used when set explicitly
```

DjordyKoert (approved):

DjordyKoert:
Thank you @lhausammann! 😄

PR #2251, opened 2024-04-01
feat: Add ability to configure UI through configuration

HypeMC:
| Q             | A                                                                                                                         |
|---------------|---------------------------------------------------------------------------------------------------------------------------|
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -

Adds the ability to configure the UI through configuration:

```yaml
nelmio_api_doc:
  html_config:
    assets_mode: bundle
    redocly_config:
      expandResponses: '200,201'
      hideDownloadButton: true
    swagger_ui_config:
      deepLinking: true
```

PR #2280, opened 2024-04-24
Respect assets mode config in controller

HypeMC:
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -

Currently the `SwaggerUiController` doesn't respect the `assets_mode` configuration added in #2251.

```yaml
nelmio_api_doc:
  html_config:
    assets_mode: cdn
```

As far as I can tell, there's no reason for the default any more as it can be set through the configuration and overridden in the `DumpCommand` command using the `html-config` option.

### DIFF
--- a/src/Controller/SwaggerUiController.php
+++ b/src/Controller/SwaggerUiController.php
@@ -12,7 +12,6 @@
 namespace Nelmio\ApiDocBundle\Controller;
 
 use Nelmio\ApiDocBundle\Exception\RenderInvalidArgumentException;
-use Nelmio\ApiDocBundle\Render\Html\AssetsMode;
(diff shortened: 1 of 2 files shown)


deluxetom (approved):

DjordyKoert:
Thank you @HypeMC :)

DominicLuidold:
First of all: @HypeMC thanks!

Just as a quick side note: this has (unintentionally?) changed the default asset mode to CDN which was a bit unexpected as the release notes didn't mention that change (we noticed as our API docs weren't working in review environments where the CDN calls got blocked by CSP).

/cc @DjordyKoert

PR #2288, opened 2024-05-22
fix(#2287): symfony assertions (NotBlank) not working in DTO classes

### LINKED ISSUE #2287: [Bug]: Symfony assertions (NotBlank) not working in DTO classes
### Version

4.26.1

### Description

I found issues with '\Symfony\Component\Validator\Constraints\NotBlank'. When merged fix for this bug https://github.com/nelmio/NelmioApiDocBundle/issues/2222 can't set the property required if it has default value. 

Example:

```php
<?php

namespace App\Dto\Api\V1\Response\Contact;

use App\Dto\Api\V1\Response\ResponseDto;
use App\Entity\Contact;

class GetContactsDto
{
    /**
     * @var Contact[]
     */
    protected array $items = [];

    #[\Symfony\Component\Validator\Constraints\NotBlank()]
    protected int $count = 0;

    /**
     * @return Contact[]
     */
    public function getItems(): array
    {
        return $this->items;
    }

    /**
     * Set the value of items.
     *
     * @param Contact[] $items
     *
     * @return self
     */
    public function setItems(array $items)
    {
        $this->items = $items;

        return $this;
    }

    /**
     * Get the value of count.
     */
    public function getCount(): int
    {
        return $this->count;
    }

    /**
     * Set the value of count.
     *
     * @return self
     */
    public function setCount(int $count)
    {
        $this->count = $count;

        return $this;
    }
}
```

JSON: 

```json
"GetContactsDto": {
                "required": [
                ],
                "properties": {
                    "items": {
                        "title": "Set the value of items.",
                        "type": "array",
                        "items": {
                            "$ref": "#/components/schemas/Contact"
                        },
                        "default": []
                    },
                    "count": {
                        "title": "Get the value of count.",
                        "type": "integer",
                        "default": 0
                    }
                },
                "type": "object"
            },
```

This is important when using the OpenAPI documentation in Angular for example. All properties of response DTO classes should be non-undefinable.

I will make a pull request later.

### Additional context

_No response_

heminei:
| Issues        | Fix #2287|

### DIFF
--- a/src/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
+++ b/src/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
@@ -82,10 +82,6 @@ class SymfonyConstraintAnnotationReader
                     return;
                 }
 
-                if (!Generator::isDefault($property->default)) {
-                    return;
-                }
-
                 $existingRequiredFields = Generator::UNDEFINED !== $this->schema->required ? $this->schema->required : [];
                 $existingRequiredFields[] = $propertyName;
 


DjordyKoert:
Could you fix the failing tests? 😄

src/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
```diff
@@ -82,10 +82,6 @@ private function processPropertyAnnotations($reflection, OA\Property $property,
                     return;
                 }
 
-                if (!Generator::isDefault($property->default)) {
-                    return;
-                }
-
```
DominicLuidold, line 88: I don't think that this is the correct approach to solve this issue. Please see https://github.com/nelmio/NelmioApiDocBundle/issues/2287#issuecomment-2138858323 for my thoughts on this and another approach how to fix this outside of the bundle.

PR #2289, opened 2024-05-25
docs: update symfony_attributes.rst with info about required library

Oni4i:
Found out that MapQueryString Attribute doesn't work without component symfony/property-access (ObjectNormalizer::class). And there is no info in doc about that, also the problem is not obvious for regular MapQueryString users as the exception highlight only that there is no normalizer that supports this

| Updating doc?      | yes

### DIFF
--- a/docs/symfony_attributes.rst
+++ b/docs/symfony_attributes.rst
@@ -6,6 +6,17 @@ NelmioApiDocBundle has the ability to automatically create documentation from **
 MapQueryString
 -------------------------------
 
+Install required component
+
+.. code-block:: terminal
+
+    $ composer require symfony/property-access
+
+.. include:: /components/require_autoload.rst.inc
+
+Usage
+-----
+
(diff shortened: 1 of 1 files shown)


docs/symfony_attributes.rst
```diff
@@ -6,6 +6,17 @@ NelmioApiDocBundle has the ability to automatically create documentation from **
 MapQueryString
 -------------------------------
 
+Install required component
+
+.. code-block:: terminal
+
+    $ composer require symfony/property-access
```
DjordyKoert, line 13: I think you mean the `property-info` component instead of `property-access`. `property-access` isn't used anywhere in this project.

Also `symfony/property-info` should be installed together with this project because it's required in `package.json` https://github.com/nelmio/NelmioApiDocBundle/blob/9379c44da4b4e6ef36a81f3135d3d158838a7c16/composer.json#L29C10-L29C31

ToastM8 (approved):
#2324 @

ToastM8 (commented):
@ToastM8
#2289

ToastM8 (approved):

ToastM8:
/ #2289 
```swift

```

PR #2298, opened 2024-06-12
feat(#2297): (Symfony 7.1) Add MapRequestPayload array parameter handling

### LINKED ISSUE #2297: [Bug]: Symfony 7.1 - MapRequestPayload resolves array with new "type" parameter
### Version

4.26.2

### Description

Since Symfony 7.1, the MapRequestPayload has a new parameter `type` that allows for array parameters to be resolved in a Controller. Link to the docs: https://symfony.com/doc/7.1/controller.html#mapping-request-payload
```php
public function dashboard(
    #[MapRequestPayload(type: UserDTO::class)] array $users
): Response
{
    // ...
}
```
This currently results in an error "Schema of type "\array" can't be generated, no describer supports it". And could be resolved by checking for the `type` parameter of the `MapRequestPayload` attribute aswell as checking whether the Controller parameter is of type `array` in the `SymfonyMapRequestPayloadDescriber->describe` function .

### Additional context

_No response_

mazumba:
| Issues        | Fix #2297  |

With Symfony 7.1, a new `type` parameter has been introduced for `MapRequestPayload` that allows array Controller parameters to be resolved. This currently leads to an error when the Describer tries to register a new Model, because the parameter type `array` is used in this case. To resolve the issue, the given `type` parameter form the `MapRequestPayload` should be used instead. 

I tried to test this as gracefully as possible. Any more elegant suggestions are welcome!


mazumba:
Not sure if skipping the test is correct in this case. And I noticed that I was maybe a tiny bit too early with this, since 7.1 is fairly fresh 😅

DjordyKoert:
> Not sure if skipping the test is correct in this case. 

It is 😉 

> And I noticed that I was maybe a tiny bit too early with this, since 7.1 is fairly fresh 😅

You can't start early enough, thanks a lot!

mazumba:
I would love this to also be merged into the `4.x` branch. Do I just create a new PR with 100% successful checks into the `4.x` branch?

DjordyKoert:
Thank you @mazumba :)

PR #2311, opened 2024-07-09
feat: support multiple authorisations to be stored and restored

bobvandevijver:
| Bug fix?      | maybe                                                                                                                    |
| New feature?  | maybe                                                                    |
| Issues        | -  |

I have an API description which has multiple authorisation options, which is supported fine by the Swagger interfaces. However, the store/restore functionality always overrides the previous stored value, so only the last input is restored on page load. This change solves that.

As far as I know everything I used should work fine on relevant browsers.
  - Object.keys: https://caniuse.com/mdn-javascript_builtins_object_keys
  - forEach: https://caniuse.com/mdn-javascript_builtins_array_foreach

I do not believe that the build failures are related to this PR....


public/init-swagger-ui.js
```diff
@@ -25,26 +25,52 @@ function loadSwaggerUI(userOptions = {}) {
 
   const storageKey = 'nelmio_api_auth';
 
-  // if we have auth in storage use it
-  if (sessionStorage.getItem(storageKey)) {
-    try {
-      ui.authActions.authorize(JSON.parse(sessionStorage.getItem(storageKey)));
-    } catch (ignored) {
-      // catch any errors here so it does not stop script execution
+  function getAuthorizationsFromStorage(){
```
bobvandevijver, line 28:
```suggestion
  function getAuthorizationsFromStorage() {
```

public/init-swagger-ui.js
```diff
@@ -25,26 +25,52 @@ function loadSwaggerUI(userOptions = {}) {
...
+    if (sessionStorage.getItem(storageKey)) {
+      try {
+        return JSON.parse(sessionStorage.getItem(storageKey))
```
bobvandevijver, line 31:
```suggestion
        return JSON.parse(sessionStorage.getItem(storageKey));
```

bobvandevijver:
@DjordyKoert Anything I can do to get this merged?

DjordyKoert:
Sorry for the late response.

It's generally quite cumbersome to test / validate that any changes to the javascript code will function 😢

DjordyKoert:
Thank you for the contribution at least ❤️

PR #2329, opened 2024-08-02
fix(SymfonyConstraintAnnotationReader): disallow null if NotNull attribute is present

### LINKED ISSUE #2309: [Bug]: Using #[NotNull] / #[NotBlank] doesn't influence nullable
### Version

4.26.1

### Description

It is possible (/feasible?) to apply `#[NotNull]` / `#{NotBlank]` constraints on nullable properties in some scenarios (i.e.: not let Symfony serializer error when value is not provided by allowing it to assign `null` instead). But while these annotations are handled by this bundle, they don't seem to affect `nullable`, but only `required` of the object type?

So IMO the `SymfonyConstraintAnnotationReader` should not just only update the list of required fields, but also do `$property->nullable = false;` (or at least if it's unset / default)

### Additional context

_No response_

jeremyVignelles:
| Issues        | Fix #2309 |

if a property is set as something like this:

```php
#[OA\Property()]
#[Assert\NotNull(groups: ['validation'])]
private ?string $prop = null;
```

that property will be marked as nullable AND required.
This PR ensures that the nullable property is also set appropriately when a NotNull is set

NOTE: PHPUnit already fails on master

### DIFF
--- a/src/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
+++ b/src/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
@@ -90,6 +90,7 @@ class SymfonyConstraintAnnotationReader
                 $existingRequiredFields[] = $propertyName;
 
                 $this->schema->required = array_values(array_unique($existingRequiredFields));
+                $property->nullable = false;
             } elseif ($annotation instanceof Assert\Length) {
                 if (isset($annotation->min)) {
                     $property->minLength = $annotation->min;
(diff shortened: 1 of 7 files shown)


DjordyKoert (approved):

DjordyKoert:
Could you update with master? :)

jeremyVignelles:
Done, unit tests and code style fixed, commits squashed into one :)
You're good to go :)

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,22 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
```
DjordyKoert, line 1: Could you remove this file?

DjordyKoert (changes requested):

.devcontainer/devcontainer.json
jeremyVignelles: Sorry, was careful about not pushing it the first time but forgot about it when rebasing.
The file has been removed.

DjordyKoert (approved):

DjordyKoert:
Thank you :)

PR #2334, opened 2024-08-14
feat: create top level Tag from Tag annotations

### LINKED ISSUE #2333: [Feature]: Create top level OpenApi Tag when processing Tag annotation
### Description

When processing `OA\Tag` annotation inside the `OpenApiPhpDescriber`, only the `name` of the tag is used and added to the `Operation::$tags` list. Then Swagger-php create the missing top level `Tag`s when encountering an unknown tag name used in `Operation`.
The issue is that the additional informations provided in the `OA\Tag` annotation (`description` and `externalDocs`) are lost.
The `OpenApiPhpDescriber` can easily create the top level `Tag` on the fly.

_I'll make a pull request shortly after creating this issue ;)_

### Additional context

_No response_

noemi-salaun:
Close #2333

| Issues        | Fix #2333  |

When processing `OA\Tag` annotation inside the `OpenApiPhpDescriber`, the corresponding top level OpenApi Tag is created on the fly, so the description and the externalDocs are not lost. If the same Tag name is encountered multiple time, it is merged into the already existing top level tag.

_There are 3 failing tests and 2 phpstan errors that are not related to my change. They were already there before my changes._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+next
+-----
+* Create top level OpenApi Tag from Tags top level annotations/attributes
+
 4.26.0
 -----
 * Add ability to configure UI through configuration
--- a/src/Describer/OpenApiPhpDescriber.php
+++ b/src/Describer/OpenApiPhpDescriber.php
@@ -135,6 +135,9 @@ final class OpenApiPhpDescriber
                     $annotation->validate();
                     $mergeProperties->tags[] = $annotation->name;
 
+                    $tag = Util::getTag($api, $annotation->name);
+                    $tag->mergeProperties($annotation);
+
                     continue;
                 }
 
--- a/src/OpenApiPhp/Util.php
+++ b/src/OpenApiPhp/Util.php
@@ -71,6 +71,30 @@ final class Util
     }
 
     /**
+     * Return an existing Tag object from $api->tags[] having its member name set to $name.
+     * Create, add to $api->tags[] and return this new Tag object and set the property if none found.
+     *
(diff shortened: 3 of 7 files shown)


DjordyKoert:
Could you also write a functional test for this which tests your changes to the `OpenApiPhpDescriber`? :)

noemi-salaun:
@DjordyKoert Do tests already exist for the OpenApiPhpDescriber so I can complete them? Or can you point me to another functional test I can take inspiration from? Please

DjordyKoert:
I think the easiest way to test this is to createa a minimal replication your configuration which can validate that your changes now do what it's supposed to do.

New tests cases can be added here (by adding a `yield`) https://github.com/nelmio/NelmioApiDocBundle/blob/master/tests/Functional/ControllerTest.php#L88

1. Create a new controller for your test https://github.com/nelmio/NelmioApiDocBundle/tree/master/tests/Functional/Controller
2. Create any classes (DTO's for example) here https://github.com/nelmio/NelmioApiDocBundle/tree/master/tests/Functional/Entity and use these in your controller class
3. Run the test. On first run it should create a `.json` for your newly made controller. Validate if the generate openapi json is correct https://github.com/nelmio/NelmioApiDocBundle/tree/master/tests/Functional/Fixtures

noemi-salaun:
@DjordyKoert 
Thanks for the guide, that is perfect.
I added a very simple and focused test case inspired by what I have seen in other test Controllers
It was simpler than I thought :+1:

DjordyKoert (approved):

DjordyKoert:
Thank you @noemi-salaun ❤️

Issue #2341, opened 2024-09-22
[Question]: Is it possible to configure a different serializer per endpoint?


herndlm:
### Version

4.24.0

### Question

I have a project using FOSRestBundle and want to switch new endpoints from JMS to Symfony Serializer. To have the correct types in the API docs inferred from the new models while not breaking old ones, I'd basically would need to be able to disable `nelmio_api_doc.models.use_jms` for new endpoints only.

I also asked a similar thing via https://github.com/symfony/symfony/discussions/58348

Any ideas or would a contribution also be accepted if this is currently impossible?

herndlm:
Fyi it looks like I'll end up with a local patch where I add the following to `JMSModelDescriber::supports()`:
```php
if (($model->getOptions()['useJms'] ?? null) === false) {
	return false;
}
```

As well as setting `options: ['useJms' => false]` in `SymfonyMapRequestPayloadDescriber` and `SymfonyMapQueryStringDescriber` where it instantiates `Model`. Those, IMO can never use JMS serializer anyway?

That "fixes" Swagger docs for those Symfony attributes as well as allowing me to opt-out of JMS Serializer usage for specific responses via e.g.
```php
#[OA\Response(response: Response::HTTP_OK, description: 'OK', content: new Model(type: MyModel::class, groups: ['read'], options: ['useJms' => false]))]
```

Not sure if that's a good approach or not, it feels not that invasive and should help us move forward while keeping BC.

I could contribute this here too of course if a there is a need for it.

kevinpapst:
Thanks @herndlm for your work on that PR. 

Are you open to share your reasons for switching from JMS to Symfony? 
I have the same setup: FOSRestBundle with JMS Serializer.

While goetas is doing a great job with both bundles, maintenance is pretty low and I am a bit afraid of the future for both bundles. So your PR peaked my interest and I wonder if you could share some thoughts about your move away from JMS.

Issue #1886, opened 2021-10-12
Symfony serializer groups


pkly:
Hi, we have jms installed because of legacy reasons, but we use the symfony serializer for all our entities, as a lot of things happen real-time.
The issue is recently we started using the symfony serializer groups, and were unable to get them working in nelmio easily.
Is there a way I can actually have fields that have specific groups shown when displaying my models in nelmio? It seems it's only supporting JMS groups.

pkly:
Yes, all it does is make all my models disappear.
Since the symfony serializer is much better in 5.x and even in 4.4 which we're still using it's perfectly fine, we'd rather not use jms simply so we can have 1:1 api stuff.
I could probably look into it more, I'm not quite sure how the JMS groups are found and detected, but if the bundle itself checks the fields then a callback check could be added (possibly hidden behind a config option by default?) for the symfony annotation for groups?
the `@SerializedName` also aren't supported, which is just irritating more than anything, for now we just create separate classes for the api responses so it's clean enough

GuilhemN:
We're trying to replicate the behavior of the serializers, and only properties that would  be deserialized/serialized by the Symfony serializer appear in the Nelmio documentation.

One possibility I see is that your fields are configured with JMS groups annotations but not with the Symfony `@Groups` annotation, are they?
Without further details on how you call `@Model` and the look of your models themselves it will be complicated to help further unfortunately.

pkly:
We do have JMS installed, but don't actually use it (leftovers), instead we use the Symfony groups.

What I wanted to be able to do is to pass an entity class with the groups for symfony serialization, but i'm pretty sure they're just ignored instead.

herndlm:
please check if https://github.com/nelmio/NelmioApiDocBundle/pull/2342 might be useful for this case

pkly:
This is a 3 year old issue and I don't even remember what exactly the issue is. I'm pretty sure we've since removed jms completely and it works fine. The issue was never disabling jms on context, it was that without jms you'd get no results, if I remember correctly.

PR #2342, opened 2024-09-23
feat(2056): support opt out of JMS serializer usage

### LINKED ISSUE #2056: [feature] Dinamic configure the serializer for model. Really?
When you install JMS serializer, the bundle is automatically configured to use it.
Have you tried adding in your config:
```
nelmio_api_doc:
    models: { use_jms: false }
```
?
(from the documentation: https://symfony.com/bundles/NelmioApiDocBundle/current/index.html#general-php-objects)

_Originally posted by @GuilhemN in https://github.com/nelmio/NelmioApiDocBundle/issues/1886#issuecomment-941089881_

---

@GuilhemN,

Be able to configure a serializer for a method in a controller, or even for the entire controller class, or for a model. Models are better.

In large projects, migrating from one serializer to another is long and problematic, it would be great to be able to configure this parameter for each model, by default - a common `use_jms' ...

example: 
`@SWG\Schema(ref=@Model(type=Generate\Command::class, options={"use_jms": false}, groups={"default"}))`

Best solution - jms to symfony annotation/attributes conveter 😂

herndlm:
| Issues        | Fix #2056  #2341  |

Always registers `JMSModelDescriber` when the bundle is active, independently of the configured global `models.use_jms` option.
If `models.use_jms` is set, then `JMSModelDescriber` will be added with a priority of 50, just as before. If `models.use_jms` is not set, it will be added with -50 so that `ObjectModelDescriber` is used instead. This should not lead to any behavior change.

Adds support to `options.useJms` in the model by early exiting in either `JMSModelDescriber` or `ObjectModelDescriber` when it is either explicitly `false` or `true`. Together with the priorities of the describer this allows to have either JMS or Symfony serializer as default and switch per model if necessary.

In https://github.com/nelmio/NelmioApiDocBundle/issues/2341 I played around with this in an app and it worked as expected.

(sorry btw, I'm a fan of rebase + force-push and realized too late that mentioning the issue in the first commit is a bad idea then because GitHub spams whenever I push..)

Closes https://github.com/nelmio/NelmioApiDocBundle/issues/2056
Closes https://github.com/nelmio/NelmioApiDocBundle/issues/2341

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # CHANGELOG
 
+## next
+
+* Added support to opt out of JMS serializer usage per endpoint by setting `useJms` in the serializationContext.
+  ```php
+  #[OA\Response(response: 200, content: new Model(type: UserDto::class, serializationContext: ["useJms" => false]))]
+  ```
+
 ## 4.30.0
 * Create top level OpenApi Tag from Tags top level annotations/attributes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # CHANGELOG
 
-## next
+## 4.31.0
 
 * Added support to opt out of JMS serializer usage per endpoint by setting `useJms` in the serializationContext.
   ```php
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -509,6 +509,18 @@ General PHP objects
         nelmio_api_doc:
             models: { use_jms: false }
 
+    Alternatively, it is also possible to opt out of JMS serializer usage per endpoint by setting `useJms` in the serializationContext:
+
+    .. configuration-block::
+
+        .. code-block:: php-annotations
+
+            /** @OA\Response(response=200, @Model(type=UserDto::class, serializationContext={"useJms"=false})) */
+
+        .. code-block:: php-attributes
+
+            #[OA\Response(response: 200, content: new Model(type: UserDto::class, serializationContext: ["useJms" => false]))]
+
     When using the JMS serializer combined with `willdurand/Hateoas`_ (and the `BazingaHateoasBundle`_),
     HATEOAS metadata are automatically extracted
 
--- a/src/ModelDescriber/JMSModelDescriber.php
+++ b/src/ModelDescriber/JMSModelDescriber.php
@@ -261,6 +261,10 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
 
     public function supports(Model $model): bool
     {
+        if (($model->getSerializationContext()['useJms'] ?? null) === false) {
+            return false;
+        }
+
         $className = $model->getType()->getClassName();
 
         try {
--- a/tests/Functional/Configs/JMS.yaml
+++ b/tests/Functional/Configs/JMS.yaml
@@ -1,0 +1,3 @@
+nelmio_api_doc:
+  models:
+    use_jms: true
--- a/tests/Functional/Controller/JmsOptOutController.php
+++ b/tests/Functional/Controller/JmsOptOutController.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the NelmioApiDocBundle package.
+ *
+ * (c) Nelmio
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Nelmio\ApiDocBundle\Tests\Functional\Controller;
+
+use Nelmio\ApiDocBundle\Annotation\Model;
+use Nelmio\ApiDocBundle\Tests\Functional\Entity\JMSUser;
+use OpenApi\Attributes as OA;
+use Symfony\Component\Routing\Annotation\Route;
+
+#[Route(host: 'api.example.com')]
+final class JmsOptOutController
+{
+    #[Route('/api/jms', methods: ['GET'])]
+    #[OA\Response(
+        response: 200,
+        description: 'Success',
+        content: new Model(type: JMSUser::class)
+    )]
+    public function jms()
+    {
+    }
+
+    #[Route('/api/jms_opt_out', methods: ['GET'])]
+    #[OA\Response(
+        response: 200,
+        description: 'Success',
+        content: new Model(type: JMSUser::class, serializationContext: ['useJms' => false])
+    )]
+    public function jmsOptOut()
+    {
+    }
+}
--- a/tests/Functional/ControllerTest.php
+++ b/tests/Functional/ControllerTest.php
@@ -11,8 +11,10 @@
 
 namespace Nelmio\ApiDocBundle\Tests\Functional;
 
+use JMS\SerializerBundle\JMSSerializerBundle;
 use OpenApi\Annotations as OA;
 use Symfony\Component\HttpKernel\Attribute\MapRequestPayload;
+use Symfony\Component\HttpKernel\Bundle\Bundle;
 use Symfony\Component\HttpKernel\Kernel;
 use Symfony\Component\Routing\Loader\Configurator\RoutingConfigurator;
 
@@ -42,9 +44,10 @@ final class ControllerTest extends WebTestCase
      * @dataProvider provideUniversalTestCases
      *
      * @param array{name: string, type: string}|null $controller
+     * @param Bundle[]                               $extraBundles
      * @param string[]                               $extraConfigs
      */
-    public function testControllers(?array $controller, ?string $fixtureName = null, array $extraConfigs = []): void
+    public function testControllers(?array $controller, ?string $fixtureName = null, array $extraBundles = [], array $extraConfigs = []): void
     {
         $controllerName = $controller['name'] ?? null;
         $controllerType = $controller['type'] ?? null;
@@ -59,7 +62,7 @@ final class ControllerTest extends WebTestCase
             $routes->withPath('/')->import(__DIR__."/Controller/$controllerName.php", $controllerType);
         };
 
-        $this->configurableContainerFactory->create([], $routingConfiguration, $extraConfigs);
+        $this->configurableContainerFactory->create($extraBundles, $routingConfiguration, $extraConfigs);
 
         $apiDefinition = $this->getOpenApiDefinition();
 
@@ -88,7 +91,18 @@ final class ControllerTest extends WebTestCase
                 'type' => $type,
             ],
             'PromotedPropertiesDefaults',
+            [],
             [__DIR__.'/Configs/AlternativeNamesPHP81Entities.yaml'],
+        ];
+
+        yield 'JMS model opt out' => [
+            [
+                'name' => 'JmsOptOutController',
+                'type' => $type,
+            ],
+            'JmsOptOutController',
+            [new JMSSerializerBundle()],
+            [__DIR__.'/Configs/JMS.yaml'],
         ];
 
         if (version_compare(Kernel::VERSION, '6.3.0', '>=')) {
@@ -110,6 +124,7 @@ final class ControllerTest extends WebTestCase
                     'type' => $type,
                 ],
                 'MapQueryStringCleanupComponents',
+                [],
                 [__DIR__.'/Configs/CleanUnusedComponentsProcessor.yaml'],
             ];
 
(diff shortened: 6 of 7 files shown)


herndlm:
phpstan and unit tests were already not working locally for me and I didn't do much debugging yet :/

src/Model/Model.php
```diff
@@ -65,7 +65,7 @@ public function getSerializationContext(): array
 
     public function getHash(): string
     {
-        return md5(serialize([$this->type, $this->getSerializationContext()]));
+        return md5(serialize([$this->type, $this->serializationContext, $this->options]));
```
herndlm, line 68: ah dang, I noticed that this is actually breaking something in my use case. My model includes an property with a list ob objects and a PHPDoc type-hint of for them. https://github.com/nelmio/NelmioApiDocBundle/blob/277fa17b912be31b5170e49f0924b1028058c938/src/PropertyDescriber/ObjectPropertyDescriber.php#L47 is creating the model, but because no options from the original model are passed it leads to a different hash and then incorrect usage of the JMS serializer :/

herndlm: Changing that interface to pass through the options is not OK I suppose, using context to add `useJms` would be fine though, since it is passed through already, but feels like a hack

DjordyKoert: `options` is only used for [`symfony/form`](https://symfony.com/doc/current/forms.html#passing-options-to-forms). 

I would rather add `useJms` into `serializationContext`. This would also prevent us from having to add another parameter `options` to every describer (and `PropertyDescriberInterface`). Are you able to change this? 😃 

> the model could also become a new useJms argument / property itself

I would rather not add a new property specific for this use-case to the `Model` class, it would start cluttering the constructor quite a bit if we were to do this everytime 😄

DjordyKoert (changes requested):

src/Model/Model.php
herndlm: of course, that makes sense. sure, I'll adapt it right away 😊

herndlm: done. only one thing I'm not super happy about - the changes (setting useJms: false) in `SymfonyMapRequestPayloadDescriber` lead to model schema duplication in a existing test for `Controller2209` because the same model is used before already. In that case, since JMS is off, changing the serializationContext does not add any benefits. Not sure if we're just fine with this or I should e.g. inject the global `useJms` and only explicitly set it to false IF JMS is really active.

WDYT?

DjordyKoert: There shouldn't be any need to set `useJms` there. 

If users want to set that value they can explicely set it on the `#[MapRequestPayload]` attribute: 
```php

#[MapRequestPayload(serializationContext: ['useJms' => true])]
```

src/RouteDescriber/RouteArgumentDescriber/SymfonyMapQueryStringDescriber.php
```diff
@@ -38,8 +38,9 @@ public function describe(ArgumentMetadata $argumentMetadata, OA\Operation $opera
 
         $modelRef = $this->modelRegistry->register(new Model(
             new Type(Type::BUILTIN_TYPE_OBJECT, $argumentMetadata->isNullable(), $argumentMetadata->getType()),
-            groups: $this->getGroups($attribute),
-            serializationContext: $attribute->serializationContext,
+            $this->getGroups($attribute),
+            null,
+            ['useJms' => false] + $attribute->serializationContext
```
DjordyKoert, line 43:
```suggestion
            $attribute->serializationContext
```

src/RouteDescriber/RouteArgumentDescriber/SymfonyMapRequestPayloadDescriber.php
```diff
@@ -44,8 +44,9 @@ public function describe(ArgumentMetadata $argumentMetadata, OA\Operation $opera
 
         $modelRef = $this->modelRegistry->register(new Model(
             new Type(Type::BUILTIN_TYPE_OBJECT, false, $typeClass),
-            groups: $this->getGroups($attribute),
-            serializationContext: $attribute->serializationContext,
+            $this->getGroups($attribute),
+            null,
+            ['useJms' => false] + $attribute->serializationContext
```
DjordyKoert, line 49:
```suggestion
            $attribute->serializationContext
```

DjordyKoert (changes requested):

src/Model/Model.php
herndlm: what I slightly dislike there is that e.g. `MapRequestPayload` is always deserialized using Symfony serializer (via https://github.com/symfony/symfony/blob/3c7124ea938b741f5bfffd97364320eac8232d63/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php).

But if I have JMS installed, it would try to use JMS for Swagger docs instead, which is always wrong IMO. And my thought was that setting `'useJms' => false` would make it properly describe the DTO out of the box even if JMS is installed.

src/ModelDescriber/ObjectModelDescriber.php
```diff
@@ -235,6 +235,10 @@ private function describeProperty(array $types, Model $model, OA\Schema $propert
 
     public function supports(Model $model): bool
     {
+        if (($model->getSerializationContext()['useJms'] ?? null) === true) {
+            return false;
+        }
+
```
DjordyKoert, line 241: This check now always requires `useJms` to be defined, which will cause this describer to never be run UNLESS it's defined as `['useJms' => true]` on every model.

This check can be removed by relying on the order (priority) of the called describers.

 ```suggestion
```

**Review by DjordyKoert — CHANGES_REQUESTED** [review]

File: src/Model/Model.php (hunk 1 above)
Comment (herndlm) [reply src/Model/Model.php]: it's really just relevant if JMS is used by default. I think we'll turn of JMS by default and only enable it for specific old endpoints and then this would not affect me. but it still bothers me slightly 😅

Comment (DjordyKoert) [reply src/Model/Model.php]: I agree that it sounds weird but I would still prefer the user of this bundle to explicitly define how they want it to behave.

The current master version is currently also using the `JMSSerializer` whenever the jms is available for describing models from `#[MapRequestPayload]`.

File: src/ModelDescriber/ObjectModelDescriber.php (hunk 4 above)
Comment (herndlm) [reply src/ModelDescriber/ObjectModelDescriber.php:241]: sorry, I don't fully get what you mean, the order of those has been dealt with, this code only makes sure that if Symfony is the default (higher order) one but `useJms` is true that it actually uses JMS instead.

Comment (herndlm) [reply src/ModelDescriber/ObjectModelDescriber.php:241]: ah wait, I think I got it. you mean the `JMSModelDescriber` should _always_ run first and that way we don't need this, right?

Comment (herndlm) [reply src/ModelDescriber/ObjectModelDescriber.php:241]: wait no, that doesn't work. I'd like to be able to have Symfony by default (`nelmio_api_doc.models.use_jms: false`) but still use JMS for some endpoints BUT preferably not have configure `'useJms' => false` in all new endpoints, but only  `'useJms' => true` in the old ones. If that makes sense?

Comment (DjordyKoert) [reply src/ModelDescriber/ObjectModelDescriber.php:241]: > ah wait, I think I got it. you mean the JMSModelDescriber should always run first and that way we don't need this, right? I made it more dynamical but that leads to this conditional being needed.

Correct, the priority of 50 is going to ensure the `JMSModelDescriber` is ran before the others.

Comment (herndlm) [reply src/ModelDescriber/ObjectModelDescriber.php:241]: I agree, it feels a bit dirty. What about this
- we use the global `use_jms` to configure listener priority to have no behavior changes
  - if it's not configured explicitly, it will be set to true if the bundle is there
  - if it's set to true the JMS describer will run first
  - if it's set to false the ObjectModel describer will run fist
- instead of adding `useJms` I could add a `disableJms` and `disableSymfony` config. not sure about the names, but it should suggest that it only leads to disabling one of them
- JMS describer exits early if `disableJms` is configured
- ObjectModel describer exits early if `disableSymfony` (or however we call it) is configured

that way less magic is involved and people can still have either one by default and prefer the other one for specific endpoints by disabling the default one

Comment (DjordyKoert) [reply src/ModelDescriber/ObjectModelDescriber.php:241]: > but still use JMS for some endpoints BUT preferably not have configure 'useJms' => false in all new endpoints, but only 'useJms' => true in the old ones. If that makes sense?

From your issue I read:
> I'd basically would need to be able to disable nelmio_api_doc.models.use_jms for new endpoints only.

This is of course a personal preference but IMO using a way to opt-out on your new endpoints is better than deciding to opt-in on your old endpoints. That way there is:
1. No need to change your config (`nelmio_api_doc.models.use_jms`)
2. You can globally disable `useJms` and you won't necessarily have to edit your `#[MapRequestPayload]`'s after all your old endpoint are gone.

Comment (herndlm) [reply src/ModelDescriber/ObjectModelDescriber.php:241]: would you accept the opt out of both JMS or Symfony I suggested in the previous comment? that could be a good and simple solution to deal with cases in both direction.

Comment (herndlm) [reply src/ModelDescriber/ObjectModelDescriber.php:241]: ok that's reasonable. I hoped I could somehow opt-in to JMS 😊 but we can deal with this. I'll adapt the PR so that it only allows us to opt-out of JMS.

Comment (DjordyKoert) [reply src/ModelDescriber/ObjectModelDescriber.php:241]: > I hoped I could somehow opt-in to JMS

You could of course decorate the `ObjectModelDescriber` and add this logic yourself on your project(s) if this is something you want / need. 😄 
https://symfony.com/doc/current/service_container/service_decoration.html

Comment (herndlm) [reply src/ModelDescriber/ObjectModelDescriber.php]: ok done, we successfully discussed away most of the original code 😂

**Comment by herndlm** [comment]
just for the record if somebody is looking for this: if opting out of Symfony serializer is needed for special reasons - it could be achieved with a patch as simple as the following:

```patch
diff --git a/src/ModelDescriber/ObjectModelDescriber.php b/src/ModelDescriber/ObjectModelDescriber.php
index f119bde..d1ed4f3 100644
--- a/src/ModelDescriber/ObjectModelDescriber.php
+++ b/src/ModelDescriber/ObjectModelDescriber.php
@@ -235,6 +235,10 @@ class ObjectModelDescriber implements ModelDescriberInterface, ModelRegistryAwar

     public function supports(Model $model): bool
     {
+        if (($model->getSerializationContext()['useSymfony'] ?? null) === false) {
+            return false;
+        }
+
         return Type::BUILTIN_TYPE_OBJECT === $model->getType()->getBuiltinType()
             && (class_exists($model->getType()->getClassName()) || interface_exists($model->getType()->getClassName()));
     }
```

DjordyKoert (approved):

DjordyKoert:
Thank you very much for your contributions @herndlm 😄

kevinpapst:
I am describing my models in YAML:
```yaml
nelmio_api_doc:
    models:
        use_jms: true
        names:
            - { alias: FooBar, type: App\Entity\FooBar, groups: [Default, Entity] }
```
and annotation
```
    #[OA\Response(response: 200, description: '...', content: new OA\JsonContent(type: 'array', items: new OA\Items(ref: '#/components/schemas/FooBar')))]
```
Any chance I can use that new feature? 
If not, any hint how/where to start implementing support?

herndlm:
> I am describing my models in YAML:
> 
> ```yaml
> nelmio_api_doc:
>     models:
>         use_jms: true
>         names:
>             - { alias: FooBar, type: App\Entity\FooBar, groups: [Default, Entity] }
> ```
> 
> and annotation
> 
> ```
>     #[OA\Response(response: 200, description: '...', content: new OA\JsonContent(type: 'array', items: new OA\Items(ref: '#/components/schemas/FooBar')))]
> ```
> 
> Any chance I can use that new feature? If not, any hint how/where to start implementing support?

seems to be not possible right now, but it look like just 2 places have to be adapted.

- https://github.com/nelmio/NelmioApiDocBundle/blob/322c47b9fa3c5913ec9bbeba537a32407151be73/src/DependencyInjection/NelmioApiDocExtension.php#L304-L305
- https://github.com/nelmio/NelmioApiDocBundle/blob/322c47b9fa3c5913ec9bbeba537a32407151be73/src/Model/ModelRegistry.php#L70

I think being able to specify the serializationContext there too makes sense and I can open a PR for it as well. In the end @DjordyKoert has to decide I suppose :)

kevinpapst:
Feature parity would be great ❤️

herndlm:
fyi this might close also https://github.com/nelmio/NelmioApiDocBundle/issues/1886

PR #2344, opened 2024-09-25
fix: pass through serializationContext into models where possible

herndlm:
| New feature?  | o                                                                    |
| Issues        | Fix #...  |

I noticed that information I'm adding into a JMS serializer model gets lost inside it because `serializationContext` is not passed through everywhere, but it most likely should? the thing I slightly dislike though is that it _could_ include `groups` and in the scope of `JMSModelDescriber` where groups are determined in another way and overwritten I have to unset that key from the serializationContext to avoid side-effects. WDYT @DjordyKoert?

### DIFF
--- a/src/ModelDescriber/JMSModelDescriber.php
+++ b/src/ModelDescriber/JMSModelDescriber.php
@@ -200,7 +200,7 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
                 continue;
             }
 
-            $this->describeItem($item->type, $property, $context);
+            $this->describeItem($item->type, $property, $context, $model->getSerializationContext());
             $context->popPropertyMetadata();
         }
         $context->popClassMetadata();
@@ -281,8 +281,9 @@ class JMSModelDescriber implements ModelDescriberInterface, ModelRegistryAwareIn
      * @internal
      *
      * @param mixed[] $type
+     * @param mixed[] $serializationContext
      */
-    public function describeItem(array $type, OA\Schema $property, Context $context): void
+    public function describeItem(array $type, OA\Schema $property, Context $context, array $serializationContext): void
     {
         $nestedTypeInfo = $this->getNestedTypeInArray($type);
         if (null !== $nestedTypeInfo) {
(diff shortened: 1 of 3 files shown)


src/ModelDescriber/JMSModelDescriber.php
```diff
@@ -333,8 +334,9 @@ public function describeItem(array $type, OA\Schema $property, Context $context)
             }
 
             $groups = $this->computeGroups($context, $type);
+            unset($serializationContext['groups']);
```
DjordyKoert, line 337: It shouldn't be needed to unset this right?  The `groups` key already gets overwritten if it's set

https://github.com/nelmio/NelmioApiDocBundle/blob/master/src/Model/Model.php#L40-L42

herndlm: I need to debug this further I guess, but apparently because it stays in `serializationContext` it can influence something and leads to JMSFunctionalTest failing. But maybe this a symptom of something else that should be fixed?

```
1) Nelmio\ApiDocBundle\Tests\Functional\JMSFunctionalTest::testModelPictureDocumentation
Model "JMSPicture" does not exist.
Failed asserting that false is not false.

2) Nelmio\ApiDocBundle\Tests\Functional\JMSFunctionalTest::testModeChatDocumentation
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
     'type' => 'object'
     'properties' => Array (
         'picture' => Array (
-            '$ref' => '#/components/schemas/JMSPicture'
+            '$ref' => '#/components/schemas/JMSPicture_mini'
         )
     )
     'schema' => 'JMSChatUser'
 )
```

herndlm: ah I'm an idiot, exactly what you linked is the issue. groups for this test case is `null` and then it gets (incorrectly) set to what it got from the serializationContext.

herndlm: but what I mean is, I have to unset it to avoid unwanted side-effects there

DjordyKoert (approved):

PR #2345, opened 2024-09-27
feat: support configuring all params in alternative name models

herndlm:
| Issues        | Fix #...  |

Adds feature parity to model creation in yaml via `nelmio_api_doc.models.names` by also supporting `options` and `serializationContext`.

Tested in a local project, e.g.
```yaml
    models:
        names:
            - { alias: Error, type: App\Model\Error, options:  { foo: bar }, serializationContext: { useJms: false }  }
```

lead to
grafik

inspired by https://github.com/nelmio/NelmioApiDocBundle/pull/2342#issuecomment-2372260584
fyi @kevinpapst

it's the first time I touch this Symfony Config TreeBuilder stuff btw..


herndlm:
fyi @DjordyKoert

src/DependencyInjection/Configuration.php
```diff
@@ -175,6 +175,17 @@ public function getConfigTreeBuilder(): TreeBuilder
...
                                         ->end()
                                     ->end()
+                                    ->variableNode('options')
+                                        ->defaultValue(null)
+                                        ->validate()
+                                            ->ifTrue(function ($v) { return null !== $v && !is_array($v); })
+                                            ->thenInvalid('Model options must be either `null` or an array.')
+                                        ->end()
+                                    ->end()
+                                    ->arrayNode('serializationContext')
```
kevinpapst, line 185: Might be good to validate the possible keys? Otherwise people can put in arbitrary data here.
Better be strict about it, because if you e.g. use a wrong name, you then get a clear hint from the Symfony exception which keys are available.

DjordyKoert: It should be possible to use any value here. This might even be used for users who do some logic based inside their custom-made describers.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## next
+
+* Added support to configure `options` and `serializationContext` via `nelmio_api_doc.models.names` as well.
```
DjordyKoert, line 5:
```suggestion
* Added support to configure `options` and `serializationContext` via `nelmio_api_doc.models.names`.
```

src/DependencyInjection/Configuration.php
```diff
@@ -175,6 +175,17 @@ public function getConfigTreeBuilder(): TreeBuilder
                                             ->thenInvalid('Model groups must be either `null` or an array.')
                                         ->end()
                                     ->end()
+                                    ->variableNode('options')
+                                        ->defaultValue(null)
+                                        ->validate()
+                                            ->ifTrue(function ($v) { return null !== $v && !is_array($v); })
```
DjordyKoert, line 181: Missing test for this

tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -113,6 +113,21 @@ public function testAlternativeNames(): void
...
+                    ],
+                    [
+                        'alias' => 'Foo1',
+                        'type' => 'App\Foo',
+                        'options' => ['foo' => 'bar'],
+                    ],
+                    [
+                        'alias' => 'Foo1',
+                        'type' => 'App\Foo',
+                        'serializationContext' => ['useJms' => false],
```
DjordyKoert, line 129: Lets also test that anything can be used here

```suggestion
                        'serializationContext' => ['useJms' => false, 'foo' => 'bar'],
```

tests/Model/ModelRegistryTest.php
```diff
@@ -117,6 +119,8 @@ public function testNameCollisionsAreLoggedWithAlternativeNames(): void
             $ref->getShortName() => [
                 'type' => $ref->getName(),
                 'groups' => ['group1'],
+                'options' => null,
+                'serializationContext' => [],
```
DjordyKoert, line 123: Old tests shouldn't be needed to be changed right?

src/DependencyInjection/Configuration.php
kevinpapst: Ah, in that case, prototype is fine. Thanks for the clarification 👍

DjordyKoert (changes requested):

tests/Model/ModelRegistryTest.php
herndlm: that's because in ModelRegistry it is assumed that those new keys in `$alternativeNames` are set, which they always should be, because they have a default AFAIK.
But, we can keep BC here by not assuming that, sure :) adapted

src/DependencyInjection/Configuration.php
herndlm: added :)

DjordyKoert (approved):

DjordyKoert:
@herndlm sorry had a busy weekend 😅 

Anyways thank you for the PR's! ❤️

herndlm:
Thanks! I wasn't expecting you to check something here on the weekend, no worries :)

PR #2377, opened 2024-11-06
fix: override code based defaults with explicit ones

### LINKED ISSUE #2330: [Bug]: explicit setting of default value in Property causes error "Multiple definitions for [property]"
### Version

4.25.0

### Description

Since v4.25.0 we have the issue that we can no longer declare default values in a Property definition when this value differs from the value declared in code.

As an example we have this code in a larger codebase:

```php
class X 
{
	#[Property(type: 'boolean', default: true)]
    public int $aboutEnabled = 0;
}
```

When running `bin/console nelmio:apidoc:dump` this results in 

```
15:13:00 CRITICAL  [app] User Warning: Multiple definitions for @OA\Property()->default
     Using: 
  Skipping: \X->aboutEnabled in /path/to/X.php ["exception" => ErrorException { …}] ["log_uid" => "nelmio:apidoc:dump#8509bba1f254"]
15:13:00 CRITICAL  [console] Error thrown while running command "'nelmio:api:dump'". Message: "User Warning: Multiple definitions for @OA\Property()->default\n     Using: \n  Skipping: \X->aboutEnabled in /path/to/X.php" ["exception" => ErrorException { …},"command" => "'nelmio:api:dump'","message" => """  User Warning: Multiple definitions for @OA\Property()->default\n       Using: \n    Skipping: \X->aboutEnabled in /path/toX.php  """] ["log_uid" => "nelmio:apidoc:dump#8509bba1f254"]

In DefaultLogger.php line 31:
                                                                                                           
  User Warning: Multiple definitions for @OA\Property()->default                                           
       Using:                                                                                              
    Skipping: \X->aboutEnabled in /path/to/X.php  
                                                                                                           

nelmio:apidoc:dump [--area [AREA]] [--format FORMAT] [--server-url SERVER-URL] [--html-config HTML-CONFIG] [--no-pretty]
```

With version v4.24.1 everything works as expected and we get the expected OpenAPI output.

When I move the code https://github.com/nelmio/NelmioApiDocBundle/blob/12a0ff1334b439874a604c3cf6f2df7994cea54f/src/ModelDescriber/ObjectModelDescriber.php#L157-L162 behind this part https://github.com/nelmio/NelmioApiDocBundle/blob/12a0ff1334b439874a604c3cf6f2df7994cea54f/src/ModelDescriber/ObjectModelDescriber.php#L173-L176 everything works as expected as we were used to in version v4.24.1

The problem that we have is that we are using this package to create an API documentation for a legacy code-base where a lot of such quirks are necessary to declare the values that will be set from a user-perspective.

As far as I understood the code and the general working so far an explicitly manually set value of a property should always overwrite any inflected value from code-analysis. 

Am I wrong in this assumption? Or has this changed?

### JSON OpenApi

See above

### Additional context

_No response_

heiglandreas:
## Description

This change allows that defaults that are set in attributes can override the ones that are read from the code.

This fixes https://github.com/nelmio/NelmioApiDocBundle/issues/2330

Closes #2330

- [x] Bug Fix
- [ ] Feature
- [ ] Refactor
- [ ] Deprecation
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] CI

## Checklist

### DIFF
--- a/src/ModelDescriber/Annotations/ReflectionReader.php
+++ b/src/ModelDescriber/Annotations/ReflectionReader.php
@@ -43,9 +43,6 @@ class ReflectionReader
         OA\Property $property,
         ?array $validationGroups = null
     ): void {
-        if (PHP_VERSION_ID < 80000) {
-            return;
-        }
         // The default has been set by an Annotation or Attribute
         // We leave that as it is!
         if (Generator::UNDEFINED !== $property->default) {
@@ -53,7 +50,7 @@ class ReflectionReader
         }
 
         $serializedName = $reflection->getName();
-        foreach (['', 'get', 'is', 'has', 'can', 'add', 'remove', 'set'] as $prefix) {
+        foreach (['get', 'is', 'has', 'can', 'add', 'remove', 'set'] as $prefix) {
             if (0 === strpos($serializedName, $prefix)) {
                 $serializedName = substr($serializedName, strlen($prefix));
             }
@@ -106,7 +103,10 @@ class ReflectionReader
         $this->schema = $schema;
     }
 
-    private function getDefaultFromMethodReflection(\ReflectionMethod $reflection): mixed
+    /**
+     * @return mixed|string
+     */
+    private function getDefaultFromMethodReflection(\ReflectionMethod $reflection)
     {
         if (0 !== strpos($reflection->name, 'set')) {
             return Generator::UNDEFINED;
@@ -129,22 +129,22 @@ class ReflectionReader
         return $param->getDefaultValue();
     }
 
-    public function getDefaultFromPropertyReflection(\ReflectionProperty $reflection): mixed
+    /**
+     * @return mixed|string
+     */
+    public function getDefaultFromPropertyReflection(\ReflectionProperty $reflection)
     {
         $propertyName = $reflection->name;
         if (!$reflection->getDeclaringClass()->hasProperty($propertyName)) {
             return Generator::UNDEFINED;
         }
 
-        $reflectionProp = $reflection->getDeclaringClass()->getProperty($propertyName);
-        if (!$reflectionProp->hasDefaultValue()) {
+        $defaultValue = $reflection->getDeclaringClass()->getDefaultProperties()[$propertyName] ?? null;
+
+        if (null === $defaultValue) {
             return Generator::UNDEFINED;
         }
 
-        if (null === $reflectionProp->getDefaultValue()) {
-            return Generator::UNDEFINED;
-        }
-
-        return $reflectionProp->getDefaultValue();
+        return $defaultValue;
     }
 }
--- a/src/ModelDescriber/Annotations/ReflectionReader.php
+++ b/src/ModelDescriber/Annotations/ReflectionReader.php
@@ -1,0 +1,150 @@
+<?php
+
+/*
+ * This file is part of the NelmioApiDocBundle package.
+ *
+ * (c) Nelmio
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Nelmio\ApiDocBundle\ModelDescriber\Annotations;
+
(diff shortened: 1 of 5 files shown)


src/ModelDescriber/ObjectModelDescriber.php
```diff
@@ -175,6 +168,22 @@ public function describe(Model $model, OA\Schema $schema)
...
+            if (Generator::UNDEFINED !== $property->default) {
+                // Fix for https://github.com/nelmio/NelmioApiDocBundle/issues/2222
+                // When a default value has been set for the property, we can
+                // remove the field from the required fields.
+                if (is_array($schema->required) && ($key = array_search($propertyName, $schema->required, true)) !== false) {
+                    unset($schema->required[$key]);
+                }
+            }
+            if ([] === $schema->required) {
+                $schema->required = Generator::UNDEFINED;
```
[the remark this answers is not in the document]
heiglandreas: According to https://swagger.io/docs/specification/v3_0/data-models/keywords/ OPenAPI 3.0 uses https://datatracker.ietf.org/doc/html/draft-wright-json-schema-validation-00 for validation. Which states in [Section 5.15](https://datatracker.ietf.org/doc/html/draft-wright-json-schema-validation-00#section-5.15) that 

>  The value of this keyword MUST be an array.  This array MUST have at
   least one element.  Elements of this array MUST be strings, and MUST
   be unique.

Hence I reset the empty array back to `Generator::UNDEFINED` so that that doesn't become a problem downstream on the users side.

src/ModelDescriber/Annotations/ReflectionReader.php
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * Read default values of a property from the function or property signature.
+ *
+ * This needs to be called before the SymfonyConstraintAnnotationReader,
+ * otherwise required properties might be considered wrongly.
+ *
+ * @internal
+ */
+class ReflectionReader
```
DjordyKoert, line 26:
```suggestion
final class ReflectionReader
```

src/ModelDescriber/Annotations/ReflectionReader.php
```diff
@@ -0,0 +1,150 @@
...
+namespace Nelmio\ApiDocBundle\ModelDescriber\Annotations;
+
+use Nelmio\ApiDocBundle\Util\SetsContextTrait;
+use OpenApi\Annotations as OA;
+use OpenApi\Generator;
+
+/**
+ * Read default values of a property from the function or property signature.
+ *
+ * This needs to be called before the SymfonyConstraintAnnotationReader,
```
DjordyKoert, line 21:
```suggestion
 * This needs to be called before the {@see SymfonyConstraintAnnotationReader},
```

src/ModelDescriber/Annotations/ReflectionReader.php
```diff
@@ -0,0 +1,150 @@
...
+ * @internal
+ */
+class ReflectionReader
+{
+    use SetsContextTrait;
+
+    /**
+     * @var OA\Schema
+     */
+    private $schema;
```
DjordyKoert, line 33:
```suggestion
    private OA\Schema $schema;
```

DjordyKoert (changes requested):

src/ModelDescriber/ObjectModelDescriber.php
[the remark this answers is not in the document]
heiglandreas: But as I used a different approach now I completely removed it 😃

heiglandreas:
All problems should be addressed, the behaviour is almost the same as before. 

I had to change two tests where a default value was provided in the code that wasn't passed into the OpenAPI-spec previously but is passed in now.

This could be considered a BC-break but I'd argue that it was broken in the first place and is therefore a Bug-Fix

DjordyKoert (approved):
LGTM

DjordyKoert:
Thank you @heiglandreas ❤️

PR #2390, opened 2024-11-11
fix: set nullable true when default value is null

heiglandreas:
## Description

When reflection shows that the default parameter of a a property is null, then the `nullable` parameter is set to `true` for that property.

Currently that needs to be set via the attribute but that is a bit redundant as the information is already available from reflection

This change only does that for properties and for setter methods.

Right now getter methods that allow a null value to be returned are not taken into account for setting the `nullable` property.

This resets the state prior to v4.25.0 and is related to #2330

This is a refactor of #2377

- [x] Bug Fix
- [ ] Feature
- [x] Refactor
- [ ] Deprecation
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] CI

## Checklist

### DIFF
--- a/src/ModelDescriber/Annotations/ReflectionReader.php
+++ b/src/ModelDescriber/Annotations/ReflectionReader.php
@@ -54,6 +54,11 @@ final class ReflectionReader
         if ($reflection instanceof \ReflectionMethod) {
             $methodDefault = $this->getDefaultFromMethodReflection($reflection);
             if (Generator::UNDEFINED !== $methodDefault) {
+                if (null === $methodDefault) {
+                    $property->nullable = true;
+
+                    return;
+                }
                 $property->default = $methodDefault;
 
                 return;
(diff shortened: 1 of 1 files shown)


src/ModelDescriber/Annotations/ReflectionReader.php
```diff
@@ -134,12 +146,49 @@ public function getDefaultFromPropertyReflection(\ReflectionProperty $reflection
             return Generator::UNDEFINED;
         }
 
-        $defaultValue = $reflection->getDeclaringClass()->getDefaultProperties()[$propertyName] ?? null;
+        if (PHP_VERSION_ID < 80000) {
```
DjordyKoert, line 149: Why does this need to check the PHP version? I would rather this not be the case because the next major `5.x` will move to PHP 8.1 minimum

heiglandreas: As `hasDefaultValue` and `getDefaultValue` are only available since PHP8.0 I needed this "workaround" for the PHP7 users that uses a quirky way of getting the default values. That way also does not take into account possible DocBlock types so it might report back an unexpected default value when an implicit `null` is set. In PHP8.0 and higher we check for DocBlock type-infos to make a possible NULL default value plausible or not.

In he next major release this part can safely be removed as PHP8.1 supports the required functions 😉

PR #2407, opened 2024-12-09
feat(#1461): filter JMS properties by route area

### LINKED ISSUE #1461: Documentation for versioned api
Hi,
I'm thinking what is your solution to document versioned api. For example I have api in version 1.0 and i want to upgrade api to 1.1. In new version I add new field which is returned via api, how can I show it in documentation. I'm not sure how to create documentation depends on api version. Do you have some idea?

jneidel:
## Description

This PR uses the area configured in the route (like `/api/{area}/doc/`) to filter out properties on the model, which should be excluded based on JMS's versioning (also see demo below.)

Closes #1461
Refs #2129

- [ ] Bug Fix
- [x] Feature
- [ ] Refactor
- [ ] Deprecation
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] CI

## Checklist

## Demo from our project

We have an entity property:
```php
// MyEntity.php

#[Serializer\Until('v3')]
private \DateTimeInterface $createdAt;                                                                                                     
```

Opening the docs on the route: `/api/v4.3/doc/`,
which is configured as:
```yaml
# config/routes/nelmio_api_doc.yaml

nelmio_api_doc_index:
  path: /api/{area}/doc/
  defaults: { _controller: nelmio_api_doc.controller.swagger_ui }
  methods: [GET]
```

**before this PR**
Screenshot 2024-12-09 at 12 53 27
**after this PR**
Screenshot 2024-12-09 at 12 54 45

## Implementation

I opted for the word "area", since it is used in the route definition.
I proceeded as adviced in #1461.

Could you point me to where you would like to see tests created/adjusted?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # CHANGELOG
 
+* Add filtering model properties by route area to `JMSModelDescriber`
+
 ## 4.33.5
 * Added new optional parameter `$context` to` PropertyDescriberInterface::supports()`
 
@@ -25,7 +27,7 @@
 ## 4.32.3
 
 * Deprecated `Nelmio\ApiDocBundle\Annotation` namespace in favor of `Nelmio\ApiDocBundle\Attribute` namespace in preparation for 5.x. Consider upgrading to the new attribute syntax.
-```diff 
+```diff
 - use Nelmio\ApiDocBundle\Annotation\Areas;
 - use Nelmio\ApiDocBundle\Annotation\Model;
 - use Nelmio\ApiDocBundle\Annotation\Operation;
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # CHANGELOG
 
+* Add filtering model properties by route area to `JMSModelDescriber`
```
jneidel, line 3:
```suggestion
* Added filtering model properties by route area to `JMSModelDescriber`
```

src/ModelDescriber/JMSModelDescriber.php
```diff
@@ -232,6 +233,11 @@ public function getSerializationContext(Model $model): Context
             }
         }
 
+        $area = $model->getArea();
+        if (null !== $area) {
+            $context->addExclusionStrategy(new VersionExclusionStrategy($area));
```
DjordyKoert, line 238: Could you write a test? 😄

DjordyKoert (changes requested):

src/ModelDescriber/JMSModelDescriber.php
jneidel: I honestly tried, but could not figure out how to test it. I could have asked for help back then, but I didn't. Now it would be even harder to do, even with help.

PR #2420, opened 2025-01-15
chore: allow zircote/swagger-php v5

bobvandevijver:
## Description

zircote/swagger-php v5 has been released, and seems to only require minimal updates. See https://github.com/zircote/swagger-php/blob/master/docs/guide/migrating-to-v5.md.

I did need to migrate the current processor implementation to the one that is using the new processor pipeline introduced with 4.10.0, which is why I bumped the minimum constraint. In order to not mess up dependency injection, I was required to wrap the Generator class in order to allow adding additional processors using the existing compiler pass.

- [ ] Bug Fix
- [x] Feature
- [ ] Refactor
- [ ] Deprecation
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] CI

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "symfony/options-resolver": "^5.4 || ^6.4 || ^7.1",
         "symfony/property-info": "^5.4.10 || ^6.4 || ^7.1",
         "symfony/routing": "^5.4 || ^6.4 || ^7.1",
-        "zircote/swagger-php": "^4.6.1 || ^5.0"
+        "zircote/swagger-php": "^4.11.1 || ^5.0"
     },
     "require-dev": {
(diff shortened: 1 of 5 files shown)


bobvandevijver:
Well, that went well 😄 I'm looking into it

bobvandevijver:
Looking at the code coverage report: I do not think that the `before` option has ever worked at all, because the tags are name indexed array, and thus not parsed correctly. I've solved that with #2421, and there will be a merge conflict with this change.

Because of the string before support I needed to upgrade the minimum to 4.11.1, because otherwise more code from zircote would needed to be integrated here.

DjordyKoert (approved):
Thank you! ❤️

PR #2423, opened 2025-01-16
test: fix missing import + serializer was not enabled when needed

xavierleune:
## Description

Some attributes where not properly imported and the serializer was not enabled, so some tests were not doing what was expected (the actual code is fine).

## What type of PR is this? (check all applicable)
- [ ] Bug Fix
- [ ] Feature
- [ ] Refactor
- [ ] Deprecation
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] CI

## Checklist
- [ ] I have made corresponding changes to the documentation (`docs/`)
- [ ] I have made corresponding changes to the changelog (`CHANGELOG.md`)

PR #2435, opened 2025-01-21
fix(MapQueryParameter): convert pcre regex to ecma

xavierleune:
## Description
Hello everyone,
The current QueryParameterDescriber may generate an invalid definition, as [we should have only ECMA 262 patterns](https://swagger.io/docs/specification/v3_0/data-models/data-types/#strings).
This pull request ensure a better compliance, converting regexp to ecma syntax. It also fixes tests for the MapQueryParameterController, as in #2423 the namespace was not imported, so fixtures were incorrect.

Thanks !

- [x] Bug Fix
- [ ] Feature
- [ ] Refactor
- [ ] Deprecation
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] CI

## Checklist

### DIFF
--- a/src/RouteDescriber/RouteArgumentDescriber/SymfonyMapQueryParameterDescriber.php
+++ b/src/RouteDescriber/RouteArgumentDescriber/SymfonyMapQueryParameterDescriber.php
@@ -128,7 +128,7 @@ final class SymfonyMapQueryParameterDescriber implements RouteArgumentDescriberI
         }
 
         if (FILTER_VALIDATE_REGEXP === $filter) {
-            return ['type' => 'string', 'pattern' => $options['regexp']];
+            return ['type' => 'string', 'pattern' => $this->getEcmaRegexpFromPCRE($options['regexp'])];
         }
 
         if (FILTER_VALIDATE_URL === $filter) {
@@ -140,5 +140,31 @@ final class SymfonyMapQueryParameterDescriber implements RouteArgumentDescriberI
         }
 
         return [];
+    }
+
+    private function getEcmaRegexpFromPCRE(string $pcreRegex): string
+    {
+        // Check if PCRE regex has delimiters
+        if (preg_match('/^(.)(.*)\1([a-zA-Z]*)$/s', $pcreRegex, $matches)) {
+            [$fullMatch, $delimiter, $pattern, $flags] = $matches;
+            // Remove escaped delimiters in the pattern
+            $pattern = str_replace('\\'.$delimiter, $delimiter, $pattern);
+        } else {
+            // No delimiter has been found, let's consider it's a valid regex without delimiter. Happens when regexp has been takend from "requirements" in route.
+            $pattern = $pcreRegex;
+        }
+
+        $pattern = str_replace(['\A', '\z'], ['^', '$'], $pattern); // Supported features but different syntax
+
+        // Check for unsupported PCRE specific constructs
+        $unsupportedFeatures = [
+            '\Z', // End of string before newline (not supported in JavaScript)
+            '\R', // Any Unicode newline sequence (not supported in JavaScript)
+            '\K', // Resets the start of the current match (not supported in JavaScript)
+        ];
+        $pattern = str_replace($unsupportedFeatures, '', $pattern);
+
+        // Return only the pattern (without flags or delimiters)
+        return $pattern;
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## 4.38.1
+- Support of attribute MapQueryParameter with a regexp has been improved, it now converts the regexp from PCRE to ECMA-262 for better compliance with OpenApi. 
+
 ## 4.38.0
 * Added a `#[Ignore]` attribute that allows a property to be excluded from the generated schema.
 ```php
--- a/tests/Functional/Controller/MapQueryParameterController.php
+++ b/tests/Functional/Controller/MapQueryParameterController.php
@@ -11,6 +11,7 @@
 
 namespace Nelmio\ApiDocBundle\Tests\Functional\Controller;
 
+use OpenApi\Attributes as OA;
 use Symfony\Component\HttpKernel\Attribute\MapQueryParameter;
 use Symfony\Component\Routing\Annotation\Route;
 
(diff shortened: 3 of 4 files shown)


xavierleune:
Hi @DjordyKoert, you may have missed this pull request, what do you think about it ?
Have a nice day

xavierleune:
In the meantime, as a workaround, for anyone that got here, the following processor does the job to:
```php
<?php

namespace App\ApiDoc;

use OpenApi\Analysis;
use OpenApi\Generator;

/**
 * This Processor fixes a bug in NelmioApiDocBundle, see {https://github.com/nelmio/NelmioApiDocBundle/pull/2435}
 * It may be removed after this pull request has been merged.
 */
class ConvertPCREToEcma
{
    public function __invoke(Analysis $analysis): void
    {
        foreach ($analysis->annotations as $annotation) {
            if (property_exists($annotation, 'paths') && $annotation->paths !== Generator::UNDEFINED) {
                foreach ($annotation->paths as $path) {
                    foreach ($path::$_nested as $nested) {
                        $nestedProperty = is_array($nested) ? $nested[0] : $nested;
                        if ($path->$nestedProperty === Generator::UNDEFINED || $path->$nestedProperty->parameters === Generator::UNDEFINED) {
                            continue;
                        }
                        foreach ($path->$nestedProperty->parameters as $parameter) {
                            if ($parameter->schema === Generator::UNDEFINED || $parameter->schema->pattern === Generator::UNDEFINED) {
                                continue;
                            }
                            $parameter->schema->pattern = $this->getEcmaRegexpFromPCRE($parameter->schema->pattern);
                        }
                    }
                }
            }
        }
    }
    private function getEcmaRegexpFromPCRE(string $pcreRegex): string
    {
        // Check if PCRE regex has delimiters
        if (!preg_match('/^(.)(.*)\1([a-zA-Z]*)$/s', $pcreRegex, $matches)) {
            throw new \InvalidArgumentException('Invalid PCRE regex format. Missing delimiters.');
        }

        [$fullMatch, $delimiter, $pattern, $flags] = $matches;

        // Check for unsupported PCRE specific constructs
        $unsupportedFeatures = [
            '\A', // Start of string (use ^ in JavaScript)
            '\z', // End of string (use $ in JavaScript)
            '\Z', // End of string before newline (not supported in JavaScript)
            '\R', // Any Unicode newline sequence (not supported in JavaScript)
            '\K', // Resets the start of the current match (not supported in JavaScript)
        ];

        foreach ($unsupportedFeatures as $feature) {
            if (str_contains($pattern, $feature)) {
                throw new \InvalidArgumentException("Unsupported PCRE feature found: {$feature}");
            }
        }

        // Remove escaped delimiters in the pattern
        $pattern = str_replace('\\'.$delimiter, $delimiter, $pattern);

        // Return only the pattern (without flags or delimiters)
        return $pattern;
    }
}
```

src/RouteDescriber/RouteArgumentDescriber/SymfonyMapQueryParameterDescriber.php
```diff
@@ -141,4 +141,35 @@ private function describeValidateFilter(?int $filter, int $flags, array $options
...
+        [$fullMatch, $delimiter, $pattern, $flags] = $matches;
+
+        // Check for unsupported PCRE specific constructs
+        $unsupportedFeatures = [
+            '\A', // Start of string (use ^ in JavaScript)
+            '\z', // End of string (use $ in JavaScript)
+            '\Z', // End of string before newline (not supported in JavaScript)
+            '\R', // Any Unicode newline sequence (not supported in JavaScript)
+            '\K', // Resets the start of the current match (not supported in JavaScript)
+        ];
```
DjordyKoert, line 161: Could you write tests for these? :)

src/RouteDescriber/RouteArgumentDescriber/SymfonyMapQueryParameterDescriber.php
```diff
@@ -141,4 +141,35 @@ private function describeValidateFilter(?int $filter, int $flags, array $options
 
         return [];
     }
+
+    private function getEcmaRegexpFromPCRE(string $pcreRegex): string
+    {
+        // Check if PCRE regex has delimiters
+        if (!preg_match('/^(.)(.*)\1([a-zA-Z]*)$/s', $pcreRegex, $matches)) {
+            throw new \InvalidArgumentException('Invalid PCRE regex format. Missing delimiters.');
```
DjordyKoert, line 149: Maybe we shouldn't throw here to preserve original behaviour? That way it 'tries its best' to generate documentation.

```suggestion
            return $pcreRegex;
```

src/RouteDescriber/RouteArgumentDescriber/SymfonyMapQueryParameterDescriber.php
```diff
@@ -141,4 +141,35 @@ private function describeValidateFilter(?int $filter, int $flags, array $options
...
+        }
+
+        [$fullMatch, $delimiter, $pattern, $flags] = $matches;
+
+        // Check for unsupported PCRE specific constructs
+        $unsupportedFeatures = [
+            '\A', // Start of string (use ^ in JavaScript)
+            '\z', // End of string (use $ in JavaScript)
```
DjordyKoert, line 157: Can we not convert these to the proper format?

src/RouteDescriber/RouteArgumentDescriber/SymfonyMapQueryParameterDescriber.php
```diff
@@ -141,4 +141,35 @@ private function describeValidateFilter(?int $filter, int $flags, array $options
...
+            '\Z', // End of string before newline (not supported in JavaScript)
+            '\R', // Any Unicode newline sequence (not supported in JavaScript)
+            '\K', // Resets the start of the current match (not supported in JavaScript)
+        ];
+
+        foreach ($unsupportedFeatures as $feature) {
+            if (str_contains($pattern, $feature)) {
+                throw new \InvalidArgumentException("Unsupported PCRE feature found: {$feature}");
```
DjordyKoert, line 165: It might not always be possible for users of this bundle to update the regexp for php's `filter_var` because this might result in different behaviour.

```suggestion
                return $pcreRegex;
```

DjordyKoert (changes requested):

xavierleune:
Hi @DjordyKoert,
Thanks for your feedback, the code is now way more tolerant and will try to generate a regexp and won't throw exceptions anymore.
This will improve the DX a lot, you're perfectly right 👍

DjordyKoert (approved):
Looking good, thank you :)

Issue #2502, opened 2025-06-16
[Bug]: Symfony 7.3 - The "Symfony\Component\PropertyInfo\PropertyInfoExtractor::getTypes()" method is deprecated, use "Symfony\Component\PropertyInfo\PropertyInfoExtractor::getType()" instead.


AxaliaN:
### Version

5.3.0

### Description

Heyo, since Symfony 7.3.0, a new deprecation warning is being raised:

The \"Symfony\\Component\\PropertyInfo\\PropertyInfoExtractor::getTypes()\" method is deprecated, use \"Symfony\\Component\\PropertyInfo\\PropertyInfoExtractor::getType()\" instead.

It is triggered by the rendering of the docs.

DjordyKoert:
using `type_info: true` in your bundle config should fix this 😄 

https://symfony.com/bundles/NelmioApiDocBundle/current/configuration_reference.html#type-info

AxaliaN:
Hey @DjordyKoert that fixes most of them, but there is 1 left:

User Deprecated: Since symfony/property-info 7.3: The \"Symfony\\Component\\PropertyInfo\\Type\" class is deprecated. Use \"Symfony\\Component\\TypeInfo\\Type\" class from \"symfony/type-info\" instead.

Which is triggered in \Nelmio\ApiDocBundle\Model\ModelRegistry, but maybe that's a separate isuse 😅

DjordyKoert:
Yeah the property-info component is deprecating it's `Type` class in favor of type-info's `Type` class, but a lot of the internal code heavily relies on `Symfony\Component\PropertyInfo\Type`.

So it's quite the challenge to change all the usages without making breaking changes 😓

adrian-delgado:
The `Symfony\Component\PropertyInfo\PropertyInfoExtractor::getType()` was only introduced in [PropertyInfo 7.1](https://github.com/symfony/property-info/blob/7.3/CHANGELOG.md#71):
> 7.1
    ...
    Add PropertyTypeExtractorInterface::getType() as experimental

But v5 of this bundle is marked as compatible with [symfony/property-info 6.4 || ^7.1](https://github.com/nelmio/NelmioApiDocBundle/blob/5.x/composer.json#L33), resulting in the following error when used with Symfony 6 and `type_info: true`:
```
Symfony\Component\ErrorHandler\Error\UndefinedMethodError:
[...]

PR #2503, opened 2025-06-20
feat: Support for generic types

bnowak:
## Description

It adds support for generating api schema of generic types.

- [ ] Bug Fix
- [x] Feature
- [ ] Refactor
- [ ] Deprecation
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] CI

## Checklist


bnowak:
Regarding generated model/component schema names for generic type classes, I was also thinking about improving it a bit in the same fashion as in phpdoc (eg. `GenericClass<string>`, `GenericClass<int>`, `GenericClass<list<string>>` etc.). IMO that would be much more readable for humans rather than current default approach with enumerating (`GenericClass1`, `GenericClass2`, `GenericClass3` ...).

I have some wip code about that, however I spotted some issues:
- ideally to handle names like that we'd need replace [Model's legacy type property](https://github.com/nelmio/NelmioApiDocBundle/blob/5.x/src/Model/Model.php#L18) with version from [TypeInfo component](https://github.com/symfony/type-info/blob/7.3/Type.php). It can store generic type variables, so that's enough to simply refactor/adjust [`ModelRegistry::getTypeShortName`](https://github.com/nelmio/NelmioApiDocBundle/blob/5.x/src/Model/ModelRegistry.php#L191) method and that's it. However, replacing that would be more work and we'd need to drop support for SF property-info component and make SF type-info required (now it's optional). So I guess that's option for future and rather long term solution.
- the other option is to use introduced in this PR `Model::$serializationContext[GenericClassDescriber::TEMPLATES_KEY]` types map to resolve generic type names, however it seems like workaround with a few type-based and type-info component-based conditions in [ModelRegistry](https://github.com/nelmio/NelmioApiDocBundle/blob/5.x/src/Model/ModelRegistry.php) codebase. It would be dirty, but would work.

@DjordyKoert could you kindly take a look on that PR please?

also, please respond with your feelings about above generation of model names for generics. If option 2 if fine for you - I could address that in separate PR :)

cheers

bnowak:
Heya @DjordyKoert. Any chances to review this soon please? :sweat_smile:

DjordyKoert:
> ideally to handle names like that we'd need replace [Model's legacy type property](https://github.com/nelmio/NelmioApiDocBundle/blob/5.x/src/Model/Model.php#L18) with version from [TypeInfo component](https://github.com/symfony/type-info/blob/7.3/Type.php). It can store generic type variables, so that's enough to simply refactor/adjust [ModelRegistry::getTypeShortName](https://github.com/nelmio/NelmioApiDocBundle/blob/5.x/src/Model/ModelRegistry.php#L191) method and that's it. However, replacing that would be more work and we'd need to drop support for SF property-info component and make SF type-info required (now it's optional). So I guess that's option for future and rather long term solution.

I am working on a solution for replacing the deprecated type with type-info in my free-time. As you already expected this is quite a challenge to ensure no breaking changes are made e.t.c.

Ideally we use that as it's just easier to work with.

Also relates to deprecation warning mentioned in https://github.com/nelmio/NelmioApiDocBundle/issues/2502

DjordyKoert:
> Regarding generated model/component schema names for generic type classes, I was also thinking about improving it a bit in the same fashion as in phpdoc (eg. `GenericClass<string>`, `GenericClass<int>`, `GenericClass<list<string>>` etc.). IMO that would be much more readable for humans rather than current default approach with enumerating (`GenericClass1`, `GenericClass2`, `GenericClass3` ...).

It's more readable, but we can't use `<` & `>` as schema names according [to the spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.1.md#components-object).

At the time of writing it must satisfy the `^[a-zA-Z0-9\.\-_]+$` regex. We could use `.` instead in theory (`GenericClass.int`, `GenericClass.list.string`, `GenericClass.string`).

bnowak:
@DjordyKoert thanks for response. As mentioned naming changes are more dependent on refactor and future work, so I wouldn't like to block current PR because of that.

Apart of that, current changes look fine for you? Any adjustment/fixes needed or we can merge them as it is?

DjordyKoert:
> @DjordyKoert thanks for response. As mentioned naming changes are more dependent on refactor and future work, so I wouldn't like to block current PR because of that.
> 
> Apart of that, current changes look fine for you? Any adjustment/fixes needed or we can merge them as it is?

I will try to take a closer look at it some time soon :)

DjordyKoert (approved):
Nice work! Thanks a lot ❤️

bnowak:
Thanks as well 😉

PR #2542, opened 2025-09-01
feat(Model): Allow customizing the name of generated schemas.

### LINKED ISSUE #2113: Allow the use of namespace as the Model name
It would be nice if we had the option to use the namespace as the model name.

In my project we follow a approach where we have multiple DTOs for the same entity, each one including the attributes needed for a use case. 

So, we end up having multiple models with the same name that are renamed automatically to ModelName2, ModelName3 etc.

I would be nice to have the option to use the namespace as the name or to be able to extend the name generator to do that.

### LINKED ISSUE #2515: [Question]: Custom model names
### Version

5.0.0

### Question

Hey, 
I was wondering whether there was any reason against having something like a `$customName` parameter for `Model`. This would then allow to rename the generated schema to something user defined instead of using the class name. In my project I am really struggling this. I only use part of the properties of some object using groups, but the generated schema still has the name of the object. Even tho this name does not make sense for the parameter set I'm using. 

I know i can define alternative names in the config file, but that gets hard to maintain as it grows. It also does not really make sense to me to separate this from the rest of the definitions.

Sorry if I'm misunderstanding something and I'm doing it all wrong

### Additional context

_No response_

### LINKED ISSUE #2540: [Feature]: Support for overriding model naming strategy
### Description

Model names for DTOs are based on the class name (without namespace) and assigned a incremental suffix number when there are duplicates. In large projects (eg: mono repo with multiple apps) or depending on the naming strategies for the PHP classes, we can end up with the same name quite often. 

Instead it would be interesting to be able to choose how the model name is generated (and then apply a suffix if needed). By providing an instance of some kind of:

```php
interface ModelNamingStrategy {
  public function getName(Type $type): string;
  public function supports(Type $type): bool;
}
```

To avoid breaking change the default should stay the same but it's likely that people will want to generate the name using the FQCN, so one that does that _could_ be provided by the package (but not a big deal if that's left for the user to implement)

(related to https://github.com/nelmio/NelmioApiDocBundle/issues/2113)

### Additional context

_No response_

DjordyKoert:
## Description

Adds a new `name` property the `#[Model]` attribute & `Model` class for registering schema's.

This makes it possible to customize the name that needs to be registered for a model schema.

This PR also comes with a refactor to the "alternative names" logic, to ensure deduplication logic (introduced in https://github.com/nelmio/NelmioApiDocBundle/pull/2461) is also retained for manually registered schema's through the bundle config.

> [!NOTE]
> This PR, also changes that manually registered schema's get registered **first** now. (Instead of registering these last)

Closes #2540, closes #2113, closes #2515

- [ ] Bug Fix
- [x] Feature
- [x] Refactor
- [ ] Deprecation
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] CI

## Checklist
- [x] I have made corresponding changes to the documentation (`docs/`)

### DIFF
--- a/src/Model/ModelRegistry.php
+++ b/src/Model/ModelRegistry.php
@@ -82,7 +82,11 @@ final class ModelRegistry
     public function register(Model $model, ?string $alternativeName = null): string
     {
         $hash = $model->getHash();
-        $identifier = $this->getTypeShortName($model->getType()).$model->name;
+        $identifier = \sprintf(
+            '%s|%s',
+            $this->getTypeShortName($model->getType()),
+            $model->name ?? $alternativeName,
+        );
 
         $schema = null;
         if (!isset($this->names[$hash])) {
--- a/src/Model/ModelRegistry.php
+++ b/src/Model/ModelRegistry.php
@@ -29,6 +29,11 @@ final class ModelRegistry
     private array $registeredModelNames = [];
 
     /**
+     * @var Model[]
+     */
+    private array $alternativeNames = [];
+
+    /**
      * @var string[] List of hashes of models that have not been registered yet
      */
     private array $unregistered = [];
@@ -67,13 +72,16 @@ final class ModelRegistry
         $this->api = $api;
         $this->logger = new NullLogger();
         foreach (array_reverse($alternativeNames) as $alternativeName => $criteria) {
-            $this->register(new Model(
+            $this->alternativeNames[] = $model = new Model(
                 new Type('object', false, $criteria['type']),
                 $criteria['groups'],
                 $criteria['options'] ?? [],
                 $criteria['serializationContext'] ?? [],
                 name: $alternativeName,
-            ));
+            );
+            $this->names[$model->getHash()] = $alternativeName;
+            $this->registeredModelNames[$alternativeName] = $model;
+            Util::getSchema($this->api, $alternativeName);
         }
     }
 
--- a/src/Model/ModelRegistry.php
+++ b/src/Model/ModelRegistry.php
@@ -72,19 +72,21 @@ final class ModelRegistry
                 $criteria['groups'],
                 $criteria['options'] ?? [],
                 $criteria['serializationContext'] ?? [],
-                name: $alternativeName,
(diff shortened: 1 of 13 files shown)


src/Model/ModelRegistry.php
```diff
@@ -87,23 +87,23 @@ public function __construct($modelDescribers, OA\OpenApi $api, array $alternativ
...
+            $this->registeredModelNames[$name] = $model;
 
             $schema = $this->describeSchema($model, null);
             // Only try to match schemas if we successfully got one
             if (null !== $schema) {
-                foreach ($this->schemasPerType[$type] ?? [] as $schemaAndModel) {
-                    if ($schemaAndModel['schema'] === json_encode($schema->jsonSerialize())) {
-                        $newHash = $schemaAndModel['model']->getHash();
-                        unset($this->names[$hash], $this->registeredModelNames[$hash]);
+                $schemaJson = json_encode($schema);
```
mathroc, line 104: wouldn't the hash be enough instead of the json encoded schema? (and more efficient) or do you want to dedup models with different hashes that would generate the same schema?

mathroc (approved):
left a comment but it's just me wondering, might be completely wrong since I'm not familiar with the codebase

thank you for looking into this!

src/Model/ModelRegistry.php
DjordyKoert: Correct, we want to dedup models with different hashes that generate the same schema https://github.com/nelmio/NelmioApiDocBundle/pull/2461 😄

mathroc: makes sense :+1:

japotta08:
https://documentationacademy.org/

ABOsalah866 (approved):

ABOsalah866 (approved):

ABOsalah866 (approved):

ABOsalah866 (approved):

ABOsalah866 (approved):

ABOsalah866 (approved):

Issue #2145, opened 2023-11-07
OpenAPI Response Schema Not Reflecting Serialization Context for Nested Groups


ferhado:
## Issue Description
The OpenAPI-generated response schema for my `User` entity does not properly reflect the serialization context defined for a nested property. Specifically, the `company` property within `User` should serialize with the `company:relation` group when the `:company` group is used.

## Example for Clarification

In `UserController.php`:
```php
#[OA\Response(
  response: 200,
  description: "",
  content: new Model(type: User::class, groups: ["detail", ":company"])
)]
#[Route('/users/{id<\d+>}', methods: ["GET"])]
public function getUserById(Request $request) {
  $id = $request->get("id");
  $user = $this->userRepo->find($id);
  return $this->json($user, context: ["groups" => ["detail", ":company"]]);
}
```

In `User.php`:
```php
// User Entity:
class User implements UserInterface, PasswordAuthenticatedUserInterface {
  #[ORM\Id]
  #[ORM\GeneratedValue]
  #[ORM\Column]
  #[Groups(["user:rel", "list", "detail"])] // expected for main Entity
  private ?int $id = null;

  #[ORM\Column(length: 100)]
  #[Groups(["user:rel", "list", "detail", "write"])] // expected for main Entity
  private ?string $name = null;

  #[ORM\Column(length: 180, unique: true)]
  #[Groups(["user:rel", "list", "detail", "write", "auth"])] // expected for main Entity
  private ?string $username = null;

  #[ORM\Column]
  private array $roles = [];

  #[ORM\Column]
  #[Groups(["write", "auth"])]
  private ?string $password = null;

  #[Groups([":company"])] // expected for main Entity
  #[Context(context: ["groups" => ["company:relation"]])] // <------------- Not reflected Context
  #[ORM\ManyToOne
  #[ORM\JoinColumn(nullable: true)]
  private ?Company $company = null;

  #[ORM\Column(type: "boolean")]
  #[Groups(["list", "detail", "write"])] // expected for main Entity
  private bool $isActive = true;
```
In `Company.php`:
```php
class Company {
  #[ORM\Id]
  #[ORM\GeneratedValue]
  #[ORM\Column]
  #[Groups(["list", "detail", "company:rel"])] // expected for nested Entity
  private ?int $id = null;

  #[ORM\Column(length: 100, nullable: true)]
  #[Groups(["list", "detail", "company:rel"])] // expected for nested Entity
  private ?string $name = null;

[...]

PR #2551, opened 2025-09-08
fix: Handle Symfony serializer `datetime_format` context for date formatting

Matthew-Kilpatrick:
## Description
Symfony serializer supports a `Context` attribute which can provide details for how to normalize a property. Specifically, a `datetime_format` key can be provided to specify how to format a `DateTime` field. This PR handles a `datetime_format` value of `Y-m-d` and sets the `format` to `date`, rather than the current incorrect value of `date-time`.

I don't believe there are any issues this PR will close, though this was brought up in a comment at https://github.com/nelmio/NelmioApiDocBundle/issues/2145#issuecomment-1932547474

- [x] Bug Fix
- [ ] Feature
- [x] Refactor
- [ ] Deprecation
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] CI

## Checklist
- [x] I have made corresponding changes to the documentation (`docs/`)
- [] I have made corresponding changes to the changelog (`CHANGELOG.md`)

## Uncertainties
- The most suitable place I could find to place this was alongside the Symfony constraint annotation readers (`SymfonyConstraintAnnotationReader`), and making this a more generic `SymfonyAnnotationReader` class. The class was documented as `@internal` so I don't believe this would be a breaking change. If you'd rather I go down a different route, please give me some pointers of where/how you'd like this implemented.
- I'm torn whether this should be classed as a feature (from the bundle perspective, functionality to parse this field doesn't yet exist) or bug (because from a user perspective, the bundle is generating the wrong format) - and as such, haven't updated the changelog yet.
- I don't think any documentation changes should be required for this, so have checked this off on the checklist.

### DIFF
--- a/src/ModelDescriber/Annotations/AnnotationsReader.php
+++ b/src/ModelDescriber/Annotations/AnnotationsReader.php
@@ -60,13 +60,14 @@ class AnnotationsReader
     /**
      * @param \ReflectionProperty|\ReflectionMethod $reflection
      * @param string[]|null                         $serializationGroups
+     * @param array<string, mixed>                  $context
      */
-    public function updateProperty($reflection, OA\Property $property, ?array $serializationGroups = null): void
+    public function updateProperty($reflection, OA\Property $property, array &$context, ?array $serializationGroups = null): void
     {
         $this->openApiAnnotationsReader->updateProperty($reflection, $property, $serializationGroups);
         $this->phpDocReader->updateProperty($reflection, $property);
         $this->reflectionReader->updateProperty($reflection, $property);
-        $this->symfonyAnnotationReader->updateProperty($reflection, $property, $serializationGroups);
+        $this->symfonyAnnotationReader->updateProperty($reflection, $property, $context, $serializationGroups);
     }
 
     /**
--- a/src/ModelDescriber/Annotations/AnnotationsReader.php
+++ b/src/ModelDescriber/Annotations/AnnotationsReader.php
@@ -23,7 +23,7 @@ class AnnotationsReader
 {
     private PropertyPhpDocReader $phpDocReader;
     private OpenApiAnnotationsReader $openApiAnnotationsReader;
-    private SymfonyConstraintAnnotationReader $symfonyConstraintAnnotationReader;
+    private SymfonyAnnotationReader $symfonyAnnotationReader;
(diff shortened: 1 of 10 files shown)


DjordyKoert:
Thanks for opening a PR!

I think we should expand the `$context` argument (which is passed to various describers) based on whatever is set in the `#[Context]`. 

Then the `DateTimePropertyDescriber` can for example set a different format based on the passed context.

Matthew-Kilpatrick:
Thanks for your feedback, that does seem like a much better solution than what I originally went with!

As far as I could tell, there's no cases yet where the `$context` value passed to the describers has property/method-level context, and the current behaviour seemed to pass `$model->getSerializationContext()`. So, I've done some slight refactoring so that `ObjectModelDescriber` instantiates an array of property context data, which gets passed by reference to `AnnotationsReader`, so that annotation readers can update the context here. This then gets passed to `describeProperty `in `ObjectModelDescriber`, where it's merged with the `$model->getSerializationContext()` (I originally tried instantiating it to this in `describe`, but some of the JMS tests fail as I think the context is updated).

I've kept the processing for checking if the `Context` attribute exists in the `SymfonyAnnotationReader` class which I renamed from `SymfonyConstraintAnnotationReader` in my previous commit as this feels like the best place for this to live, but if you'd like this to go somewhere else, please let me know!

The alternative I considered which would have required less refactoring was to store to the `_context` property of the `OA\Schema` passed to the describers & annotation readers - although, this felt a lot messier (I'd rather use the intended parameter getting passed around).

I've pushed a new commit with these changes to help keep track of my changes, but can squash everything down into a single commit once you're happy with everything before merging!

DjordyKoert:
I'm not sure if passing the context data into a `symfony_context` key is the appropriate solution. 

That means we cannot easily use the same references Symfony does internally.

```php
$context['symfony_context']['datetime_format']
```

vs

```php
$context['datetime_format']
```

I think the second variant makes it theoretically easier for users extending the bundle can have the same expectations as the serializer context.

DjordyKoert:
I personally also want to still refactor these `AnnotationReader`s because they are quite complex, hard to understand & difficult to expand upon easily.

This is not related to your PR though and more of a personal 'I want to fix this in the future', but more logic in these classes won't make it easier to refactor these. 😢

PR #2641, opened 2025-12-02
style(tests): add missing `void` return types

lacatoire:
This PR adds explicit void return types to methods that don't return any value.


DjordyKoert:
Thank you! ❤️

PR #2643, opened 2025-12-02
Replace deprecated LegacyTypeConverter::createType() with Type::fromString()

lacatoire:
This PR removes the deprecated LegacyTypeConverter::createType() usage and replaces it with Type::fromString(), which is the recommended API since Symfony 7.2.

Only the safe and forward-compatible part of the LegacyTypeConverter is removed toTypeInfoType() and toLegacyType() remain untouched, as they still provide conversions that have no public equivalent in Symfony.
Happy to adjust if needed.


DjordyKoert:
There is no `Type::fromString` method?

PR #2659, opened 2025-12-27
feat: Add support for static callbacks in Symfony choice constraints

vveselinov:
## Description

This PR adds support for generating enums from Symfony Choice constraints when the constraint uses a PHP 8.5 static callback defined directly inside an attribute.

For example
```php
class TestDto
{
    #[Assert\Choice(callback: static function () {
        return ['test1', 'test2'];
    })]
    public $property1;
}
```

In this case, the generator will correctly infer the allowed values and produce the corresponding enum.

Notes
* This feature is only available on PHP 8.5+
* Existing behavior for method-based callbacks remains unchanged

- [ ] Bug Fix
- [x] Feature
- [ ] Refactor
- [ ] Deprecation
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] CI

## Checklist

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -15,7 +15,8 @@ $finder = (new PhpCsFixer\Finder())
     ->in(__DIR__)
     ->exclude('var')
     ->exclude('tests/Functional/cache')
-    ->exclude('tests/Functional/ModelDescriber/Fixtures');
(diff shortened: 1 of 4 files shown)


vveselinov:
The PHP 8.5 code in the tests breaks runners for lower versions. I am unsure how to proceed without a huge refactoring of the tests.

DjordyKoert:
Thank you! 😄

PR #2687, opened 2026-02-18
fix: LegacyType deprecation warning in Configuration

### LINKED ISSUE #2677: [Bug]: checking LegacyType triggers deprecation warning
### Version

5.9.2

### Description

usage of the `Symfony\Component\PropertyInfo\Type` as legacy type in this bundle's `Configuration` to define the default value and validation for the 'type_info' config triggers a deprecation warning when using `bin/console debug:container --deprecations` in an application using this bundle

could this check perhaps be inverted by checking for the `Symfony\Component\TypeInfo\Type` instead?

### JSON OpenApi

_No response_

### Additional context

_No response_

lacatoire:
## Summary

- Check for the presence of `Symfony\Component\TypeInfo\Type` instead of the absence of the deprecated `Symfony\Component\PropertyInfo\Type` to avoid triggering a deprecation warning on Symfony 7.x
- Use `!method_exists(PropertyInfoExtractor::class, 'getTypes')` to reliably detect Symfony 8+ for the validation that prevents setting `type_info: false`

Fixes #2677

### DIFF
--- a/tests/DependencyInjection/ConfigurationTest.php
+++ b/tests/DependencyInjection/ConfigurationTest.php
@@ -17,6 +17,7 @@ use PHPUnit\Framework\TestCase;
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\Config\Definition\Processor;
 use Symfony\Component\PropertyInfo\PropertyInfoExtractor;
+use Symfony\Component\PropertyInfo\Type as LegacyType;
 use Symfony\Component\TypeInfo\Type as TypeInfoType;
 
 class ConfigurationTest extends TestCase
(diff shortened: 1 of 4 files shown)


tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -263,7 +263,7 @@ public static function provideInvalidConfiguration(): \Generator
             ];
         }
 
-        if (!class_exists(LegacyType::class)) {
+        if (!class_exists('Symfony\Component\PropertyInfo\Type')) {
```
DjordyKoert:
```suggestion
        if (!class_exists(LegacyType::class)) {
```

PR #2688, opened 2026-02-18
fix: Count constraint uses minProperties/maxProperties for maps

### LINKED ISSUE #2578: [Bug]: Symfony Constraints Count should set (min|max)Properties instead of (min|max)Items on maps (`array<X, Y>`)
### Version

5.6.5

### Description

Let's say you have a property like this one :

```php
    /**
     * @var array<string, mixed>
     */
    #[Assert\NotNull, Assert\Count(min: 1)]
    public array $someMap;
```

The generated OpenApi will currently be :

```json
{
    "someMap": {
        "type": "object",
        "minItems:" 1
    }
}
```

But "object" doesn't have a "minItems" (nor max) property in OpenAPI, only "minProperties" (and max). So currently I've gone around it in our custom describer using this hack :

```php
if ($this->isArrayMap($property)) {
    if (Generator::UNDEFINED !== $property->minItems) {
        $property->minProperties = $property->minItems;
        $property->minItems = Generator::UNDEFINED /** @phpstan-ignore assign.propertyType */;
    }
    if (Generator::UNDEFINED !== $property->maxItems) {
        $property->maxProperties = $property->maxItems;
        $property->maxItems = Generator::UNDEFINED /** @phpstan-ignore assign.propertyType */;
    }
}
```

Which also reveals that "(min|max)Items" typing is wrong (it says `@var int` but is initialized with `Generator::UNDEFINED`).

The fix should most probably be in `SymfonyConstraintAnnotationReader` lines 90-96.

lacatoire:
## Summary

- When a schema has type `object` (e.g. `array<string, Y>` maps), the `Count` constraint now correctly sets `minProperties`/`maxProperties` instead of `minItems`/`maxItems`, per the OpenAPI specification
- Schemas with type `array` continue using `minItems`/`maxItems` as before

Fixes #2578

### DIFF
--- a/src/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
+++ b/src/ModelDescriber/Annotations/SymfonyConstraintAnnotationReader.php
@@ -97,11 +97,20 @@ class SymfonyConstraintAnnotationReader
             } elseif ($attribute instanceof Assert\Regex) {
                 $this->appendPattern($property, $attribute->getHtmlPattern());
             } elseif ($attribute instanceof Assert\Count) {
-                if (isset($attribute->min)) {
-                    $property->minItems = $attribute->min;
-                }
-                if (isset($attribute->max)) {
-                    $property->maxItems = $attribute->max;
+                if ('object' === $property->type) {
+                    if (isset($attribute->min)) {
+                        $property->minProperties = $attribute->min;
+                    }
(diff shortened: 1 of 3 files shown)


DjordyKoert:
@lacatoire thank you very much for all the pr's!! ❤️

